PR #482, opened 2022-03-08
Simplify ELB

### LINKED ISSUE #474: ELB simplification

aguillenv:
Resolves #474


test/modules/aws-elb-integration.ts
```diff
@@ -198,122 +167,3 @@ describe('ELB Integration Testing', () => {
     .remove(dbAlias, 'not-needed')
     .then(...finish(done)));
 });
-
-describe('ELB Integration Testing SP', () => {
```
depombo, line 202: I think there is a few `SELECT` checks in this suite that are not in the one above?

depombo (approved):

src/modules/aws_elb@0.0.1/index.ts
```diff
@@ -56,15 +43,13 @@ export const AwsElbModule: Module = new Module({
...
-      out.defaultActions = await Promise.all(l.DefaultActions?.map(async a => {
-        const dbAct = await ctx.orm.findOne(AwsAction, {
-          where: {
-            targetGroup: { targetGroupArn: a.TargetGroupArn },
-          },
-          relations: ['targetGroup'],
-        });
-        if (!dbAct) return AwsElbModule.utils.actionMapper(a, ctx);
-        return dbAct;
+      await Promise.all(l.DefaultActions?.map(async a => {
```
dfellis, line 46: Should this be refactored into a `for await` loop instead?

test/modules/aws-ecs-integration.ts
```diff
@@ -111,9 +111,26 @@ describe('ECS Integration Testing', () => {
...
+            (aws_load_balancer_id, aws_security_group_id)
+        VALUES
+            ((SELECT id FROM aws_load_balancer WHERE load_balancer_name = '${serviceLoadBalancerName}' LIMIT 1),
+              (SELECT id FROM aws_security_group WHERE group_name = 'default' LIMIT 1));
+        INSERT INTO aws_listener
+            (aws_load_balancer_id, port, protocol, action_type, target_group_id)
+        VALUES 
+            ((SELECT id FROM aws_load_balancer WHERE load_balancer_name = '${serviceLoadBalancerName}' LIMIT 1), 
+              ${hostPort}, 'HTTP', 'forward', (SELECT id FROM aws_target_group WHERE target_group_name = '${serviceTargetGroupName}' LIMIT 1));
+      COMMIT;
```
dfellis, line 133: Does this *need* to be atomic, or are you just making sure cruft isn't left behind in the DB if it fails?

aguillenv: yes, just making sure that when we execute multiple statements won't left something weird behind

src/modules/aws_elb@0.0.1/index.ts
aguillenv: It could yes, probably better

Issue #450, opened 2022-02-28
Simplify module migration file management


depombo:
- [x] Do not require importing the migration file in the module `index.ts` file (automatically find it instead with `require` trickery)
- [x] Make `gen-module` aware of module dependencies so it only includes migrations and entities that are related to the module in question.

PR #485, opened 2022-03-08
Implicit migration definition

dfellis:
Part of #450

### DIFF
--- a/src/modules/aws_account@0.0.1/index.ts
+++ b/src/modules/aws_account@0.0.1/index.ts
@@ -4,7 +4,6 @@ import { AWS, } from '../../services/gateways/aws'
 import { AwsAccountEntity, } from './entity'
 import * as allEntities from './entity'
 import { Context, Crud, Mapper, Module, } from '../interfaces'
-import { awsAccount1644463885917, } from './migration/1644463885917-aws_account'
 
 export const AwsAccount: Module = new Module({
   name: 'aws_account',
@@ -74,8 +73,4 @@ export const AwsAccount: Module = new Module({
       }),
     }),
   },
-  migrations: {
-    postinstall: awsAccount1644463885917.prototype.up,
-    preremove: awsAccount1644463885917.prototype.down,
-  },
-});
+}, __dirname);
--- a/src/modules/interfaces.ts
+++ b/src/modules/interfaces.ts
@@ -1,3 +1,5 @@
+import { readdirSync, } from 'fs'
+
 import { QueryRunner, } from 'typeorm'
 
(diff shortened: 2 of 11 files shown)


src/modules/aws_account@0.0.1/index.ts
```diff
@@ -74,8 +73,4 @@ export const AwsAccount: Module = new Module({
       }),
     }),
   },
-  migrations: {
-    postinstall: awsAccount1644463885917.prototype.up,
-    preremove: awsAccount1644463885917.prototype.down,
-  },
-});
+}, __dirname);
```
dfellis, line 76: I would love to get rid of this bit here, but I couldn't figure out a way with the current structure.

src/modules/interfaces.ts
```diff
@@ -253,43 +255,9 @@ export interface ModuleInterface {
...
-    // run your own schema and attach the foreign key references from them to your own table(s), and
-    // when you're removing, you need to remove your foreign key references and your tables before
-    // your dependencies can be properly cleaned up.
-    //
-    // For `upgrade` and `downgrade`, though, it's likely that both `pre` and `post` will be used in
-    // both directions when they are used. Why? If you're upgrading your library and also upgrading
-    // the dependency version at the same time, if any columns you are attached to are changing or
-    // are being removed, you need to run a migration before they are upgraded to detach your
-    // foreign key(s) from them, and then after they are done you need to run a migration to
-    // re-attach to (or otherwise deal with) the new schema.
```
dfellis, line 284: We never used any of the complexity here, and it represents a bunch of footguns for module developers, both from lack of testing on our part and simply from potentially having different schema outcomes depending on which path "through" these various options you took. So dropping it for a much simpler approach.

src/services/iasql.ts
```diff
@@ -775,20 +775,13 @@ ${Object.keys(tableCollisions)
...
-      if (md.migrations?.preinstall) {
-        await md.migrations.preinstall(queryRunner);
-      }
-    }
     for (const md of rootToLeafOrder) {
-      if (md.migrations?.postinstall) {
-        await md.migrations.postinstall(queryRunner);
+      if (md.migrations?.install) {
+        await md.migrations.install(queryRunner);
       }
```
depombo, line 785: Do we have a unit tests for this or a way of knowing if this will continue to work as expected with this change?

dfellis: So yes. We have the `dep-sorting` unit test and then every `module` test installs one-or-more modules during its own testing and would fail otherwise. Also the `simple-integration` test is designed to stress the add db, install modules, uninstall modules, remove db paths.

aguillenv (approved):

PR #486, opened 2022-03-09
Improved migration generation

dfellis:
The second half of #450

The migrations I intend to revert, but they're proof that regenerating it works flawlessly. :D

However, the handwritten stored procedures get blown away. We need to decide what, if anything, we're gonna do about that. Perhaps they can be defined in the module `index.ts` file and automatically inserted on install and removed on uninstall outside of the entity migration stuff? Not sure.


src/modules/aws_account@0.0.1/index.ts
```diff
@@ -4,12 +4,10 @@ import { AWS, } from '../../services/gateways/aws'
...
 import * as allEntities from './entity'
 import { Context, Crud, Mapper, Module, } from '../interfaces'
-import { awsAccount1644463885917, } from './migration/1644463885917-aws_account'
+import * as metadata from './module.json'
 
 export const AwsAccount: Module = new Module({
-  name: 'aws_account',
-  version: '0.0.1',
-  dependencies: [],
+  ...metadata,
```
dfellis, line 10: This can probably also be automated in the `interfaces.ts` file. Not sure if it makes sense to do something like that now or later.

depombo:
The PR looks great! My only comment is if we can update https://github.com/iasql/iasql-engine/blob/main/CONTRIBUTING.md#how-to-develop-iasql in tandem with this PR to get a sense of how things would look like after we land it :)

> However, the handwritten stored procedures get blown away. We need to decide what, if anything, we're gonna do about that. Perhaps they can be defined in the module index.ts file and automatically inserted on install and removed on uninstall outside of the entity migration stuff? Not sure.

I'm hoping to get rid of them within the next couple of days so we should be good with things as-is!

dfellis:
> The PR looks great! My only comment is if we can update https://github.com/iasql/iasql-engine/blob/main/CONTRIBUTING.md#how-to-develop-iasql in tandem with this PR to get a sense of how things would look like after we land it :)
> 
> > However, the handwritten stored procedures get blown away. We need to decide what, if anything, we're gonna do about that. Perhaps they can be defined in the module index.ts file and automatically inserted on install and removed on uninstall outside of the entity migration stuff? Not sure.
> 
> I'm hoping to get rid of them within the next couple of days so we should be good with things as-is!

Oh, yeah, I forgot about that. I'll update it.

CONTRIBUTING.md
```diff
@@ -29,27 +29,19 @@ Development of a new module is expected to follow this pattern:
...
 Development of an existing module is expected to follow this pattern:
 
-1. Remove the entities and migrations of the dependent modules, if any.
-2. Remove the migration of the module you are trying to modify and comment out the usage of this migration in the module index file.
-3. Make the changes to the entities that you want to make.
-4. Run the `yarn gen-module my_existing_module_name` script and have it generate a new migration file. (`my_existing_module_name` needs to match the directory name for the module in question.)
-5. Attach the new migration file to the index file and git restore dependent modules.
+1. Make the changes to the entities that you want to make.
+2. Run the `yarn gen-module my_existing_module_name` script and have it generate a new migration file. (`my_existing_module_name` needs to match the directory name for the module in question.)
+5. Commit the removal of the old migration and add in the new one. You can only have one migration file per module.
```
aguillenv, line 38: nit: in markdown does not matter but for consistency
```suggestion
3. Commit the removal of the old migration and add in the new one. You can only have one migration file per module.
```

aguillenv (approved):

.module_migration.sh
```diff
@@ -22,6 +22,26 @@ psql postgresql://postgres:test@localhost:5432/postgres -c "CREATE DATABASE __ex
...
+# Convert the array of modules into configuration paths for TypeORM
+ENTITIES=""
+MIGRATIONS=""
+NL=$'\n'
+for (( n=0; n < ${#MODARR[*]}; n++))
+do
+  MOD=`echo ${MODARR[n]} | xargs`
+  ENTITIES="${ENTITIES}\"modules/${MOD}/entity/*.ts\",${NL}"
+  MIGRATIONS="${MIGRATIONS}\"modules/${MOD}/migration/*.ts\",${NL}"
+done
```
aguillenv, line 42: this change let us `gen module` without doing the `git rm` of the dependencies, right?

src/modules/aws_account@0.0.1/index.ts
aguillenv: The automation would be just the import of the `metadata`?

dfellis: Yeah. It's not *exactly* being declared twice, but since this now requires a `module.json` in a particular place to generate the migration, why not have the code assume that and not need to import here to re-export to the interface class?

.module_migration.sh
```diff
@@ -22,6 +22,26 @@ psql postgresql://postgres:test@localhost:5432/postgres -c "CREATE DATABASE __ex
...
+
+# Generate the TypeORM config
 cat <<EOF > ormconfig.js
```
depombo, line 45: nit: you could set env variables if you wanted which feels more natural for a bash script https://typeorm.io/#/using-ormconfig/using-environment-variables

CONTRIBUTING.md
depombo, line 38: Would this also be true when add module upgrading?

depombo (approved):

CONTRIBUTING.md
dfellis: So we *should* be treating these as immutable and copying the directory when we really change the schema, so regenerating the schema (like I made possible with this PR) should be rare.

depombo: ah yes, that makes sense!

PR #487, opened 2022-03-09
Run test on random region (except for ec2)

aguillenv:
Ignoring EC2 due to the special casing with AMIs.


.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    branches: []
```
aguillenv, line 11: Temporal! just to avoid test suite interference right now

dfellis: It's running anyways :(

test/modules/aws-account-integration.ts
```diff
@@ -21,7 +22,7 @@ describe('AwsAccount Integration Testing', () => {
   // TODO: Restore some mechanism to verify credentials
   /*it('does not create a test DB with fake credentials', (done) => void iasql.add(
     dbAlias,
-    'us-west-2',
+    getRandomRegion(),
```
dfellis, line 25: You need to do this at the top of the test one time, or it'll do different regions for the different sets of tests and could cause failures

dfellis: Just for this test, afaict

depombo (approved):

depombo (approved):

PR #490, opened 2022-03-09
Simplify the 'entityId' functions to follow the same pattern everywhere

dfellis:


depombo (approved):

src/modules/aws_ec2@0.0.1/index.ts
```diff
@@ -36,7 +36,7 @@ export const AwsEc2Module: Module = new Module({
   mappers: {
     instance: new Mapper<Instance>({
       entity: Instance,
-      entityId: (i: Instance) => i.instanceId ?? i.name,
+      entityId: (i: Instance) => i.instanceId ?? i.id?.toString() ?? '',
```
aguillenv, line 39: when it could not be an i.id?

src/modules/aws_cloudwatch@0.0.1/index.ts
```diff
@@ -26,7 +26,7 @@ export const AwsCloudwatchModule: Module = new Module({
   mappers: {
     logGroup: new Mapper<LogGroup>({
       entity: LogGroup,
-      entityId: (e: LogGroup) => e?.logGroupName,
+      entityId: (e: LogGroup) => e.logGroupName ?? e.id,
```
aguillenv, line 29: Seen a pattern in other modules, shouldn't this be `e.id.toString()`?

src/modules/aws_ecr@0.0.1/index.ts
```diff
@@ -245,7 +245,7 @@ export const AwsEcrModule: Module = new Module({
     }),
     repositoryPolicy: new Mapper<AwsRepositoryPolicy>({
       entity: AwsRepositoryPolicy,
-      entityId: (e: AwsRepositoryPolicy) => e.repository?.repositoryName + '',
+      entityId: (e: AwsRepositoryPolicy) => e.repository?.repositoryName + '' ?? e.id.toString(),
```
aguillenv, line 248: that `+ ''` seems like a mistake?

src/modules/aws_ecs_fargate@0.0.1/index.ts
```diff
@@ -145,7 +145,7 @@ export const AwsEcsFargateModule: Module = new Module({
   mappers: {
     cluster: new Mapper<AwsCluster>({
       entity: AwsCluster,
-      entityId: (e: AwsCluster) => e?.clusterArn ?? 'default',
+      entityId: (e: AwsCluster) => e.clusterArn ?? e.id?.toString() ?? '',
```
aguillenv, line 148: why do we default to `''`?

src/modules/aws_ec2@0.0.1/index.ts
dfellis: Only when it comes from AWS, but then `instanceId` is guaranteed to exist, but I can't convince Typescript of that.

src/modules/aws_cloudwatch@0.0.1/index.ts
dfellis: Ahh, so I think it's actually not doing anything here. The output type *must* be a string, but since `logGroupName` is a required field, Typescript believes it'll never fall back to `e.id` and didn't complain about the mismatched type of it.

src/modules/aws_ecs_fargate@0.0.1/index.ts
dfellis: Because I can't convince Typescript that the cloud-derived cluster entity is guaranteed to have the `e.clusterArn` when it won't have an `id` and vice-versa for the DB-derived entity.

src/modules/aws_ecr@0.0.1/index.ts
dfellis: Yeah, that'd probably turn into an `'undefined'` string. I can remove it.

PR #492, opened 2022-03-10
display better error message when cli version is undefined

depombo:


src/router/index.ts
```diff
@@ -28,9 +28,9 @@ v1.use(express.json({ limit: '10000MB' }));
 v1.use((req, res, next) => {
   const headers = req.headers;
   const cliVersion = headers["cli-version"] as string;
-  if (semver.compare(MIN_CLI_VERSION, cliVersion) > 0) {
+  if (cliVersion === undefined || semver.compare(MIN_CLI_VERSION, cliVersion) > 0) {
```
aguillenv, line 31: nit: maybe compare with falsy in case someone pass `null` or `''`?
```suggestion
  if (!cliVersion || semver.compare(MIN_CLI_VERSION, cliVersion) > 0) {
```

aguillenv (approved):

PR #495, opened 2022-03-10
Switch to a declarative entity ID by default

### LINKED ISSUE #476: Restore declarative entity identification

dfellis:
Resolves #476

I couldn't make it 100% because one of the entity IDs is based on the unique ID of the entity it joins (which is really weird), but also I realized I would not be able to automate the DB mapper for that entity anyways because it's `read` function is similarly weird. So I'm keeping `entityId` around as a function that is automatically populated if an override is not provided.

### DIFF
--- a/src/modules/aws_account@0.0.1/entity/aws_account.ts
+++ b/src/modules/aws_account@0.0.1/entity/aws_account.ts
@@ -4,11 +4,14 @@ import {
   PrimaryGeneratedColumn,
 } from 'typeorm'
 
+import { cloudId, } from '../../../services/cloud-id'
+
 @Entity({
   name: 'aws_account',
 })
 export class AwsAccountEntity {
   @PrimaryGeneratedColumn()
+  @cloudId
   id: number;
 
   @Column()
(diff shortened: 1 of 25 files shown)


src/modules/aws_account@0.0.1/entity/aws_account.ts
```diff
@@ -4,11 +4,14 @@ import {
...
 } from 'typeorm'
 
+import { cloudId, } from '../../../services/cloud-id'
+
 @Entity({
   name: 'aws_account',
 })
 export class AwsAccountEntity {
   @PrimaryGeneratedColumn()
+  @cloudId
```
depombo, line 14: I guess this never goes in the cloud, but we have to pick one?

depombo (approved):

src/modules/aws_account@0.0.1/entity/aws_account.ts
dfellis: Yeah. If I want to not provide a custom `entityId` I have to pick a column. That's all there was to this, basically.

PR #496, opened 2022-03-10
generate a db name when one is not provided

depombo:


src/services/iasql.ts
```diff
@@ -24,13 +24,20 @@ export function recordCount(records: { [key: string]: any, }[]): [number, number
   return [ dbCount, cloudCount, bothCount, ];
 }
 
+// config for unique name generator
+// e.g. big_red_donkey
```
dfellis, line 28: purple monkey dishwasher

dfellis (approved):
I don't see a problem with this, but couldn't the CLI do this and not require a change in both?

depombo:
> I don't see a problem with this, but couldn't the CLI do this and not require a change in both?

I did this in case we ever need to do name reservation (right now it just needs to be unique per user account) and check other requirements Postgres/IP might have for the name. We did something similar with any cloud because azure was super picking about resource names and so on.

src/services/iasql.ts
depombo: 😂

PR #499, opened 2022-03-10
Support optional data-only flag to export

depombo:


src/services/iasql.ts
```diff
@@ -174,7 +174,10 @@ export async function dump(dbAlias: string, user: any) {
   const pgUrl = `postgres://${encodeURIComponent(config.dbUser)}:${encodeURIComponent(
     config.dbPassword
   )}@${config.dbHost}/${dbId}`;
-  const { stdout, } = await exec(`pg_dump --inserts -x ${pgUrl}`, { shell: '/bin/bash', });
+  const { stdout, } = await exec(
+    `pg_dump ${dataOnly ? '--data-only --column-inserts, --rows-per-insert=50 --exclude-table-data=aws_account --on-conflict-do-nothing' : ''} --inserts -x ${pgUrl}`,
```
dfellis, line 178: Nit: the `--inserts` should be in the `: ''` section since it's actually a different directive from `--column-inserts`

```suggestion
    `pg_dump ${dataOnly ? '--data-only --column-inserts, --rows-per-insert=50 --exclude-table-data=aws_account --on-conflict-do-nothing' : '--inserts'}  -x ${pgUrl}`,
```

dfellis (approved):

src/services/iasql.ts
depombo: it is right after, it is hard to see though `pg_dump ${...} --inserts -x ${pgUrl}`,

PR #508, opened 2022-03-11
add missing subnet az to enum

depombo:


src/modules/aws_vpc@0.0.1/entity/index.ts
```diff
@@ -126,6 +126,7 @@ export enum AvailabilityZone {
   us_east_2b = 'us-east-2b',
   us_east_2c = 'us-east-2c',
   us_west_1a = 'us-west-1a',
+  us_west_1b = 'us-west-1b',
```
aguillenv, line 129: I guess it makes sense but are we sure this is a valid availability zone and now a weird thing from AWS where they have zone a and c but not b? 😅

aguillenv (approved):

src/modules/aws_vpc@0.0.1/entity/index.ts
dfellis: I noticed it missing when I generated the list, but I assumed that the whole set did actually arrive and let it be. If that query for all zones didn't actually return all zones then I'm a bit concerned.

dfellis (approved):
We can try with this in there and see what happens?

src/modules/aws_vpc@0.0.1/entity/index.ts
depombo: how did you generate the list? maybe it was added / re-added since you generated it?

dfellis: I used the Node REPL with the AWS gateway and my AWS account. I don't have the exact code anymore, but it was basically:

1. `getRegions` and map to array of region names.
2. for each region, getAvailabilityZones, and map to array of availability zone names
3. sort

Done in old school promise style, passing a callback to `.then` because it's really awkward to do async/await in a REPL due to the "top level scope in JS cannot `await`" BS.

PR #509, opened 2022-03-11
Refactor ECS container definition

### LINKED ISSUE #488: AWS allows an ACTIVE ECS task on a nonexistent ECR
> Our current model for container definition has "4" fields to be able tor retrieve the image. Those are: docker_image, repository_id or public_repository_id (just one of them could have something at a time) + tag
> 
> The problem with this is that we have found that AWS let you have an ACTIVE task definition with its container definition pointing to an unexisting ECR, for example. So this fails trying to sync/install. 
> Another problem that I've just realized and did not take into account is that we can have a container pointing to a public ECR thats not in the account. right now that would fail because it will not find it.
> 
> So I have thought in 3 possible options:
> 1. We just remove the complexity of trying to figure out if we need to relate to a ECR and just have a image field that is the string image like <image/image-uri>:<tag>. Right now we have to do some processing to understand the container image and look for the ECR manually. For AWS the image field is just that, a string.
> 2. Option 1 but we do keep the ECR FKs (priv or pub) but just as a reference to be able to make a relation.
> 3.  We leave it as we have it and if we are not able to find the ECR (at least the private one) we ignore that container. I'm not sure how we would hanlde the public ECRs that are not from the account in this case.

from @aguillenv

aguillenv:
Resolves #488 

- Try to link with ECR, if not possible, default to image `string` field
- Handle images with `digest`
- Speed up ECS test suite by reusing the same ELB components
- Update ELB migration (Residual changes, probably from a manual update?)


test/modules/aws-ecs-integration.ts
```diff
@@ -97,43 +122,35 @@ describe('ECS Integration Testing', () => {
...
+        INSERT INTO aws_container_definition ("name", image, digest, essential, memory_reservation, host_port, container_port, protocol, env_variables, task_definition_id, log_group_id)
+        VALUES('${containerNameDigest}', '${image}', '${imageDigest}', false, ${containerMemoryReservation}, ${hostPort + 2}, ${containerPort + 2}, '${protocol}', '{ "test": 2}', (select id from aws_task_definition where family = '${tdFamily}' and status is null limit 1), (select id from log_group where log_group_name = '${logGroupName} limit 1'));
+      COMMIT;
     `));
 
     it('check container definition insertion', query(`
       SELECT *
       FROM aws_container_definition
-      WHERE name = '${containerName}' AND docker_image = '${image}' AND tag = '${imageTag}';
+      WHERE name = '${containerName}' AND image = '${image}';
```
depombo, line 138: this is the one without the type safety and was causing the error, right?

depombo (approved):

test/modules/aws-ecs-integration.ts
aguillenv: hmm I do not follow the question. The issue was that when no repository found it was thawing an [error](https://github.com/iasql/iasql-engine/pull/509/files#diff-5ec517f3f47bf40ef7dea8ec93f031da5633f9ed205c9322c8491046a4752b4cL263). Now it just try to get the related ECR but if it is not found, it just save the value in the `image` field as a `string`.

src/modules/aws_ecs_fargate@0.0.1/entity/aws_container_definition.ts
```diff
@@ -19,7 +19,10 @@ export enum TransportProtocol {
   UDP = "udp"
 }
 
-@Check(`"docker_image" is not null or "repository_id" is not null  or "public_repository_id" is not null`)
+// `image` > `repository` > `publicRepository`
+// `digest` > `tag` > null
+@Check(`("image" is null and ("repository_id" is not null or "public_repository_id" is not null)) or "image" is not null`)
+@Check(`("image" is null and ("tag" is not null or "digest" is not null)) or "image" is not null`)
```
dfellis, line 25: With these as two separate checks won't it cause a failure outright on `"image" is null`? Because the first one could pass but the second would not and vice versa?

aguillenv: Hmm what case do you have in mind? I do not see it fail? 
If `image is null` It is checking that an ECR exists, if an ECR exists a `tag` or `digest` have to exists.
If `image is not null` both pass

aguillenv: Oh there's a case where someone define the container with an ECR but no `tag` or `digest` and it would fail. Let me review that!

dfellis: So, I thought you mentioned some sort of "default tag" yesterday? Is that not the case?

The other failure direction I thought I saw with the first check I now see wouldn't happen, though.

aguillenv: Ok so it would be just remove the second check. It is not really necessary, we keep `tag` and `digest` as nullable/optional values.

dfellis: Wait, keep it, but just make it such that you can't have both at the same time, but you can have neither?

dfellis:
```suggestion
@Check(`("image" is null and ("repository_id" is not null or "public_repository_id" is not null)) or "image" is not null`)
@Check(`("image" is null and (("tag" is not null and "digest" is null) or ("tag" is null and "digest" is not null) or ("tag" is null and "digest" is null)) or "image" is not null`)
```

aguillenv: Oh well, I was being permissive and just have this around (and eventually added in the docs?) `digest` > `tag` > null meaning that if both defined, `digest` will be taken into account first since it is more specific. Same for ECR and `image` fields being defined at the same time, `image` > `repository` > `publicRepository`. I know this is opinionated, we can defined as we want.

test/modules/aws-ecs-integration.ts
depombo: I mean the error you found in the production logs after the call earlier this week

src/modules/aws_ecs_fargate@0.0.1/entity/aws_container_definition.ts
aguillenv: > Suggested change 
> @Check(`("image" is null and ("repository_id" is not null or "public_repository_id" is not null)) or "image" is not null`)
> @Check(`("image" is null and ("tag" is not null or "digest" is not null)) or "image" is not null`)
> @Check(`("image" is null and ("repository_id" is not null or "public_repository_id" is not null)) or "image" is not null`)
> @Check(`("image" is null and (("tag" is not null and "digest" is null) or ("tag" is null and "digest" is not null) or ("tag" is null and "digest" is null)) or "image" is not null`)

If we want this, the `image` check still need to go, because you can have a ECR uri only apparently, with no tag or digest. It would be something like: 
```suggestion
@Check(`("image" is null and ("repository_id" is not null or "public_repository_id" is not null)) or "image" is not null`)
@Check(`("tag" is not null and "digest" is null) or ("tag" is null and "digest" is not null)`)
```

dfellis: That's missing the path where they are both null, though?

test/modules/aws-ecs-integration.ts
aguillenv: oh, it is this:
> The issue was that when no repository found it was throwing an [error](https://github.com/iasql/iasql-engine/pull/509/files#diff-5ec517f3f47bf40ef7dea8ec93f031da5633f9ed205c9322c8491046a4752b4cL263).

src/modules/aws_ecs_fargate@0.0.1/entity/aws_container_definition.ts
aguillenv: indeed 
```suggestion
@Check(`("image" is null and ("repository_id" is not null or "public_repository_id" is not null)) or "image" is not null`)
@Check(`("tag" is null and "digest" is null) or ("tag" is not null and "digest" is null) or ("tag" is null and "digest" is not null)`)
```

dfellis (approved):

PR #511, opened 2022-03-11
Add install/uninstall test suite

### LINKED ISSUE #510: test `install`/`uninstall` Including interactions between installing and uninstalling other modules

aguillenv:
Resolves #510


src/services/iasql.ts
```diff
@@ -706,9 +706,12 @@ export async function modules(all: boolean, installed: boolean, dbAlias: string,
   }
 }
 
-export async function install(moduleList: string[], dbAlias: string, user: any) {
+export async function install(moduleList: string[], dbAlias: string, user: any, allModules = false) {
```
depombo, line 709: this could be an optional flag in the cli too `iasql install --all`

depombo:
rather than be in their own nested test suite, shouldn't these commands be sandwiched in between creates and deletes?

aguillenv:
> rather than be in their own nested test suite when there are little to no resources in the account, shouldn't these commands be sandwiched in between creates and deletes?

That wouldn't be just the `sync` test? I was going to create a second PR adding `sync`s after some `insert`/`updates`

aguillenv:
> > That wouldn't be just the `sync` test? I was going to create a second PR adding `sync`s after some `insert`/`updates`
> 
> `sync` after any `insert`/`update`/`delete` and before an `apply` will clear the staged changes / undo the queries which is a good test which we need have, but it has very different behavior compared to `uninstall`/`install` after `insert`/`updates` which makes sure it can re-import what is currently in the account. Doing an `uninstall`/`install` after `delete`s when the account is mostly empty is basically a no-op though?
> 
> in terms of testing `sync` for properly importing new infra (testing it for undoing `insert `/`update`/`delete` queries before an `apply` can be done in the CI), we could do it in the staging job by connecting a second db to the same staging account so: connect two dbs to the same account, run apply on one db and then run `sync` on the other one? [iasql/iasql-on-iasql#20](https://github.com/iasql/iasql-on-iasql/issues/20)

For me a `sync` after an `apply` would be the same as `uninstall/install. Agree that the `sync` test have to be with 2 dbs and add something in one and `sync` in the other.

What I'll do is
- Leave this isolated suites to catch possible relation/schema issues between modules on `install`/`uninstall`.
- Add `uninstall`/`install` after `insert`/`update`'s applies.
- A follow up PR with a test in `common` testing `sync` using 2 db connections.

depombo:
> 
> For me a `sync` after an `apply` would be the same as `uninstall/install`. Agree that the `sync` test have to be with 2 dbs and add something in one and `sync` in the other.
> 

I suppose they effectively are. It is just a different code path so I see your point of doing both one after the other

> What I'll do is
> 
> * Leave this isolated suites to catch possible relation/schema issues between modules on `install`/`uninstall`.
> * Add `uninstall`/`install` after `insert`/`update`'s applies.

👍

> * A follow up PR with a test in `common` testing `sync` using 2 db connections.

https://github.com/iasql/iasql-on-iasql/pull/21 covers this. we are however missing the `sync` after any `insert`/`update`/`delete` and before an `apply` will clear the staged changes / undo the queries ?

src/services/iasql.ts
dfellis: Would very quickly become a bad idea as the number of modules balloons, though.

depombo (approved):

PR #520, opened 2022-03-14
ECS Service force new deployment

### LINKED ISSUE #518: Add ECS service force update

aguillenv:
Resolves #518 

image

image


depombo:
> > should we add a test for it?
> 
> We could add a tests that if `true` after the successful `apply` it goes back to `false`. Or do you have something more elaborate in mind?

Yeah, that sounds good.

Issue #469, opened 2022-03-05
Cron job for scheduling changes and CLI-like commands in PostgreSQL


depombo:
Something like https://github.com/graphile/worker could make this happen without changes to our infra

depombo:
> Something like https://github.com/graphile/worker could make this happen without changes to our infra

I played around with this. 
https://github.com/iasql/iasql-engine/compare/scheduler?expand=1

The problem is that it spins up a worker process per db, so I think we would very quickly hit a wall with it?

depombo:
> > Something like https://github.com/graphile/worker could make this happen without changes to our infra
> 
> I played around with this. https://github.com/iasql/iasql-engine/compare/scheduler?expand=1
> 
> The problem is that it spins up a worker process per db, so I think we would very quickly hit a wall with it?

Nvm it actually does not spin a worker per process

Issue #459, opened 2022-03-02
Safer `sync` and faster `apply`


depombo:
Fire triggers on `INSERT`s and `UPDATE`s to keep track of changes in some sort of system table. `apply` goes off these changes and `sync` won't run if there are pending changes unless there is some sort of `force` flag.

depombo:
Depends on #469

depombo:
Probably something like
- [ ] Triggers on all our tables that inserts into a `staged_changes` table https://dba.stackexchange.com/questions/241173/postgresql-trigger-to-keep-track-of-table-changes
- [ ] Have the `iasql` service in the engine go off this table

depombo:
> * [ ]  Triggers on all our tables that inserts into a `iasql_staged_changes` table https://dba.stackexchange.com/questions/241173/postgresql-trigger-to-keep-track-of-table-changes

Maybe we can do something like this to automate this https://www.githubstatus.com

depombo:
We should keep this reddit discussion in mind https://www.reddit.com/r/devops/comments/u3b3wf/comment/i4o89s9/. Apparently Pulumi by default does not check the state of the cloud as this issue is suggesting we do, while Terraform does.

We need this for things like ASG and Spot in EC2, but perhaps we might not want to rely on this for that to work in case the user always wants to use the `force`/`refresh` flag all the time?

dfellis:
> We should keep this reddit discussion in mind https://www.reddit.com/r/devops/comments/u3b3wf/comment/i4o89s9/. Apparently Pulumi by default does not check the state of the cloud as this issue is suggesting we do too, while Terraform always does.
> 
> We need this for things like #960 in EC2, but perhaps we might not want to rely on this for that to work in case the user always wants to use the `force`/`refresh` flag all the time?

I think this is a partial misunderstanding of what we intend to do here? We still need to check the cloud state for each record we saw changed, which is equivalent to Terraform. What we currently do is a full check of everything in their cloud every time, which this eliminates.

depombo:
Ah! I think I get it 👍

PR #521, opened 2022-03-14
Scheduler using `graphile-worker`

depombo:
Resolves https://github.com/iasql/iasql-engine/issues/469. Every db gets its own scheduler with their own db connection and reuses it for operations done within it. It also refactors the error handling function into an initial logger class and the entities used in the Typeorm connection.


src/services/iasql.ts
```diff
@@ -127,9 +130,10 @@ export async function remove(dbAlias: string, user: any) {
   let conn;
   try {
     const { dbId, dbUser } = await dbMan.getMetadata(dbAlias, user);
+    scheduler.stop(dbId);
     conn = await createConnection(dbMan.baseConnConfig);
     await conn.query(`
-      DROP DATABASE ${dbId} WITH (FORCE);
+      DROP DATABASE ${dbId};
```
depombo, line 136: this was causing connections to be closed forcefully which made the scheduler unhappy. Here we wait for clean up after itself gracefully.

src/services/iasql.ts
```diff
@@ -127,9 +130,10 @@ export async function remove(dbAlias: string, user: any) {
   let conn;
   try {
     const { dbId, dbUser } = await dbMan.getMetadata(dbAlias, user);
+    scheduler.stop(dbId);
     conn = await createConnection(dbMan.baseConnConfig);
     await conn.query(`
-      DROP DATABASE ${dbId} WITH (FORCE);
```
aguillenv, line 132: Are we sure this is not necessary?

src/modules/aws_account@0.0.1/migration/1644463885917-aws_account.ts
```diff
@@ -5,9 +5,31 @@ export class awsAccount1644463885917 implements MigrationInterface {
...
+        await queryRunner.query(`
+            create or replace procedure iasql_schedule_sync()
+            language plpgsql
+            as $$
+            begin
+                PERFORM graphile_worker.add_job('scheduleSync');
+                raise notice 'Scheduled sync';
+            end;
+            $$;
+        `)
```
depombo, line 27: once we have https://github.com/iasql/iasql-engine/issues/459, we can make add/replace for a blocking version that could calls `pg_sleep` and outputs something to `stdout`

src/services/iasql.ts
aguillenv: Already got a response

depombo: hahahah

src/modules/aws_account@0.0.1/migration/1644463885917-aws_account.ts
```diff
@@ -5,9 +5,31 @@ export class awsAccount1644463885917 implements MigrationInterface {
...
+                raise notice 'Scheduled apply';
+            end;
+            $$;
+        `);
+        await queryRunner.query(`
+            create or replace procedure iasql_schedule_sync()
+            language plpgsql
+            as $$
+            begin
+                PERFORM graphile_worker.add_job('scheduleSync');
```
aguillenv, line 23: how does the bd know the existence of `graphile_worker.*`?

src/services/scheduler.ts
```diff
@@ -0,0 +1,48 @@
...
+const workerShutdownEmitter = new EventEmitter();
+
+// graphile-worker here functions as a library, not a child process.
+// It manages its own database schema
+// (graphile_worker) and migrations in each user db using our credentials
+export async function start(dbAlias: string, dbId: string, user: any) {
+  // Run a worker to execute jobs:
+  const runner = await run({
+    // TODO pass a connection pool here instead of telling it to create a new one
+    connectionString: dbMan.ourPgUrl(dbId),
```
aguillenv, line 17: if it is being started with out connection string, how it will `apply` users changes?

test/modules/aws-account-integration.ts
```diff
@@ -41,9 +41,17 @@ describe('AwsAccount Integration Testing', () => {
     VALUES ('fake', 'creds', 'us-west-2')
   `));
 
-  it('does absolutely nothing when you apply this', apply);
+  // it('does absolutely nothing when you apply this', apply);
+
+  it('does absolutely nothing when you apply this', query(`
+    CALL iasql_schedule_apply();
+  `));
```
aguillenv, line 48: Should we be testing `CALL iasql_schedule_sync();` too?

depombo: yeah, I can call it even if it is a noop

src/modules/aws_account@0.0.1/migration/1644463885917-aws_account.ts
depombo: it doesn't, we guarantee the schema will be there by starting the scheduler (which does it's own migrations) before we run our own migrations https://github.com/iasql/iasql-engine/pull/521/files#diff-c8f7acf4155d692e9663c0ae07f0ab80ceb39ab0c6d7e15d9a2e987aa1a04ac9R55

aguillenv: oh ok, so the 'graphile-worker' run method creates the schema and the procedure add_job

src/services/scheduler.ts
depombo: yeah, I'm removing that comment since the connection that starts the scheduler dies with the request so it doesn't make sense

src/modules/aws_account@0.0.1/migration/1644463885917-aws_account.ts
depombo: yes!

src/router/module.ts
```diff
@@ -50,7 +51,7 @@ mod.post('/install', async (req, res) => {
   try {
     res.json(await iasql.install(list, dbAlias, req.user));
   } catch (e: any) {
-    res.status(400).json(e.message);
+    res.status(500).json(logger.error(e));
```
depombo, line 54: this is what it always should have been?

dfellis: 400s are for user error, 500s are for server errors. It was a 400 here because failure to install was *expected* to always be because they're trying to install something that doesn't exist and not because installation itself fails.

depombo: Well, it could be a bug on our end. I do get your point and will just leave as 500

aguillenv (approved):

PR #536, opened 2022-03-16
Cleanup after tests

### LINKED ISSUE #473: "Auto-vacuum" task to clean up test accounts over the weekend
Since we can't rely on the state of IaSQL in development itself to do this, we should consider using something like: https://github.com/rebuy-de/aws-nuke

aguillenv:
Resolves #473


test/helpers.ts
```diff
@@ -65,3 +66,58 @@ export function runQuery(dbAlias: string, queryString: string, assertFn?: (res:
...
 }
+
+async function cleanDB(modules: string[] = []): Promise<void> {
+  const dbAlias = `cleandb${Date.now()}`;
+  const awsRegion = /**region ?? */ process.env.AWS_REGION ?? 'barf';
+  console.log(`Cleaning ${dbAlias} in ${awsRegion}...`);
+  await iasql.add(dbAlias, awsRegion, process.env.AWS_ACCESS_KEY_ID ?? 'barf', process.env.AWS_SECRET_ACCESS_KEY ?? 'barf', 'not-needed');
+  console.log('DB created...');
+  await iasql.install(modules, dbAlias, 'not-needed');
+  console.log(`Modules ${modules} installed...`);
```
depombo, line 77: This works, but I think what would speed up the tests here is to leave it up to after hook to spin down the database so we don't have to spin up a new one and install modules again?

aguillenv: even though that would speed up the test, I'm not sure if it makes sense to spin up a new one every time to ensure a clean run on different describes? I do not have a strong opinion on this to be honest

depombo: hmm I'm not sure I'm explaining myself / getting my point across since it wouldn't entail a spin down of the database on different jest `describe`s? what I am saying is actually not leaving it up to the `describe` to delete the db (which it is right now) to then have the `after` hook spin it up again, but instead leave it up to the `after` hook to ensure that it is clean and then delete it? I also don't have a particularly strong opinion, just whatever is simplest

aguillenv: Updating my comment with the typo, and as talked offline both have valid points on this (test speed up vs clean db for each suite). Lets leave it as is for now.

test/helpers.ts
```diff
@@ -8,7 +8,8 @@ export function execComposeUp() {
   execSync('cd test && docker-compose up -d && sleep 5');
 }
 
-export function execComposeDown() {
+export async function execComposeDown(input: { modules?: string[], region?: string }) {
```
depombo, line 11: curious, why an object and not just two optional parameters?

depombo (approved):

test/helpers.ts
aguillenv: I was having an issue with the second argument, but I think I know how to solve it now

PR #540, opened 2022-03-18
fix missing dependency determination

depombo:


src/services/iasql.ts
```diff
@@ -713,11 +713,12 @@ export async function install(moduleList: string[], dbAlias: string, user: any,
     }`);
   }
   // Check to make sure that all dependent modules are in the list
-  const missingDeps = mods.map((m: Modules.ModuleInterface) => m.dependencies.find(d => !moduleList.includes(d)));
```
depombo, line 716: `find` only returns the first value that matches the predicate so this wasn't returning a full list

aguillenv: hmm curious how it was working or why it was done with the `find`?

aguillenv (approved):

src/services/iasql.ts
depombo: it was always returning `aws_account` which was the first match and that was then filtered so it was a always a noop

src/services/iasql.ts
```diff
@@ -731,6 +723,16 @@ export async function install(moduleList: string[], dbAlias: string, user: any,
       i--;
     }
   }
+  // TODO rm special casing for aws_account
+  // Check to make sure that all dependent modules are in the list
+  const missingDeps = mods
+    .flatMap((m: Modules.ModuleInterface) => m.dependencies.filter(d => !moduleList.includes(d) && !existingModules.includes(d)))
```
depombo, line 729: added the check for `existingModules`

aguillenv: meaning the ones are already installed, right?

depombo: yes

aguillenv (approved):

Issue #539, opened 2022-03-18
Bootstrap a graphile worker for each existing db on server startup


depombo:
There is one graphile worker given to each db on `connect`/`new` which gets wiped when the server / engine deploys even if the `graphile_worker` schema remains. The `scheduler.ts` class that houses the graphile worker needs  a `bootstrap` method that is called on server startup to spin up a worker for every db that it is already managing. Currently this could be done with queries to the base connection since we just need the `dbId` and `dbUser` at the bare minimum to bootstrap. This should also be easier to do after https://github.com/iasql/iasql-engine/issues/543

PR #541, opened 2022-03-18
add plan sp and overload (un)install sp

depombo:
add `iasql_plan` and overload the `iasql_install` sp so that it can also be called with just a `text` argument. without this one needs to always pass an array even if it only has one `text` element in it.
before: `call iasql_install(array['aws_vpc@0.0.1'])`
now: `call iasql_install('aws_vpc@0.0.1')`

### DIFF
--- a/src/migration/1647526647810-init.ts
+++ b/src/migration/1647526647810-init.ts
@@ -128,8 +128,10 @@ export class init1647526647810 implements MigrationInterface {
         await queryRunner.query(`DROP PROCEDURE "iasql_apply"`);
         await queryRunner.query(`DROP PROCEDURE "iasql_plan"`);
         await queryRunner.query(`DROP PROCEDURE "iasql_sync"`);
-        await queryRunner.query(`DROP PROCEDURE "iasql_install"`);
-        await queryRunner.query(`DROP PROCEDURE "iasql_uninstall"`);
+        await queryRunner.query(`DROP PROCEDURE "iasql_install(_mod text)"`);
+        await queryRunner.query(`DROP PROCEDURE "iasql_install(_mods text[])"`);
+        await queryRunner.query(`DROP PROCEDURE "iasql_uninstall(_mod text)"`);
+        await queryRunner.query(`DROP PROCEDURE "iasql_uninstall(_mods text[])"`);
         await queryRunner.query(`ALTER TABLE "iasql_dependencies" DROP CONSTRAINT "FK_7dbdaef2c45fdd0d1d82cc9568c"`);
         await queryRunner.query(`ALTER TABLE "iasql_dependencies" DROP CONSTRAINT "FK_9732df6d7dff34b6f6a1732033b"`);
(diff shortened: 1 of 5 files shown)


src/migration/1647526647810-init.ts
```diff
@@ -94,11 +112,21 @@ export class init1647526647810 implements MigrationInterface {
...
+        `);
     }
 
     public async down(queryRunner: QueryRunner): Promise<void> {
         await queryRunner.query(`DROP PROCEDURE "until_iasql_operation"`);
         await queryRunner.query(`DROP PROCEDURE "iasql_apply"`);
+        await queryRunner.query(`DROP PROCEDURE "iasql_plan"`);
         await queryRunner.query(`DROP PROCEDURE "iasql_sync"`);
         await queryRunner.query(`DROP PROCEDURE "iasql_install"`);
         await queryRunner.query(`DROP PROCEDURE "iasql_uninstall"`);
```
aguillenv, line 132: I think these statements will fail due to ambiguity since you overload the procedures. You have to add specific `DROP` calls for them.

test/modules/aws-account-integration.ts
```diff
@@ -45,6 +45,10 @@ describe('AwsAccount Integration Testing', () => {
 
   it('does absolutely nothing when you sync this', sync());
```
dfellis, line 46: Should these be moved from the direct function call mechanism to the `query`-based version?

depombo: this is a shorthand for that defined in `helpers`

dfellis: Oh, we changed the helper instead of deleting it? Hmm... We did that before because invoking the iasql.ts file directly was a little awkward, but now if we switch to 100% `query`-based testing the tests themselves can educate readers on how to use IaSQL.

We could make that a follow-up task?

depombo: > Oh, we changed the helper instead of deleting it? 

Yes

> Hmm... We did that before because invoking the iasql.ts file directly was a little awkward, but now if we switch to 100% `query`-based testing the tests themselves can educate readers on how to use IaSQL. We could make that a follow-up task?
 
What you are describing is already the case. The reason none of the quickstart or documentation reference has been added yet is because of https://github.com/iasql/iasql-engine/issues/539, but that will be fairly trivial to do after https://github.com/iasql/iasql-engine/pull/542

dfellis: I meant having the test suite itself also act as documentation by always using `query(...` instead of the existing helpers.

depombo: oh, sure I see your point. hmm eventually it will be all over the documentation so it doesn't seem super necessary especially since they can look at the declaration in the `helper`? the `install`-ing of multiple modules doesn't look great to put all over the place with the query replacement https://github.com/iasql/iasql-engine/blob/main/test/helpers.ts#L40 ? but otherwise I'm game for it

depombo: created https://github.com/iasql/iasql-engine/issues/544

aguillenv (approved):

dfellis (approved):

PR #542, opened 2022-03-19
Replace IP with PG metadata db

depombo:
Introduces a singleton `MetadataRepository` that manages a long-lived `iasql_metadata` database bootstrapped when the `engine` starts. The entities and migrations live in `src/metadata`. `iasql` service takes the user ID directly instead of the entire `req.user` and `dbMan` provides default values for dbs when authentication is disabled.


src/metadata/entity/index.ts
```diff
@@ -0,0 +1,42 @@
...
+@Entity()
+export class IasqlUser {
+  // auth0 generated id
+  @PrimaryColumn()
+  id: string;
+
+  @Column()
+  email: string;
+
+  @ManyToMany(() => IasqlDatabase, {
```
depombo, line 35: made this many-to-many from the get-go for when we eventually want to have multiple users access the same database, but technically every database is associated with a single user atm

ormconfig-metadata.js
```diff
@@ -0,0 +1,28 @@
+const { SnakeNamingStrategy } = require('typeorm-naming-strategies');
```
aguillenv, line 1: Out of curiosity, why do we need this metadata config file? could not live in the current form config file?

aguillenv (commented):
What I do not follow is the persistence of the tables after a restart or something? Or will they be part of the user database schemas? If so, they should be in a different schema and with some restrictions, right?

ormconfig-metadata.js
dfellis: The "main" TypeORM migrations are for the tables we automatically put into every client database, while this is now metadata for the engine itself to manage the databases in its own private database, so we don't want those tables to show up in every user's databases.

ormconfig-metadata.js
```diff
@@ -0,0 +1,28 @@
...
+
+module.exports = {
+ "type": "postgres",
+ "host": "localhost",
+ "port": 5432,
+ "username": "postgres",
+ "password": "test",
+ "database": "__example__",
```
dfellis, line 9: Do we want to use a different database name here?

src/router/db.ts
```diff
@@ -14,13 +15,18 @@ db.post('/new', async (req, res) => {
     ].filter(k => !req.body.hasOwnProperty(k)).join(', ')}`
   );
   try {
-    res.json(await iasql.add(dbAlias, awsRegion, awsAccessKeyId, awsSecretAccessKey, req.user));
+    res.json(
+      await iasql.add(
+        dbAlias, awsRegion, awsAccessKeyId, awsSecretAccessKey, dbMan.getUid(req.user), dbMan.getEmail(req.user)
```
dfellis, line 20: How are `dbMan.getUid` and `dbMan.getEmail` synchronous APIs?

src/services/db-manager.ts
```diff
@@ -137,59 +123,14 @@ export function ourPgUrl(dbId: string): string {
...
-  const dbUser = metadata[toUserKey(dbAlias)];
-  delete metadata[toDbKey(dbAlias)];
-  delete metadata[toUserKey(dbAlias)];
-  await IronPlans.setTeamMetadata(ipUser.teamId, metadata);
-  return { dbId, dbUser, dbAlias };
-}
+// TODO type user
+export function getUid(user: any): string {
+  return config.a0Enabled ? user.sub : 'iasql';
+}
```
dfellis, line 136: Oh, these are literally the Auth0 values, that's how it can be synchronous.

src/services/repositories/metadata.ts
```diff
@@ -0,0 +1,86 @@
...
+  private conn: Connection;
+  private userRepo: Repository<IasqlUser>;
+  private dbRepo: Repository<IasqlDatabase>;
+
+  async init() {
+    const conn = await createConnection(dbMan.baseConnConfig);
+    try {
+      await conn.query(`CREATE DATABASE ${this.database};`);
+    } catch(e) {
+      // postgres doesn't support IF NOT EXISTS for db creation so just ignore the error
```
dfellis, line 18: You can avoid this try-catch with a conditional here. I think the following would work:

```plpgsql
DO $$
DECLARE
  db string;
BEGIN
  SELECT datname INTO db
  FROM pg_database
  WHERE datname = ${this.database};
  IF db IS NULL THEN
    CREATE DATABASE ${this.database};
  END IF;
END $$;
```

test.env
```diff
@@ -0,0 +1,3 @@
+A0_ENABLED=true
+A0_DOMAIN=https://auth.iasql.com/
+A0_AUDIENCE=https://api.iasql.com
```
dfellis, line 3: Was this file committed by accident?

test/modules/aws-account-integration.ts
```diff
@@ -14,8 +14,8 @@ const query = runQuery.bind(null, dbAlias);
 const sync = runSync.bind(null, dbAlias);
 
 jest.setTimeout(360000);
-beforeAll(execComposeUp);
-afterAll(() => execComposeDown());
+beforeAll(async () => await execComposeUp());
+afterAll(async () => await execComposeDown());
```
dfellis, line 18: This is a useless change. `async` makes the function return a promise, which is what `execComposeUp` already returns. You can revert all of these and it'll still work perfectly fine, just like it did before.

src/router/db.ts
```diff
@@ -14,13 +15,18 @@ db.post('/new', async (req, res) => {
...
+      )
+    );
   } catch (e) {
     res.status(500).end(logger.error(e));
   }
 });
 
-db.post('/import', async (req, res) => {
+// TODO revive and test
```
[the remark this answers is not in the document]
depombo: Commenting the import route for now? There is no way to test if changes to it are correct

test/modules/aws-account-integration.ts
dfellis, line 18: In fact, since it doesn't use any of the input arguments either, wrapping them in anonymous functions is also not necessary. You can just write this and it should work fine:

```suggestion
beforeAll(execComposeUp);
afterAll(execComposeDown);
```

test/modules/aws-cloudwatch-integration.ts
```diff
@@ -12,16 +12,16 @@ const uninstall = runUninstall.bind(null, dbAlias);
 
 const modules = ['aws_cloudwatch@0.0.1'];
 jest.setTimeout(240000);
-beforeAll(execComposeUp);
-afterAll(() => execComposeDown(modules));
+beforeAll(async () => await execComposeUp());
+afterAll(async () => await execComposeDown(modules));
```
dfellis, line 16:
```suggestion
beforeAll(execComposeUp);
afterAll(execComposeDown.bind(null, modules));
```

test/modules/aws-account-integration.ts
depombo: Well, no. `afterAll` was makes Jest hang up with an error that not all async operations completed when you run it for a single test. Feel free to try it

test/modules/aws-ecr-integration.ts
```diff
@@ -15,16 +15,16 @@ const query = runQuery.bind(null, dbAlias);
 const modules = ['aws_ecr@0.0.1'];
 
 jest.setTimeout(240000);
-beforeAll(execComposeUp);
-afterAll(() => execComposeDown(modules));
+beforeAll(async () => await execComposeUp());
+afterAll(async () => await execComposeDown(modules));
```
dfellis, line 19:
```suggestion
beforeAll(execComposeUp);
afterAll(execComposeDown.bind(null, modules));
```

test/modules/aws-ecs-integration.ts
```diff
@@ -45,16 +45,16 @@ const publicRepositoryName = `${prefix}${dbAlias}publicrepository`;
 const containerNamePublicRepository = `${prefix}${dbAlias}containerpublicrepository`;
 
 jest.setTimeout(240000);
-beforeAll(execComposeUp);
-afterAll(() => execComposeDown(modules));
+beforeAll(async () => await execComposeUp());
+afterAll(async () => await execComposeDown(modules));
```
dfellis, line 49:
```suggestion
beforeAll(execComposeUp);
afterAll(execComposeDown.bind(null, modules));
```

test/modules/aws-elb-integration.ts
```diff
@@ -23,16 +23,16 @@ const lbType = LoadBalancerTypeEnum.APPLICATION;
 const lbIPAddressType = IpAddressType.IPV4;
 
 jest.setTimeout(360000);
-beforeAll(execComposeUp);
-afterAll(() => execComposeDown(modules));
+beforeAll(async () => await execComposeUp());
+afterAll(async () => await execComposeDown(modules));
```
dfellis, line 27:
```suggestion
beforeAll(execComposeUp);
afterAll(execComposeDown.bind(null, modules));
```

src/services/repositories/metadata.ts
depombo: nice

test/modules/aws-vpc-integration.ts
```diff
@@ -13,16 +13,16 @@ const availabilityZone = `${process.env.AWS_REGION ?? 'barf'}a`;
 const randIPBlock = Math.floor(Math.random() * 255);
 
 jest.setTimeout(240000);
-beforeAll(execComposeUp);
-afterAll(() => execComposeDown(modules));
+beforeAll(async () => await execComposeUp());
+afterAll(async () => await execComposeDown(modules));
```
dfellis, line 17:
```suggestion
beforeAll(execComposeUp);
afterAll(execComposeDown.bind(null, modules));
```

.github/workflows/ci.yml
```diff
@@ -114,6 +114,5 @@ jobs:
           MODULES: ${{ needs.setup-module-test.outputs['test-modules'] }}
           AWS_ACCESS_KEY_ID: ${{ secrets[env.MATRIX_AWS_ACCESS_KEY_ID] }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets[env.MATRIX_AWS_SECRET_ACCESS_KEY] }}
-          IRONPLANS_TOKEN: ${{ secrets.IRONPLANS_TOKEN }}
```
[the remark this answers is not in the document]
depombo: that was an old token for auth0 authentication which is only needed if authentication is on. this test has authentication off. there was actually a bug in `simple-integration` (`AO` instead of `A0`) which that made it so that it never tested the authentication path so they IP tokens was completely ignored / misused

test/modules/aws-account-integration.ts
dfellis: https://jestjs.io/docs/api#afterallfn-timeout

The docs do not mention an argument parameter, so it should work just fine. I'm guessing some other change happened at the same time and was misattributed to this change for fix?

depombo: https://github.com/iasql/iasql-engine/pull/542#discussion_r831152060

ormconfig-metadata.js
aguillenv: oh so I think this answers part of my question [here](https://github.com/iasql/iasql-engine/pull/542#pullrequestreview-915462650) but how do we handle persistence? where are we creating and keeping our own private db with this info? I think that part is what I do not follow

dfellis: > oh so I think this answers part of my question [here](https://github.com/iasql/iasql-engine/pull/542#pullrequestreview-915462650) but how do we handle persistence? where are we creating and keeping our own private db with this info? I think that part is what I do not follow

That's why I asked about the accidental usage of `__example__` as the database name in another of my comments. :)

Our RDS database will have one database only accessible to the root account that we store this metadata in. (Similarly for anyone running it locally, it will be that same database name in their local postgres docker container.)

Basically we will have one singleton database that is for the engine itself and not for the users of the engine.

test/modules/aws-account-integration.ts
depombo: this doesn't work

dfellis: Oh, that can explain that. They must be inspecting the type of the argument passed to it and doing different things for different types? :(

src/services/repositories/metadata.ts
depombo: actually I had tried a form of this and didn't because `QueryError: CREATE DATABASE cannot be executed from a function` going to leave a comment about that

dfellis (approved):

aguillenv (approved):

PR #545, opened 2022-03-21
Bootstrap scheduler and remove a MetadataRepo call from iasql ops

depombo:


src/router/db.ts
```diff
@@ -76,17 +78,29 @@ db.get('/remove/:dbAlias', async (req, res) => {
 
 db.post('/apply', async (req, res) => {
   const { dbAlias, dryRun } = req.body;
+  if (!dbAlias) return res.status(400).json(
+    `Required key(s) not provided: ${[
+      'dbAlias',
+    ].filter(k => !req.body.hasOwnProperty(k)).join(', ')}`
+  );
```
dfellis, line 85: If there's only one key, do we need the array and filter stuff?

src/router/db.ts
```diff
@@ -76,17 +78,29 @@ db.get('/remove/:dbAlias', async (req, res) => {
...
   }
 });
 
 db.post('/sync', async (req, res) => {
   const { dbAlias, dryRun } = req.body;
+  if (!dbAlias) return res.status(400).json(
+    `Required key(s) not provided: ${[
+      'dbAlias',
+    ].filter(k => !req.body.hasOwnProperty(k)).join(', ')}`
+  );
```
dfellis, line 100: Same here

dfellis (approved):
Only minor comments

src/router/db.ts
depombo: that is true

src/router/db.ts
dfellis:
```suggestion
  if (!dbAlias) return res.status(400).json('Required key not provided: dbAlias');
```

PR #546, opened 2022-03-21
Automate the population of the 'tables' and 'functions' fields in the 'provides' sub-object

dfellis:
This is the first half of #478

The second half is automating the `entities` field (will be in a separate PR

### DIFF
--- a/src/modules/interfaces.ts
+++ b/src/modules/interfaces.ts
@@ -349,6 +349,9 @@ export class Module {
           .replace(/\basync\b/g, '')
           .replace(/\bawait\b/g, '')
           .replace(/^/, 'function')
+          // The following are only for the test suite, but need to be included at all times
+          .replace(/\/* istanbul ignore next *\//g, '')
+          .replace(/cov_.*/g, '')
       )();
       const tables: string[] = [];
       const functions: string[] = [];
--- a/src/modules/interfaces.ts
+++ b/src/modules/interfaces.ts
@@ -304,8 +304,8 @@ export class Module {
   dependencies: string[];
   provides: {
     entities: { [key: string]: any, };
-    tables?: string[];
-    functions?: string[];
+    tables: string[];
+    functions: string[];
     context?: Context;
   };
   utils: { [key: string]: any, };
@@ -319,7 +319,12 @@ export class Module {
     this.name = def.name;
     this.version = def.version;
     this.dependencies = def.dependencies;
-    this.provides = def.provides;
+    this.provides = {
+      entities: def.provides.entities,
+      tables: def.provides.tables ?? [], // These will be populated automatically below
+      functions: def.provides.functions ?? [],
+    }
+    if (def.provides.context) this.provides.context = def.provides.context;
     this.utils = def?.utils ?? {};
     this.mappers = def.mappers;
     if (!def.migrations) {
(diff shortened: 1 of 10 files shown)


dfellis:
https://github.com/iasql/iasql-engine/runs/5633883988?check_suite_focus=true#step:5:15

Crap, the Jest coverage source rewriting is interfering with my own source rewriting... :(

dfellis:
I tested this manually and it works, but ti breaks the entire test suite.

Let me think some more about this.

depombo:
> https://github.com/iasql/iasql-engine/runs/5633883988?check_suite_focus=true#step:5:15
> 
> Crap, the Jest coverage source rewriting is interfering with my own source rewriting... :(

what source re-writing are you referring to?

dfellis:
> > https://github.com/iasql/iasql-engine/runs/5633883988?check_suite_focus=true#step:5:15
> > Crap, the Jest coverage source rewriting is interfering with my own source rewriting... :(
> 
> what source re-writing are you referring to?

This: https://github.com/iasql/iasql-engine/pull/546/files#diff-5dabb7c84cdc5f16ab1995093ebdd0e4b73fb64524e656f30a4cc86a6e1148c3R345-R355

I think I have fixed it, but we'll see in a few minutes.

src/modules/interfaces.ts
```diff
@@ -337,6 +342,32 @@ export class Module {
...
+        // TODO: Proper parsing (maybe LP?)
+        if (/^create table/i.test(text)) {
+          tables.push((text.match(/^[^"]*"([^"]*)"/) ?? [])[1]);
+        } else if (/^create or replace procedure/i.test(text)) {
+          functions.push((text.match(/^create or replace procedure ([^(]*)/i) ?? [])[0]);
+        } else if (/^create or replace function/i.test(text)) {
+          functions.push((text.match(/^create or replace function ([^(]*)/i) ?? [])[0]);
+        }
+        // Don't do anything for queries that don't match
+      }, });
```
depombo, line 368: this makes me think that we might be able to do pull off https://github.com/iasql/docs/issues/14. right now I have to massage a bit. one of the things is remove stored procedures which are not supported by dbml/dbdocs

depombo (approved):

src/modules/interfaces.ts
dfellis: This is relying on the assumption that all modules use TypeORM and they use auto-generated migration files, but that is the case right now and if we enforce the latter we could theoretically "late bind" the migrations for the modules and further reduce their size.

depombo: oh in the dbdocs case we would actually be doing the regex on the output of `pg_dump` and it would be to remove them from the sql file

depombo: not exactly the same, just related

src/modules/interfaces.ts
```diff
@@ -337,6 +342,32 @@ export class Module {
...
+          .replace(/\basync\b/g, '')
+          .replace(/\bawait\b/g, '')
+          .replace(/^/, 'function')
+          // The following are only for the test suite, but need to be included at all times
+          .replace(/\/* istanbul ignore next *\//g, '')
+          .replace(/cov_.*/g, '')
+      )();
+      const tables: string[] = [];
+      const functions: string[] = [];
+      syncified({ query: (text: string) => {
```
aguillenv, line 358: This looks good but I think my question is why is this inside the condition `if (!def.migrations) {`? shouldn't be always there?

dfellis: So nothing went through the other branch, but also if anything did, it would imply that the migrations don't exist as a separate file and this trick would not work.

In any case in the next PR we decided to just not allow "unconventional" migration definitions so the other branch no longer exists.

PR #553, opened 2022-03-22
create a user even if they have not created a db

depombo:


src/services/repositories/metadata.ts
```diff
@@ -66,9 +66,17 @@ class MetadataRepo {
...
+  async getDbs(a0Id: string, email: string): Promise<IasqlDatabase[]> {
     const user = await this.userRepo.findOne(a0Id);
-    if (!user) return [];
+    if (!user) {
+      // create the new user
+      const user = new IasqlUser();
+      user.id = a0Id;
+      user.email = email;
+      user.iasqlDatabases = [];
+      await this.userRepo.save(user);
```
dfellis, line 77: It doesn't like the shadowing, so this should work:

```suggestion
      const newUser = new IasqlUser();
      newUser.id = a0Id;
      newUser.email = email;
      newUser.iasqlDatabases = [];
      await this.userRepo.save(newUser);
```

dfellis (approved):

PR #554, opened 2022-03-22
replace procedures with functions using dblink

depombo:


src/migration/1647526647810-init.ts
```diff
@@ -15,26 +15,30 @@ export class init1647526647810 implements MigrationInterface {
...
+                -- raise exception '%', _dblink_sql;
+                PERFORM dblink_exec('iasqlopconn', _dblink_sql);
+                PERFORM dblink_disconnect('iasqlopconn');
+                raise notice '% scheduled', _opid;
                 -- times out after 45 minutes = 60 * 45 = 2700 seconds
                 -- currently the longest is RDS where the unit test has a timeout of 16m
-                while _counter < 2700 loop
-                    if (select end_date from iasql_operation where opid = _opid) is not null then
+                while _counter < 25 loop
+                    if (select end_date from iasql_operation where opid = _opid for update) is not null then
```
depombo, line 41: this shouldn't be here

src/migration/1647526647810-init.ts
```diff
@@ -15,26 +15,30 @@ export class init1647526647810 implements MigrationInterface {
...
+                _dblink_sql := format('perform graphile_worker.add_job(%L, json_build_object(%L, %L, %L, %L));', 'operation', 'opid', _opid, 'optype', _optype);
+                -- raise exception '%', _dblink_sql;
+                PERFORM dblink_exec('iasqlopconn', _dblink_sql);
+                PERFORM dblink_disconnect('iasqlopconn');
+                raise notice '% scheduled', _opid;
                 -- times out after 45 minutes = 60 * 45 = 2700 seconds
                 -- currently the longest is RDS where the unit test has a timeout of 16m
-                while _counter < 2700 loop
-                    if (select end_date from iasql_operation where opid = _opid) is not null then
+                while _counter < 25 loop
```
depombo, line 40: this is to easily test without killing the process when it doesn't poll correctly

src/migration/1647526647810-init.ts
```diff
@@ -27,17 +27,18 @@ export class init1647526647810 implements MigrationInterface {
...
-                raise notice '% started', _opid;
                 PERFORM dblink_connect('iasqlopconn', 'loopback_dblink');
-                -- _dblink_sql := format('perform graphile_worker.add_job(%L);', 'operation');
-                _dblink_sql := format('perform graphile_worker.add_job(%L, json_build_object(%L, %L, %L, %L));', 'operation', 'opid', _opid, 'optype', _optype);
+                _dblink_sql := format('insert into iasql_operation (opid, optype, params) values (%L, %L, array[''%s'']::text[]);', _opid, _optype, array_to_string(_params, ''','''));
                 -- raise exception '%', _dblink_sql;
                 PERFORM dblink_exec('iasqlopconn', _dblink_sql);
+                _dblink_sql := format('select graphile_worker.add_job(%L, json_build_object(%L, %L, %L, %L, %L, array[''%s'']::text[]));', 'operation', 'opid', _opid, 'optype', _optype, 'params', array_to_string(_params, ''','''));
+                -- raise exception '%', _dblink_sql;
+                -- allow statement that returns results in dblink https://stackoverflow.com/a/28299993
```
dfellis, line 36: Nice...

src/services/scheduler.ts
```diff
@@ -57,23 +57,27 @@ export async function start(dbId: string, dbUser:string) {
...
           let output = await promise;
-          console.log('job done, start query')
+          // once the operation completes updating the `end_date`
+          // will complete the polling
+          const query = `
+            update iasql_operation
+            set end_date = now(), output = '${output}'
+            where opid = uuid('${opid}');
+          `;
+          console.log(query);
```
dfellis, line 69: Nit: Delete this?

```suggestion
```

src/services/db-manager.ts
```diff
@@ -57,6 +57,21 @@ export const baseConnConfig: PostgresConnectionOptions = {
...
+  const conn = await createConnection({
+    ...baseConnConfig,
+    name: `${dbId}-${Math.floor(Math.random()*10000)}-dblink`,
+    database: dbId,
+  });
+  console.log(dbUser, config.dbUser);
+  // create a dblink server to reuse the connections
+  // https://aws.amazon.com/blogs/database/migrating-oracle-autonomous-transactions-to-postgresql/
+  await conn.query(`CREATE EXTENSION IF NOT EXISTS dblink;`);
+  await conn.query(`CREATE SERVER IF NOT EXISTS loopback_dblink FOREIGN DATA WRAPPER dblink_fdw OPTIONS (host '${config.dbHost}', dbname '${dbId}', port '${config.dbPort}');`);
```
dfellis, line 70: What if you did this instead?

```suggestion
  await conn.query(`CREATE SERVER IF NOT EXISTS loopback_dblink FOREIGN DATA WRAPPER dblink_fdw OPTIONS (host 'localhost', dbname '${dbId}', port '${config.dbPort}');`);
```

It is the database itself talking to itself, so why go through the network card? There may also be special behavior granted to this form?

depombo: tried it, but during testing it is already set to `localhost` I believe

src/services/scheduler.ts
depombo: so this was super useful debugging when there was an error in the query itself. I would prefer to leave it, but can remove. we definitely need to add a proper logger

src/services/scheduler.ts
```diff
@@ -16,6 +17,11 @@ const workerShutdownEmitter = new EventEmitter();
 export async function start(dbId: string, dbUser:string) {
   // use the same connection for the scheduler and its operations
   const conn = await TypeormWrapper.createConn(dbId, { name: `${dbId}-${Math.floor(Math.random()*10000)}-scheduler`, });
+  // create a dblink server for SP. used to reuse connections
+  // https://aws.amazon.com/blogs/database/migrating-oracle-autonomous-transactions-to-postgresql/
+  await conn.query(`CREATE EXTENSION IF NOT EXISTS dblink;`);
+  await conn.query(`CREATE SERVER IF NOT EXISTS loopback_dblink FOREIGN DATA WRAPPER dblink_fdw OPTIONS (host '${config.dbHost}', dbname '${dbId}', port '${config.dbPort}');`);
```
dfellis, line 23: Quick question: Is this "server" entity scoped per database? If not we'll have issues with multitenancy and maybe we should include the `${dbId}` in the name of the link?

dfellis (approved):

src/services/scheduler.ts
depombo: oh hmmm I guess the tests don't stress that either because they all run independently and create a single db. let me create 2 dbs in one test

depombo: good question

depombo: yep the test failed, fixing it

dfellis (approved):
Nice new tests, too :)

PR #555, opened 2022-03-23
Fix the iasql_ stored procedure permissions

dfellis:
graphile's security is like Fort Knox.

Proof that it works is in the contributions channel in Discord ;)

### DIFF
--- a/src/services/db-manager.ts
+++ b/src/services/db-manager.ts
@@ -63,17 +63,22 @@ export const baseConnConfig: PostgresConnectionOptions = {
 // But we'll want to decide (before public launch?) one of them and replace this
 // TODO: #2, also try to roll back the `GRANT CREATE` to something a bit narrower in the future
 export function newPostgresRoleQuery(user: string, pass: string, dbId: string) {
-  // Only do this if a0 is enabled. Otherwise, you cannot grant these privileges to the default postgres user, the query gets stuck
-  return config.a0Enabled ? `
+  return `
     CREATE ROLE ${user} LOGIN PASSWORD '${pass}';
     GRANT CONNECT ON DATABASE ${dbId} TO ${user};
     GRANT CREATE ON SCHEMA public TO ${user};
-  ` : '';
+    GRANT USAGE ON SCHEMA graphile_worker TO ${user};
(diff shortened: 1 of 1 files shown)


src/services/db-manager.ts
```diff
@@ -63,38 +63,41 @@ export const baseConnConfig: PostgresConnectionOptions = {
...
+  return `
     DROP ROLE IF EXISTS ${user};
-  ` : '';
+  `;
 }
 
 // Create a randomly generated username and password, an 8 char username [a-z][a-z0-9]{7} and a
 // 16 char password [a-zA-Z0-9!@#$%^*]{16}
 export function genUserAndPass(): [string, string] {
-    if (!config.a0Enabled) return [config.dbUser, config.dbPassword];
```
depombo, line 97: if we are doing this we need to remove the possibility of disabling auth via environment variables? otherwise I think with this code change we would generate a new user every time every time we create a db?

depombo: however, if we do that running the server locally might become harder since they would have to go through our auth

dfellis: This confused me at first, too, but `genUserAndPass` is *not* dealing with the Auth0 user and pass, it's the Postgres user and pass. This change simply makes local dev more like production by having the generated URL when connecting a new account give you a new account per database -- but it doesn't stop you from using the `postgres:test` pair if you still want to?

depombo: oh ok. well every db would get a new "user" in the metadata repository which would be harder to track? `list`-ing dbs will just give you one db? right now it groups them all under an `iasql` user. as long as we are okay with that?

dfellis: > oh ok. well every db would get a new "user" in the metadata repository which would be harder to track? `list`-ing dbs will just give you one db? right now it groups them all under an `iasql` user. as long as we are okay with that?

So you'll see that the check involving the actual auth0 user still switches to the `iasql` user in that case, so you can/will see all databases, not just one. :) (I wouldn't have been able to test locally at all if the db list constantly changed out from under me.)

depombo: oh got it!

depombo (approved):

aguillenv (approved):

aguillenv (approved):

PR #560, opened 2022-03-24
Fix `iasql_` functions for normal users and improve UX of `iasql_uninstall`

dfellis:


src/migration/1647526647810-init.ts
```diff
@@ -119,43 +119,128 @@ export class init1647526647810 implements MigrationInterface {
...
-        await queryRunner.query(`DROP FUNCTION "iasql_install(text[])"`);
-        await queryRunner.query(`DROP FUNCTION "iasql_uninstall(text)"`);
         await queryRunner.query(`DROP FUNCTION "iasql_uninstall(text[])"`);
+        await queryRunner.query(`DROP FUNCTION "iasql_uninstall(text)"`);
+        await queryRunner.query(`DROP FUNCTION "iasql_install(text[])"`);
+        await queryRunner.query(`DROP FUNCTION "iasql_install(text)"`);
+        await queryRunner.query(`DROP FUNCTION "iasql_sync"`);
+        await queryRunner.query(`DROP FUNCTION "iasql_plan"`);
+        await queryRunner.query(`DROP FUNCTION "iasql_apply"`);
+        await queryRunner.query(`DROP FUNCTION "until_iasql_operation"`);
```
dfellis, line 243: While we will *never* actually run this downgrade migration, the prior order would have crashed because almost all of the functions depend on `until_iasql_operation` and we were trying to remove it first.

src/services/db-manager.ts
```diff
@@ -67,15 +67,6 @@ export function newPostgresRoleQuery(user: string, pass: string, dbId: string) {
...
     GRANT CREATE ON SCHEMA public TO ${user};
-    GRANT USAGE ON SCHEMA graphile_worker TO ${user};
-    GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA graphile_worker TO ${user};
-    GRANT EXECUTE ON ALL PROCEDURES IN SCHEMA graphile_worker TO ${user};
-    GRANT INSERT, SELECT ON ALL TABLES IN SCHEMA graphile_worker TO ${user};
-    GRANT USAGE, SELECT ON ALL SEQUENCES IN SCHEMA graphile_worker TO ${user};
-    CREATE POLICY ${user}_sel_job ON graphile_worker.jobs FOR SELECT TO ${user} USING (true);
-    CREATE POLICY ${user}_ins_job ON graphile_worker.jobs FOR INSERT TO ${user} WITH CHECK (true);
-    CREATE USER MAPPING IF NOT EXISTS FOR ${user} SERVER loopback_dblink_${dbId} OPTIONS (user '${config.dbUser}', password '${config.dbPassword}');
-    GRANT USAGE ON FOREIGN SERVER loopback_dblink_${dbId} TO ${user};
```
dfellis, line 78: I'm so happy!

depombo: same 🙌

src/migration/1647526647810-init.ts
depombo: should we just keep this `down` migration function empty?

src/migration/1647526647810-init.ts
```diff
@@ -119,43 +119,128 @@ export class init1647526647810 implements MigrationInterface {
...
+              select current_database() into _db_id;
+              -- reuse the 'iasqlopconn' db dblink connection if one exists for the session
+              -- dblink connection closes automatically at the end of a session
+              SELECT count(1) INTO _dblink_conn_count FROM dblink_get_connections()
+                  WHERE dblink_get_connections@>'{iasqlopconn}';
+              IF _dblink_conn_count = 0 THEN
+                  PERFORM dblink_connect('iasqlopconn', 'loopback_dblink_' || _db_id);
+              END IF;
+              -- define the query to get the current tables and record counts for the modules to be removed
+              -- TODO: Are these hoops to encode into JSON and then decode back out necessary now that
```
depombo, line 164: hmm good question

depombo (approved):

src/migration/1647526647810-init.ts
dfellis: I did emphasize the "never", but y'know, "never say never" ;) It's entirely possible that we want to be able to revert for *some* reason in the future, so I think keeping it around is fine (plus the migration generation automatically does that anyway).

PR #562, opened 2022-03-24
Automate db creation using CloudFormation template

aguillenv:


src/modules/aws_account@0.0.1/index.ts
```diff
@@ -31,6 +31,12 @@ export const AwsAccount: Module = new Module({
       awsClient: null, // Just reserving this name to guard against collisions between modules.
     },
   },
+  utils: {
```
dfellis, line 34: Why is this piece here? It's not used by the module in any way...

src/router/db.ts
```diff
@@ -95,3 +95,26 @@ db.post('/sync', async (req, res) => {
     res.status(500).end(logger.error(e));
   }
 });
+
+db.get('/get/:dbAlias', async (req, res) => {
+  const { dbAlias, } = req.params;
+  console.log('dbalias', dbAlias)
+  if (!dbAlias) return res.status(400).json("Required param 'dbAlias' not provided");
+  try {
+    const dbs = await iasql.list(dbMan.getUid(req.user), dbMan.getEmail(req.user), false);
```
depombo, line 104: you can just call the `getDb` method from the respository directory since here you are fetching all of the rows?

src/services/iasql.ts
```diff
@@ -908,3 +908,25 @@ export async function uninstall(moduleList: string[], dbId: string, orm?: Typeor
   }
   return "Done!";
 }
+
+export async function getStackInfo(dbId: string, stackName: string, ormOpt?: TypeormWrapper) {
+  let orm: TypeormWrapper | null = null;
+  try {
+    orm = !ormOpt ? await TypeormWrapper.createConn(dbId) : ormOpt;
+    // Find all of the installed modules, and create the context object only for these
+    const moduleNames = (await orm.find(IasqlModule)).map((m: IasqlModule) => m.name);
```
depombo, line 917: why does the CF stack need to know what are the installed modules? to generate the permissions/policy? if so, installing new modules could get troublesome?

src/modules/aws_account@0.0.1/index.ts
aguillenv: Hmm yes, I can create this function outside the module, here does not makes sense. It was here because the `aws_account` module is the only way to have access to the newly user's credentials to be able to instantiate the `AwsClient`, then `describe` the CloudFormation stack and show the output for the user in the UI. But I only need the context of the module that I can get it outside.

src/services/iasql.ts
aguillenv: The comments are not updated, just copy&paste from `apply`, I think. Just copying the pattern to get the `aws_account` module, get its context and being able to call the AWS API  with the new credentials. This is only to call CloudFormation to get the newly created stack and show the details to the users.

src/cloudformation/newUserAndDbTemplate.yaml
```diff
@@ -0,0 +1,217 @@
...
+      Code:
+        S3Bucket: "slurp-cred" # TODO: make it an input?
+        S3Key: "newUserAndDb.zip" # TODO: to be updated
+      Tags:
+        -
+          Key: "Lambda Function Name"
+          Value: !Sub '${AWS::StackName}-LambdaFunction'
+
+  rLambdaFunctionCaller:
+    Type: AWS::Lambda::Function
```
depombo, line 148: do we spin down the lambda fn after we are done?

src/cloudformation/newUserAndDbTemplate.yaml
```diff
@@ -0,0 +1,217 @@
...
+          secret_access_key: !GetAtt 'rIamUserAccessKey.SecretAccessKey'
+          aws_region: !Ref 'pAwsRegion'
+          db_alias: !Ref 'pIasqlDbName'
+          a0_token: !Ref 'pA0Token'
+          api_url: !Ref 'pIasqlUrl'
+      Handler: index.handler
+      Timeout: 60
+      Runtime: nodejs14.x
+      Code:
+        S3Bucket: "slurp-cred" # TODO: make it an input?
```
depombo, line 140: probably rename to `iasql-cred` and maybe delete after? in which case making the name configurable has less significance?

src/cloudformation/newUserAndDbTemplate.yaml
```diff
@@ -0,0 +1,217 @@
...
+        default: IaSQL Bearer Token
+
+###
+#### Template input parameters
+####
+
+Parameters:
+  pIasqlDbName:
+    Type: String
+    Description: Do not update
```
depombo, line 35: this means it can't be updated after it is created? Maybe "immutable after creation or something"

depombo: oh this is the params we are passing from the ui. got it

src/services/gateways/aws.ts
```diff
@@ -170,6 +172,8 @@ export class AWS {
     this.ecsClient = new ECSClient(config);
     this.rdsClient = new RDSClient(config);
     this.cwClient = new CloudWatchLogsClient(config);
+    // Service endpoint need to be in the same region where we have the s3 bucket with the lambda function and where the user runs the cloudformation template
+    this.cfnClient = new CloudFormationClient({credentials: config.credentials, region: 'us-east-2'});
```
depombo, line 176: do we need to set the region? I input a region into the ui, the stack has `us-east-2` hardcoded and finally I can toggle across regions in the AWS ui to create the stack in some other region? can we just let the user pick the `region` they want to manage based on where the stack gets created?

depombo:
Screen Shot 2022-03-24 at 5 18 30 PM

aguillenv:
> I can't test the full flow because I get this Screen Shot 2022-03-24 at 5 18 30 PM

oh I'm sorry you could not test it, it is because right now the `dbAlias` is required for that flow. I'll try to think a way to keep it optional but I'm not sure it would be possible

src/cloudformation/newUserAndDbTemplate.yaml
aguillenv: the lambda function runs and returns but it will stay listed as long as the stack is not deleted. Resources will only be deleted once the user deletes the CloudFormation stack.

src/cloudformation/newUserAndDbTemplate.yaml
aguillenv: definitely, I still need to pass thezio file and the template to our s3 bucket, this is still the testing one in my account.

src/cloudformation/newUserAndDbTemplate.yaml
aguillenv: what I want to express is that users must not update the values creating the cloudformation template because those are the values from our UI and If they changed them here we loose track of them.

src/services/gateways/aws.ts
aguillenv: I can double check but the issue here is that the stack need to be created in the same region where **our** s3 bucket with the lambda function will live, otherwise the CloudFormation stack will fail on creation. That's why I've created this special client with a fixed region (the region where the s3 bucket live right now)

dfellis:
> > I can't test the full flow because I get this Screen Shot 2022-03-24 at 5 18 30 PM
> 
> oh I'm sorry you could not test it, it is because right now the `dbAlias` is required for that flow. I'll try to think a way to keep it optional but I'm not sure it would be possible

Simple solution: we make the alias required in the engine, and move the automatic alias generation when not entered by the user into the dashboard. Then an empty alias is transformed by the dashboard to a generated one and that one is saved by the engine and the cloudformation stack.

src/services/gateways/aws.ts
depombo: I see, we would have to create this dynamically in the ux flow I am describing

dfellis (approved):

PR #563, opened 2022-03-24
Move away from env files to a single env set at build time

depombo:
Also sets the sentry environment


src/config/production.ts
```diff
@@ -2,16 +2,18 @@ import { ConfigInterface, throwError, } from './config';
 
 const config: ConfigInterface = {
   port: 8088,
-  dbHost: process.env.DB_HOST ?? throwError('No DB Host defined'), // TODO: Move away from env var
+  dbHost: 'https://db.iasql.com',
```
dfellis, line 5: Drop the http part

dfellis:
```suggestion
  dbHost: 'db.iasql.com',
```

src/config/staging.ts
```diff
@@ -2,16 +2,18 @@ import { ConfigInterface, throwError, } from './config';
 
 const config: ConfigInterface = {
   port: 8088,
-  dbHost: process.env.DB_HOST ?? throwError('No DB Host defined'), // TODO: Move away from env var
+  dbHost: 'https://db-staging.iasql.com',
```
dfellis, line 5:
```suggestion
  dbHost: 'db-staging.iasql.com',
```

dfellis:
To make sure that we don't forget. I think the `IASQL_ENV` var needs to be declared in the Dockerfile, too.

dfellis (approved):

package.json
```diff
@@ -10,8 +10,8 @@
     "test": "IASQL_ENV=test jest -i",
     "coverage": "IASQL_ENV=test jest -i --coverage",
     "coverage:local": "IASQL_ENV=test export $(cat .env | xargs) && yarn coverage",
-    "start": "if [ \"$IASQL_ENV\" = \"\" ]; then IASQL_ENV=production yarn serve; elif [ \"$IASQL_ENV\" = \"local\" ]; then yarn start:dev; else yarn serve; fi",
```
aguillenv, line 13: removing this breaks the `docker-compose:dev` script to start locally with hot reload. I'll see how can we could maintain it

PR #564, opened 2022-03-25
Variadic install and uninstall

dfellis:


test/helpers.ts
```diff
@@ -38,11 +38,11 @@ export function runSync(dbAlias: string) {
 }
 
 export function runInstall(dbAlias: string, mods: string[]) {
-  return runQuery(dbAlias, `select iasql_install(array[${mods.map(m => `'${m}'`)}]);`);
+  return runQuery(dbAlias, `select iasql_install(${mods.map(m => `'${m}'`)});`);
```
aguillenv, line 41: oh really? don't you still need to pass the `array[]` since the type is `text[]`?

dfellis: Nope, I marked it `variadic` which is similar to `...rest` in Javascript.

Has to be the last parameter and will become an array of values. The difference is that here they must all be the same type.

aguillenv (approved):

depombo (approved):

depombo:
Very cool!

PR #568, opened 2022-03-26
Improve the UX of the remaining iasql_ functions

### LINKED ISSUE #538: Allow selecting or print better (error) messages on iasql ops
Screen Shot 2022-03-17 at 7 28 57 PM

Most of what PG displays by default is noisy because we are just doing an rpc call under the hood. 

Some ideas:
- [ ] For `plan` and `apply` create an ascii table like the one from the CLI with the `output`
https://www.npmjs.com/package/ascii-table OR use a virtual table/view somehow that could be generated from csv 
- [ ] just use `err.message` on `err` and create a `jsonb` `err_stack` column in `iasql_operation` that is printed with `jsonb_pretty`
- [ ] Use colored text
https://stackoverflow.com/questions/33867304/psql-displaying-ansi-colored-text
- [ ] Hide the sql stack trace or `CONTEXT`. Should be possible by setting the `log_error_verbosity` to `TERSE` 
 https://www.postgresql.org/docs/13/runtime-config-logging.html
 Might also be possible to workaround it doing something like
https://dba.stackexchange.com/questions/7214/generate-an-exception-with-a-context
- [ ] Modify `log_line_prefix` to hide the `LOCATION` and error code, e.g. `P0001`

dfellis:
Resolves #538

### DIFF
--- a/test/modules/aws-account-integration.ts
+++ b/test/modules/aws-account-integration.ts
@@ -56,7 +56,7 @@ describe('AwsAccount Integration Testing', () => {
 
   it('does absolutely nothing when you plan this', query(`
     select iasql_plan_apply();
-  `));
+  `, (res: any[]) => expect(res.length).toBe(0)));
 
   it('removes the useless row', query(`
     DELETE FROM aws_account WHERE access_key_id = 'fake'
--- a/test/modules/aws-account-integration.ts
+++ b/test/modules/aws-account-integration.ts
@@ -55,7 +55,7 @@ describe('AwsAccount Integration Testing', () => {
   it('does absolutely nothing when you sync this', sync());
 
   it('does absolutely nothing when you plan this', query(`
-    select iasql_plan();
+    select iasql_plan_apply();
   `));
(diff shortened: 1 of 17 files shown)


test/modules/aws-account-integration.ts
```diff
@@ -55,7 +55,7 @@ describe('AwsAccount Integration Testing', () => {
   it('does absolutely nothing when you sync this', sync());
 
   it('does absolutely nothing when you plan this', query(`
-    select iasql_plan();
+    select iasql_plan_apply();
```
depombo, line 58: would be nice to do a quick check and select that no records are return. not blocking though

depombo (approved):

test/modules/aws-account-integration.ts
depombo: 🙌

depombo (approved):

aguillenv:
🚀🚀🚀 awesome!

PR #581, opened 2022-03-28
add logger and logUserError for current logger.error

depombo:


src/services/logger.ts
```diff
@@ -1,21 +1,41 @@
 import * as sentry from '@sentry/node'
+import { Logger, LogFunctionFactory } from 'graphile-worker';
 
 import config from '../config';
 
-// TODO implement the rest
+const logFactory: LogFunctionFactory = (scope) => {
```
dfellis, line 6: I want to make a custom logger just for the test environments, but that can wait for another PR.

dfellis (approved):

PR #582, opened 2022-03-28
error has to be valid json

depombo:


src/services/scheduler.ts
```diff
@@ -10,6 +10,8 @@ import { IasqlDatabase } from '../metadata/entity';
 import config from '../config';
 
 const workerShutdownEmitter = new EventEmitter();
+// no max number of listeners warning
+workerShutdownEmitter.setMaxListeners(0);
```
depombo, line 14: I was to lazy to disconnect old dbs locally and I actually had more than 10 which is the default

dfellis: oof. So... I really want a stats dashboard with this reveal because I'm worried what's gonna happen to our production system if we explode the number of workers there...

dfellis: Not blocking, but we need a task for that, I think. But I wonder if Postgres can report # of connections to us on its own?

src/services/scheduler.ts
```diff
@@ -75,7 +77,9 @@ export async function start(dbId: string, dbUser:string) {
           output = typeof output === 'string' ? output : JSON.stringify(output);
           await conn.query(query);
         } catch (e) {
-          const error = logUserErr(e);
+          logUserErr(e);
+          // error must be valid JSON as a string
+          const error = JSON.stringify(e, Object.getOwnPropertyNames(e));
```
dfellis, line 82: Does this keep the `message` property?

depombo: this is what was there before my last PR. so yes

dfellis (approved):

src/services/scheduler.ts
depombo: I'm assuming the number of connections to postgres would be a much bigger problem. we at least know exactly how many dbs we have from the metadata repo. https://github.com/iasql/iasql-engine/issues/583

PR #584, opened 2022-03-28
rename endpoints and add documentation for running locally

depombo:


CONTRIBUTING.md
```diff
@@ -12,12 +12,41 @@ IaSQL is an eventually-consistent design where the SQL tables are the source of
...
+
+```bash
+curl \
+  --request POST \
+  --url 'http://localhost:8088/v1/db/connect/' \
+  --show-error --silent --fail \
+  --header 'content-type: application/json' \
+  --data '{
+    "dbAlias": "<db_name>",
+    "awsRegion": "us-east-1",
```
dfellis, line 34: May want to make this clear it doesn't always have to be `us-east-1`

CONTRIBUTING.md
```diff
@@ -12,12 +12,41 @@ IaSQL is an eventually-consistent design where the SQL tables are the source of
...
+    "awsAccessKeyId": "${process.env.AWS_ACCESS_KEY_ID}",
+    "awsSecretAccessKey": "${process.env.AWS_SECRET_ACCESS_KEY}"
+  }'
+```
+
+Now connect to the database is a simple as:
```
dfellis, line 40:
```suggestion
Now connecting to the database is a simple as:
```

dfellis (approved):

CONTRIBUTING.md
```diff
@@ -12,12 +12,41 @@ IaSQL is an eventually-consistent design where the SQL tables are the source of
...
+```
+
+`IASQL_ENV=local` configures the engine to use the values from `src/config/local.ts`. The Postgres superadmin user will be `postgres` and its password `test`. The Node.js server will start on port 8088. To create a new database in your local Postgres engine and connect it to an AWS account (and whatever region you prefer) send the following HTTP request to the local engine:
+
+
+```bash
+curl \
+  --request POST \
+  --url 'http://localhost:8088/v1/db/connect/' \
+  --show-error --silent --fail \
```
dfellis, line 30: Minor, but this line of flags is technically not necessary. It will just look a bit uglier without it?

```suggestion
```

CONTRIBUTING.md
```diff
@@ -12,12 +12,39 @@ IaSQL is an eventually-consistent design where the SQL tables are the source of
...
+```bash
+curl \
+  --request POST \
+  --url 'http://localhost:8088/v1/db/connect/' \
+  --header 'content-type: application/json' \
+  --data '{
+    "dbAlias": "<db_name>",
+    "awsRegion": "${process.env.AWS_REGION}",
+    "awsAccessKeyId": "${process.env.AWS_ACCESS_KEY_ID}",
+    "awsSecretAccessKey": "${process.env.AWS_SECRET_ACCESS_KEY}"
```
dfellis, line 34: I am clearly fuzzy-headed from the debugging I've been doing this afternoon. This won't work for like three reasons:

1. This is a bash script, not a Node.js script, so `process.env.*` doesn't exist. It should just be `$AWS_REGION` and the like instead.
2. Even if you fixed that, bash doesn't allow variable substitution within single-quoted strings, so the `$AWS_REGION` also wouldn't work.
3. You probably need the `\` newline escape even within the string context in bash, though I'm not 100% sure on that.

CONTRIBUTING.md
```diff
@@ -12,12 +12,39 @@ IaSQL is an eventually-consistent design where the SQL tables are the source of
...
+  }'
```
dfellis, line 35: Fix option 1:

```suggestion
  --data "{
    \"dbAlias\": \"<db_name>\",
    \"awsRegion\": \"$AWS_REGION\",
    \"awsAccessKeyId\": \"$AWS_ACCESS_KEY_ID\",
    \"awsSecretAccessKey\": \"$AWS_SECRET_ACCESS_KEY\"
  }"
```

dfellis: Fix option 2:

```suggestion
  --data <<EOF
{
  "dbAlias": "<db_name>",
  "awsRegion": "$AWS_REGION",
  "awsAccessKeyId": "$AWS_ACCESS_KEY_ID",
  "awsSecretAccessKey": "$AWS_SECRET_ACCESS_KEY"
}
EOF
```

depombo: this works as-is, actually

CONTRIBUTING.md
depombo: Fixed 1 and 2

dfellis (approved):

src/router/db.ts
```diff
@@ -8,8 +8,27 @@ import { logUserErr } from '../services/logger'
 
 export const db = express.Router();
 
-db.post('/new', async (req, res) => {
-  console.log('Calling /new');
+db.get('/connect/:dbAlias/:awsRegion/:awsAccessKeyId/:awsSecretAccessKey', async (req, res) => {
```
aguillenv, line 11: could you remind me why we need this GET connect?

PR #586, opened 2022-03-29
`aws_security_group` flakiness fix

### LINKED ISSUE #566: aws_security_group flakiness
The security group module is showing signs of flakiness. Misidentifying records from the cloud as being different from a record in the database and then failing due to uniqueness constraints when they are actually "the same" from the database's perspective.

Figuring out why this is happening and turning the `db.create` into a `db.update` call when this happens should fix it.

dfellis:
Resolves #566

### DIFF
--- a/src/services/iasql.ts
+++ b/src/services/iasql.ts
@@ -365,6 +365,7 @@ export async function apply(dbId: string, dryRun: boolean, ormOpt?: TypeormWrapp
         }
         records.forEach(r => {
           r.diff = findDiff(r.dbEntity, r.cloudEntity, r.idGen, r.comparator);
+          if (['SecurityGroup', 'SecurityGroupRule'].includes(r.table)) console.log(r.diff);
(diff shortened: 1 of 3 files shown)


src/services/iasql.ts
```diff
@@ -365,6 +365,7 @@ export async function apply(dbId: string, dryRun: boolean, ormOpt?: TypeormWrapp
         }
         records.forEach(r => {
           r.diff = findDiff(r.dbEntity, r.cloudEntity, r.idGen, r.comparator);
+          if (['SecurityGroup', 'SecurityGroupRule'].includes(r.table)) console.log(r.diff);
```
dfellis, line 368:
```suggestion
```

dfellis:
Pretty sure that test failure is due to contention issues (we only made it less likely to collide, not impossible)

aguillenv (approved):

depombo (approved):

PR #587, opened 2022-03-29
Attach credentials if not set on connect

aguillenv:


src/router/db.ts
```diff
@@ -46,6 +46,27 @@ db.post('/connect', async (req, res) => {
...
+db.post('/attach', async (req, res) => {
+  console.log('Calling /attach');
+  const {dbAlias, awsRegion, awsAccessKeyId, awsSecretAccessKey} = req.body;
+  if (!dbAlias || !awsRegion || !awsAccessKeyId || !awsSecretAccessKey) return res.status(400).json(
+    `Required key(s) not provided: ${[
+      'awsRegion', 'dbAlias', 'awsAccessKeyId', 'awsSecretAccessKey'
+    ].filter(k => !req.body.hasOwnProperty(k)).join(', ')}`
+  );
+  try {
+    const uid = dbMan.getUid(req.user);
```
dfellis, line 58: Wait, how does this work from their AWS lambda function?

aguillenv: I was already passing the auth token from the UI to the template.

aguillenv: https://github.com/iasql/iasql-engine/pull/587/files#diff-48dd46b95b19450a4573401c27f37a6bbc0a0806e002c71f8572bfd44c0c2ebcR16

dfellis: Got it. Very clever.

src/metadata/migration/1648476863240-init.ts
```diff
@@ -1,10 +1,10 @@
 import {MigrationInterface, QueryRunner} from "typeorm";
 
-export class init1647661129900 implements MigrationInterface {
-    name = 'init1647661129900'
+export class init1648476863240 implements MigrationInterface {
+    name = 'init1648476863240'
 
     public async up(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(`CREATE TABLE "iasql_database" ("pg_name" character varying NOT NULL, "alias" character varying NOT NULL, "region" character varying NOT NULL, "pg_user" character varying NOT NULL, CONSTRAINT "PK_a70daf5038ee78e1a4e5cb78990" PRIMARY KEY ("pg_name"))`);
+        await queryRunner.query(`CREATE TABLE "iasql_database" ("pg_name" character varying NOT NULL, "alias" character varying NOT NULL, "region" character varying NOT NULL, "pg_user" character varying NOT NULL, "is_ready" boolean NOT NULL DEFAULT true, CONSTRAINT "PK_a70daf5038ee78e1a4e5cb78990" PRIMARY KEY ("pg_name"))`);
```
dfellis, line 7: Next think: how are we going to get this change applied to the staging and prod metadata DBs? They persist between deploys, so we should probably migrate them the normal way?

aguillenv: Oh that's true. So I can create the migration and then we migrate them manually once landed, right?

dfellis: Yeah, or we can update our own deploy script to always run the migration after `iasql_apply`ing the new Fargate config?

src/metadata/entity/index.ts
```diff
@@ -20,7 +20,13 @@ export class IasqlDatabase {
...
 
   @Column()
-  pgUser: string
+  pgUser: string;
+
+  @Column({
+    // default to true for databases that already exists
+    default: true,
+  })
+  isReady: boolean;
```
depombo, line 29: would it be possible to know if a database was created with direct connect? perhaps don't default to `true` so that we can write a basic query where isReady is null that tells us how often each connection flow is used?

src/metadata/migration/1648476863240-init.ts
depombo: oh this actually treats migrations as inmutable in the more traditional sense. we should do a 2nd / additional migration and not modify the existing one like we do everywhere else. If we modify the existing migration typeorm might barf on a file checksum difference unless we blow up the table (and it's contents). we run all migrations on server startup so it should just work because it would recognize that it needs to run the 2nd migration

src/services/iasql.ts
```diff
@@ -130,26 +178,14 @@ export async function connect(
         }
       }
     }
-    await conn2.query(dbMan.newPostgresRoleQuery(dbUser, dbPass, dbId));
-    await conn2.query(dbMan.grantPostgresRoleQuery(dbUser));
-    logger.info('Done!');
     return {
       alias: dbAlias,
```
depombo, line 182: do we need to return the alias even though we provide it as a parameter?

src/metadata/migration/1648476863240-init.ts
aguillenv: sounds good

src/services/iasql.ts
depombo: just curious, this works

src/services/iasql.ts
```diff
@@ -59,21 +59,22 @@ const iasqlPlanV3 = (
...
     logger.info('Creating account for user...');
     const dbGen = dbMan.genUserAndPass();
     dbUser = dbGen[0];
     const dbPass = dbGen[1];
     dbId = dbMan.genDbId(dbAlias);
-    await MetadataRepo.saveDb(uid, email, dbAlias, dbId, dbUser, awsRegion);
-    logger.info('Establishing DB connections...');
+    const hasCredentials = !!awsAccessKeyId && !!awsSecretAccessKey;
+    await MetadataRepo.saveDb(uid, email, dbAlias, dbId, dbUser, awsRegion, hasCredentials, directConnect);
+    console.log('Establishing DB connections...');
```
dfellis, line 77:
```suggestion
    logger.info('Establishing DB connections...');
```

aguillenv: Oh I keep forgetting, thanks

src/services/iasql.ts
```diff
@@ -94,16 +95,69 @@ export async function connect(
...
+      // Attach credentials
+      await attach(uid, dbAlias, dbId, awsRegion, awsAccessKeyId, awsSecretAccessKey, conn2);
+    } else {
+      let counter = 0;
+      const checkCredInterval = setInterval(async () => {
+        const updated = await maybeUpdateStatus(uid, dbAlias, dbId);
+        if (updated) clearInterval(checkCredInterval);
+        counter++;
+        if (counter === 60) clearInterval(checkCredInterval); // try for 30 min
+      }, 30000);
```
depombo, line 108: this totally works for now, but we need two follow up tasks (not to be addressed in this PR):
1. the interval is not resilient to engine deploys. we need to move it to a graphile worker cron task. that task will also need to clean up databases that never make it to ready/active after a timeout.
2. we need to fail any iasql command that tries to talk to a pending database. also we should probably remove the `apply`/`sync` endpoints while we are at this?

depombo (approved):

depombo:
I like how much cleaner this approach is!

src/services/iasql.ts
aguillenv: Good points, definitely

PR #595, opened 2022-03-31
add aws_iam module

depombo:


src/modules/aws_iam@0.0.1/index.ts
```diff
@@ -0,0 +1,72 @@
...
+export const AwsIamModule: Module = new Module({
+  ...metadata,
+  utils: {
+    roleMapper: async (role: AWSRole, ctx: Context) => {
+      const out = new Role();
+      out.arn = role.Arn;
+      out.name = role.RoleName ?? '';
+      out.policyDocument = role.AssumeRolePolicyDocument ?? '';
+      const client = await ctx.getAwsClient() as AWS;
+      out.policyArns = await client.getRolePolicyArns(out.name);
```
aguillenv, line 17: why this? it supposed that the input param `role: AWSRole` must have the arn value

src/modules/aws_iam@0.0.1/index.ts
```diff
@@ -0,0 +1,72 @@
...
+      return out;
+    },
+  },
+  mappers: {
+    role: new Mapper<Role>({
+      entity: Role,
+      equals: (a: Role, b: Role) => Object.is(a.name, b.name) &&
+        Object.is(a.arn, b.arn) &&
+        Object.is(a.policyDocument, b.policyDocument) &&
```
aguillenv, line 26: is policy document a json as string? that could have some trouble if the json does not come always ordered from aws

dfellis: They probably store it as a string on their side (I haven't seen these docs mutated before, at least) but that is a fair point.

I don't think it is a likely problem, though, so maybe a comment that this depends on AWS not regenerating the JSON out from under us?

src/modules/aws_iam@0.0.1/index.ts
```diff
@@ -0,0 +1,72 @@
...
+        read: async (ctx: Context, ids?: string[]) => {
+          const client = await ctx.getAwsClient() as AWS;
+          const roles = Array.isArray(ids) ?
+            await Promise.all(ids.map(id => client.getRole(id))) :
+            (await client.getRoles()) ?? [];
+          return await Promise.all(roles
+            .map(r => AwsIamModule.utils.roleMapper(r, ctx))
+          );
+        },
+        update: async (es: Role[], ctx: Context) => {
```
dfellis, line 55: This update is only a replace, so can you add the `updateOrReplace` property to always return replace?

src/modules/aws_iam@0.0.1/index.ts
depombo: yes, I let me use that

src/modules/aws_iam@0.0.1/index.ts
depombo: that is a good point. I will add a comment for now and test to see if it might be problematic

src/modules/aws_iam@0.0.1/index.ts
depombo: Let me look at the module interface

src/modules/aws_iam@0.0.1/index.ts
aguillenv: oh I think I was confused. ok if the role policy earns does not come from AWS when retrieving the role and you need a second call I guess this is ok

src/modules/aws_iam@0.0.1/index.ts
depombo: I think this is also true for ec2, so let me change it there too

src/modules/aws_ecs_fargate@0.0.1/index.ts
```diff
@@ -228,12 +228,12 @@ export const AwsEcsFargateModule: Module = new Module({
     taskDefinition: new Mapper<TaskDefinition>({
       entity: TaskDefinition,
       equals: (a: TaskDefinition, b: TaskDefinition) => Object.is(a.cpuMemory, b.cpuMemory)
-        && Object.is(a.executionRoleArn, b.executionRoleArn)
+        && AwsIamModule.utils.roleEq(a.executionRole, b.executionRole)
```
aguillenv, line 231: This is ok but checking the `arns` should be enough?

src/modules/aws_iam@0.0.1/entity/role.ts
```diff
@@ -0,0 +1,32 @@
...
+  @Column({
+    nullable: true,
+  })
+  @cloudId
+  arn?: string;
+
+  // Guaranteed unique in AWS
+  // Maximum 128 characters. Use alphanumeric and '+=,.@-_' characters.
+  @PrimaryColumn()
+  name: string;
```
aguillenv, line 21: Hmm I'm not sure about this, I think this should not be decorated and add the following property to keep consistency between entities: 
```
@PrimaryGeneratedColumn()
  id?: number;
```

src/modules/aws_ecs_fargate@0.0.1/index.ts
depombo: actually, the names are guaranteed unique. I did this before realizing that

src/modules/aws_iam@0.0.1/entity/role.ts
depombo: why? the role name is guaranteed unique in aws, so it simplifies references to the table and reduces aws calls since all their methods just take the role name

src/modules/aws_iam@0.0.1/entity/role.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  @Column()
+  policy: string;
+
+  @ManyToOne(() => Role)
+  @JoinColumn({
+    name: 'role_name',
+  })
+  role: Role;
+}
```
aguillenv, line 61: I still think we are not being consistent with other modules not having the auto incremental `id` column in these entities

depombo: the role name is guaranteed unique in aws, so it simplifies references to the table and reduces aws calls since all their methods just take the role name. in some cases we need the id, but we shouldn't have it if adds complexity? there are a few modules where we could remove it too

dfellis: Because it's set by the user instead of set by AWS, I think it's fine? We did the incrementing IDs on the others because the unique ID for the records was controlled by AWS for the other modules.

yarn.lock
```diff
@@ -1,6126 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
dfellis, line 1: This file shouldn't be deleted

dfellis (approved):
Please restore the lockfile (again) :)

depombo:
> Please restore the lockfile (again) :)

It is restored. I did it because I also thought it got deleted again by mistake. It mostly seems to be deleting lines?

depombo:
On the branch
```
iasql-engine iam % yarn install
yarn install v1.22.18
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.17s.
iasql-engine iam % git diff
iasql-engine iam % 
```

Issue #574, opened 2022-03-28
Categorize errors on `iasql_apply` to better print each type


depombo:
Current message details
```
  detail: '{"message":"Forward progress halted. Some promises never resolved","metadata":{}}',
  context: 'PL/pgSQL function until_iasql_operation(iasql_operation_optype_enum,text[]) line 39 at RAISE\n' +
    'PL/pgSQL function iasql_cloud_manipulation(iasql_operation_optype_enum) line 5 at assignment\n' +
    'PL/pgSQL function iasql_apply() line 3 at RETURN QUERY',
  file: 'pl_exec.c',
  line: '3868',
  routine: 'exec_stmt_raise'
```

depombo:
- [ ] Right now metadata is coming out empty for errors
- [ ] Categorize errors so better print each type

aguillenv:
should we close this based on #611

dfellis:
I think so.

PR #598, opened 2022-04-01
Return error metadata

aguillenv:
Part of #574 

image

### DIFF
--- a/src/services/logger.ts
+++ b/src/services/logger.ts
@@ -38,4 +38,22 @@ export function logUserErr(e: any): string {
   return err;
 }
 
+function exposeErrorProperties(error: Error) {
+  const copy = Object.assign({}, error);
+  if (error.name) copy.name = error.name;
+  if (error.message) copy.message = error.message;
+  if (error.stack) copy.stack = error.stack;
+  return copy;
+}
+
+export function errorReplacer(_key: string, value: any) {
+	if (value instanceof Error) {
+		return exposeErrorProperties(value);
+	} else if (Array.isArray(value)) {
+    return value.map(val => JSON.stringify(val, errorReplacer))
+	} else {
+		return value;
+	}
+}
+
 export default singleton;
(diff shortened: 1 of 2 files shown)


src/services/scheduler.ts
```diff
@@ -83,7 +83,13 @@ export async function start(dbId: string, dbUser:string) {
...
           logUserErr(e);
           // error must be valid JSON as a string
-          const error = JSON.stringify(e, Object.getOwnPropertyNames(e));
+          const error = JSON.stringify(e, (_key, value) => {
+            if (Array.isArray(value)) {
+              return value.map(val => JSON.stringify(val, Object.getOwnPropertyNames(val)))
+            } else {
+              return value
+            }
+          });
```
dfellis, line 92: I'd recommend turning this into a helper function and on the `if (Array.isArray(value))` path, it calls itself (to handle multiple-levels of array nesting.

aguillenv: for the recursiveness had to take inspiration from [here](https://www.bennadel.com/blog/3278-using-json-stringify-replacer-function-to-recursively-serialize-and-sanitize-log-data.htm) because it was not working. If you just return the `Error` it appear as an empty object in the string.

dfellis (approved):

src/services/logger.ts
```diff
@@ -38,4 +38,22 @@ export function logUserErr(e: any): string {
...
+}
+
+export function errorReplacer(_key: string, value: any) {
+	if (value instanceof Error) {
+		return exposeErrorProperties(value);
+	} else if (Array.isArray(value)) {
+    return value.map(val => JSON.stringify(val, errorReplacer))
+	} else {
+		return value;
+	}
```
depombo, line 56: nit: can we have indentation match the rest of the file?

depombo (approved):

src/services/logger.ts
aguillenv: oops

PR #604, opened 2022-04-04
turn off auth for simple integration test

depombo:


dfellis (approved):
For future reference, we decided the staging environment tests this pass well enough.

aguillenv (approved):

depombo:
> For future reference, we decided the staging environment tests this pass well enough.

Yes, also worth pointing out that there is less branching between authenticated and non-authenticated API calls since we introduced the metadata repository

PR #608, opened 2022-04-04
iasql_platform and iasql_functions modules (WIP)

dfellis:


src/modules/iasql_platform@0.0.1/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  ...metadata,
+  mappers: {
+    iasqlModule: new Mapper<IasqlModule>({
+      entity: IasqlModule,
+      equals: (_a: IasqlModule, _b: IasqlModule) => true,
+      source: 'db',
+      cloud: new Crud({
+        // We don't actually connect to AWS for this module, because it's meta
+        create: async (_e: IasqlModule[], _ctx: Context) => { /* Do nothing */ },
+        read: (ctx: Context, names?: string[]) => ctx.orm.find(IasqlModule, names ? {
```
aguillenv, line 18: is it not weird to call the database in the `cloud` mapper definition?

aguillenv (commented):
looking good!

dfellis:
Screenshot from 2022-04-05 11-22-11

We're in the home stretch for this PR. We now have a way to upgrade the platform both internally through the module system and triggerable by the user with a postgres function that we will fill in behavior for in the future.

src/modules/interfaces.ts
```diff
@@ -1,4 +1,4 @@
-import { readdirSync, } from 'fs'
+import fs from 'fs'
```
aguillenv, line 1: out of curiosity, any reason for this change?

aguillenv (approved):

src/modules/interfaces.ts
dfellis: Easier to grep for `fs\.` since that module is inconsistent on method capitalization.

But the change also happened because I thought I could remove it but in the end couldn't.

src/modules/iasql_platform@0.0.1/index.ts
```diff
@@ -0,0 +1,9 @@
+/* THIS MODULE IS A SPECIAL SNOWFLAKE. DON'T LOOK AT IT FOR HOW TO WRITE A REAL MODULE */
```
depombo, line 1: ❄️

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Infrastructure as SQL",
   "main": "dist/index.js",
   "scripts": {
-    "build": "yarn clean && tsc",
+    "build": "yarn clean && tsc && copyfiles -u 1 src/**/*.sql dist/",
```
depombo, line 7: does this copy it into `dist/src/**/*.sql` or just `dist/*.sql`?

depombo (approved):

package.json
dfellis: In the exact same directory structure as it was in before. :) I found out about it in a blog post with someone needing something similar for a web build process but not wanting to switch to babel.

PR #612, opened 2022-04-05
Better test logging

dfellis:
Debugging what's going wrong with a test is hard right now because Jest's console.log takeover code makes it ridiculously hard to read the log output. So I've bypassed it using `process.[stdout|stderr].write` directly. Just take a look at the log output of the tests in *this* PR to see the massive improvement


depombo (approved):

aguillenv (commented):
Nice!!

PR #613, opened 2022-04-07
Fix gen-module by running migrations in dep order ourselves

dfellis:
This updates `gen-module` to use our own logic for module ordering and runs the migrations in that order.

Despite that, there's still some weirdness with `iasql_platform` that I have been able to confirm happens even using TypeORM's migration run and generate code the "proper" way when I blew away all modules besides that one and `aws_account` and *still* got weird pollution from it. It appears to be a regression of some sort in TypeORM because it didn't happen before. I'm thinking of making a simple repo to minimally reproduce the issue and reporting it to TypeORM so I can hopefully remove the weird hackery in my scripts around the `iasql_platform` module.

### DIFF
--- a/.module_migration.sh
+++ b/.module_migration.sh
@@ -22,12 +22,9 @@ psql postgresql://postgres:test@localhost:5432/postgres -c "CREATE DATABASE __ex
 # Now, we need to move out of the root directory to pull our trickery below
 cd src
 
-# Blow away the existing migration for the specified module, if one exists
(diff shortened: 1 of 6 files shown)


dfellis:
Screenshot from 2022-04-06 22-15-10

Proof it works (generated an identical migration for `aws_account`

depombo (approved):

aguillenv (approved):

PR #614, opened 2022-04-07
get rid of db mappers

depombo:


dfellis (approved):
Very nice! I figured you wouldn't be able to get rid of the security group db mapper since that module is so weird and there's the TypeORM bug with circular references (even if not circular in the DB itself), but this is a nice net-negative PR :D

aguillenv (commented):
Looks good but I'm worry about the cascades. Did you confirm locally it works as expected? Did you try a full create, update and delete manual tests for the new cascades and check the database to see is doing what we expect? Even if the tests are passing, cascades could be introducing duplicates and loading the tables.

dfellis:
> > Looks good but I'm worry about the cascades. Did you confirm locally it works as expected? Did you try a full create, update and delete manual tests for the new cascades and check the database to see is doing what we expect? Even if the tests are passing, cascades could be introducing duplicates and loading the tables.
> 
> I did not do manual tests, but shouldn't this be the purpose of test suite? We check counts on specific resource where clauses. If this is a concern then we should add more of these checks where they are missing, if they are missing?

I agree with @depombo here. The tests should be covering that and if they aren't we should fix that, too. Personally think we should land this; if it does cause a regression like that we should see it in our end-to-end testing with the quickstart examples, I think?

depombo:
Added more count checks around the entities that have cascade

test/modules/aws-ecs-integration.ts
```diff
@@ -378,6 +492,14 @@ describe('ECS Integration Testing', () => {
       WHERE service.name = '${serviceRepositoryName}';
     `, (res: any[]) => expect(res.length).toBe(1)));
 
+    it('applies service insertion', apply());
```
depombo, line 495: we actually were never applying the service insertion here which increased the test time

test/modules/aws-ecs-integration.ts
```diff
@@ -489,6 +653,14 @@ describe('ECS Integration Testing', () => {
       WHERE service.name = '${servicePublicRepositoryName}';
     `, (res: any[]) => expect(res.length).toBe(1)));
 
+    it('applies service insertion', apply());
```
depombo, line 656: we actually were never applying the service insertion here which increased the test time

aguillenv (approved):

PR #615, opened 2022-04-07
support aws service roles

depombo:


src/modules/aws_ecs_fargate@0.0.1/index.ts
```diff
@@ -95,18 +95,20 @@ export const AwsEcsFargateModule: Module = new Module({
       }
       out.cpuMemory = `vCPU${+(td.cpu ?? '256') / 1024}-${+(td.memory ?? '512') / 1024}GB` as CpuMemCombination;
       if (td.executionRoleArn) {
-        const executionRoleName = td.executionRoleArn.split(':role/')[1];
-        out.executionRole = await AwsIamModule.mappers.role.db.read(ctx, executionRoleName) ??
-          await AwsIamModule.mappers.role.cloud.read(ctx, executionRoleName);
+        const roleName = AwsIamModule.utils.roleNameFromArn(td.executionRoleArn);
+        // there can be hundreds of task defintions so don't do an aws call for each
+        out.executionRole = await AwsIamModule.mappers.role.db.read(ctx, roleName) ??
+          ctx?.memo?.cloud?.Role?.[roleName ?? ''];
```
aguillenv, line 101: If I'm not wrong the `await AwsIamModule.mappers.role.cloud.read(ctx, executionRoleName);` first checks the `memo` and if it is not there then calls the cloud. This way if roles are not loaded yet in `memo` we will not link the task with the right role, I think?

depombo: hmm the problem is if the role doesn't exist in the cloud then we make a request every time ```?? await AwsIamModule.mappers.role.cloud.read(ctx, executionRoleName)``` hmm maybe I can set the cache when this happens somehow??

depombo: how about something like
```ts
        try {
          out.taskRole = await AwsIamModule.mappers.role.db.read(ctx, roleName) ??
            await AwsIamModule.mappers.role.cloud.read(ctx, roleName);
        } catch (e) {
          // Role could have been deleted
          logger.error(`Role ${roleName} not found`, e as any);
          out.taskRole = undefined;
          // populate the cache so it doesn't fail other refs since
          // there can be hundreds of task definitions
          ctx.memo.cloud.Role[roleName] = {};
        }
        ```

depombo: discussed offline at length and arrived to this https://github.com/iasql/iasql-engine/pull/615/commits/f37c6f34da9cb2256bec8206e71f8de3f6c1a6b6. follow up task https://github.com/iasql/iasql-engine/issues/616

dfellis (approved):

PR #622, opened 2022-04-08
Fix gen-module migration creation for brand new modules

dfellis:


dfellis:
To test this, I both re-ran it on `aws_account` and I created a fake `foo` module:

Screenshot from 2022-04-08 09-24-56

The module migration generation now takes 17 seconds on my machine:

Screenshot from 2022-04-08 09-25-20

And the result is a success!

Screenshot from 2022-04-08 09-24-33

.module_migration.sh
```diff
@@ -9,13 +9,14 @@ MODULE=$1
...
 docker volume prune -f
 docker image prune -f
 
-# Next, spin them back up so we have a database
-docker-compose up --build --detach
+# Start a postgres container
+docker container run -p 5432:5432 -e POSTGRES_PASSWORD=test --name migrate-postgres -d postgres:13.4
+
 sleep 5 # Just in case
 psql postgresql://postgres:test@localhost:5432/postgres -c "CREATE DATABASE __example__"
```
aguillenv, line 21: forgot to say it in the previous PR but this could be `__iasql_migration__`

aguillenv (approved):

.module_migration.sh
dfellis: They just need to match, though. :)

PR #624, opened 2022-04-09
Use the resource name wherever possible

depombo:
This makes it a lot easier to write SQL without doing join or nested selects. This PR does not include security groups. It is rebased on top of https://github.com/iasql/iasql-engine/pull/614


aguillenv (approved):
It definitely looks a lot simpler and clean. For the kind of operation we do I do not think this will suppose a performance issue, not sure if will be for someone with a massive AWS account (joins with numeric/serial values perform a bit faster than varchars). Now in terms of SQL purists folks not sure if they are gonna like it :smile:

dfellis (approved):
:100:

PR #628, opened 2022-04-09
include record count when listing databases

depombo:


dfellis (approved):

aguillenv (approved):

src/entity/index.ts
```diff
@@ -32,6 +34,17 @@ export class IasqlDatabase {
...
+  @Column({
+    type: 'int',
+  })
+  megabytes: number;
+
+  @CreateDateColumn()
+  createdAt: Date;
+
+  @UpdateDateColumn()
+  updatedAt: Date;
```
aguillenv, line 47: Does this updates on every change automatically?

depombo: yes

dfellis: Well, it does in the ORM. It's clear in the migration that Postgres itself doesn't have this functionality, unfortunately.

dfellis (approved):

PR #634, opened 2022-04-11
Wait for LoadBalancer to be fully terminated

### LINKED ISSUE #489: `cloud.delete` path should wait until resources are terminated
the assumption with the lazyLoader is that once a task successfully completes, it is actually successfully completed, not just "successfully staged to be completed" so if the `cloud.delete` path returns before it is actually deleted this happens

Screen Shot 2022-03-09 at 10 14 30 AM

### LINKED ISSUE #590: Intermittent error applying/reverting migrations on ECS Fargate example
https://sentry.io/organizations/iasql/issues/3141742881/?project=6106929&referrer=webhooks_plugin

https://sentry.io/organizations/iasql/issues/3141648924/?project=6106929&query=is%3Aunresolved

aguillenv:
Resolves #489
Resolves #590

### DIFF
--- a/test/modules/aws-elb-integration.ts
+++ b/test/modules/aws-elb-integration.ts
@@ -229,7 +229,7 @@ describe('ELB Integration Testing', () => {
   it('check load_balancer delete', query(`
     SELECT *
     FROM security_group
-    WHERE group_name = ANY(array['${sg1}', '${sg2}']);
+    WHERE group_name IN ('${sg1}', '${sg2}');
   `, (res: any[]) => expect(res.length).toBe(0)));
 
   it('applies the change', apply());
--- a/test/modules/aws-elb-integration.ts
+++ b/test/modules/aws-elb-integration.ts
@@ -22,6 +22,8 @@ const protocol = ProtocolEnum.HTTP;
 const lbScheme = LoadBalancerSchemeEnum.INTERNET_FACING;
 const lbType = LoadBalancerTypeEnum.APPLICATION;
 const lbIPAddressType = IpAddressType.IPV4;
+const sg1 = `${prefix}${dbAlias}lbsg1`;
+const sg2 = `${prefix}${dbAlias}lbsg2`;
 
 jest.setTimeout(360000);
 beforeAll(async () => await execComposeUp());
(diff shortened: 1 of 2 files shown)


test/modules/aws-elb-integration.ts
```diff
@@ -193,6 +221,16 @@ describe('ELB Integration Testing', () => {
...
 
+  it('deletes the security group 1', query(`
+    DELETE FROM security_group
+    WHERE group_name = '${sg1}';
+  `));
+
+  it('deletes the security group 2', query(`
+    DELETE FROM security_group
+    WHERE group_name = '${sg2}';
+  `));
```
dfellis, line 232: Nit: why not do these both in the same query? `WHERE group_name in ('${sg1}', '${sg2}')`

dfellis (approved):

test/modules/aws-elb-integration.ts
```diff
@@ -221,15 +221,16 @@ describe('ELB Integration Testing', () => {
     WHERE load_balancer_name = '${lbName}';
   `, (res: any[]) => expect(res.length).toBe(0)));
 
-  it('deletes the security group 1', query(`
+  it('deletes the security groups', query(`
     DELETE FROM security_group
-    WHERE group_name = '${sg1}';
+    WHERE group_name = ANY(array['${sg1}', '${sg2}']);
```
dfellis, line 226: You shouldn't have to do all of this casting with Postgres arrays, this should work:

```suggestion
    WHERE group_name IN ('${sg1}', '${sg2}');
```

test/modules/aws-elb-integration.ts
```diff
@@ -221,15 +221,16 @@ describe('ELB Integration Testing', () => {
...
   `));
 
-  it('deletes the security group 2', query(`
-    DELETE FROM security_group
-    WHERE group_name = '${sg2}';
-  `));
+  it('check load_balancer delete', query(`
+    SELECT *
+    FROM security_group
+    WHERE group_name = ANY(array['${sg1}', '${sg2}']);
```
dfellis, line 232: Same

```suggestion
    WHERE group_name IN ('${sg1}', '${sg2}');
```

This is standard SQL syntax and I would be very surprised if it doesn't work (especially since at least one of the `iasql_*` functions uses it ;)

test/modules/aws-elb-integration.ts
aguillenv: Oh I wasn't sure that worked on Postgres

PR #645, opened 2022-04-11
cascade container defs on task def update/delete

depombo:


src/modules/aws_ecs_fargate@0.0.1/entity/container_definition.ts
```diff
@@ -31,7 +31,11 @@ export class ContainerDefinition {
   @Column()
   name: string;
 
-  @ManyToOne(() => TaskDefinition)
+  @ManyToOne(() => TaskDefinition, {
+    // if the parent task def is deleted, also delete this container def
```
dfellis, line 35: Comment seems slightly out of sync with the code here?

dfellis (approved):
Minor comment, but LGTM assuming the end-to-end test works, too :)

src/modules/aws_ecs_fargate@0.0.1/entity/container_definition.ts
depombo: typeorm is a bit funny in that you define this in the child as opposed to the parent which took me some time to realize :/

dfellis: So I mean you're also doing it if the parent task definition is updated, afaict?

depombo: ooh yes, that part of the comment is incomplete. I see what you mean, will update in the next PR

PR #646, opened 2022-04-12
Key security_group by name

dfellis:
- Switch the security groups to be keyed by the group name
- Got ec2 working
- Fixed up elb
- Get RDS working
- Get fargate working


src/modules/aws_security_group@0.0.1/entity/index.ts
```diff
@@ -12,19 +13,14 @@ import { cloudId, } from '../../../services/cloud-id'
 
 @Entity()
 export class SecurityGroup {
-  @PrimaryGeneratedColumn()
-  id?: number;
+  @PrimaryColumn()
+  groupName?: string;
```
aguillenv, line 17: Should not be nullable anymore, right?
```suggestion
  groupName: string;
```

aguillenv (approved):

dfellis:
Abandoning this. The group name field is not actually unique when there's more than one VPC. Need to make a different change to the security groups.

PR #652, opened 2022-04-13
Fix issue with VPCs created without IaSQL's knowledge from crapping up the security groups

dfellis:


src/modules/aws_security_group@0.0.1/index.ts
```diff
@@ -107,7 +107,12 @@ export const AwsSecurityGroupModule: Module = new Module({
         Object.is(a.vpc?.vpcId, b.vpc?.vpcId),
       source: 'db',
       db: new Crud({
-        create: (e: SecurityGroup[], ctx: Context) => ctx.orm.save(SecurityGroup, e),
+        create: async (e: SecurityGroup[], ctx: Context) => {
+          for (const out of e) {
+            if (out.vpc && !out.vpc.id) await AwsVpcModule.mappers.vpc.db.create(ctx, out.vpc);
```
depombo, line 112: can we leave a comment here?

test/modules/aws-security-group-integration.ts
```diff
@@ -234,11 +235,28 @@ describe('Security Group install/uninstall', () => {
...
+    INSERT INTO vpc (cidr_block)
+    VALUES ('192.${randIPBlock}.0.0/16');
+  `));
+
+  it('creates the VPC', apply());
+
   it('installs the Security Group module', install(
-    ['aws_security_group', 'aws_vpc']));
+    ['aws_security_group']));
+
```
depombo, line 250: should we check the new default SG exists here?

src/modules/aws_security_group@0.0.1/index.ts
dfellis: Sure

test/modules/aws-security-group-integration.ts
dfellis: Well, I was relying on the install itself not crapping the bed as the test for this, but I can see about how to do that.

PR #658, opened 2022-04-13
Get ELB to depend on VPC correctly (rather than just an unchecked string)

dfellis:


src/modules/aws_elb@0.0.1/entity/target_group.ts
```diff
@@ -71,8 +74,14 @@ export class TargetGroup {
   })
   port?: number;
 
-  @Column()
-  vpc: string;
+  @ManyToOne(() => Vpc, {
+    nullable: true,
```
depombo, line 78: I thought that with the join column this would not be nullable and in the case of the default vpc we would link to it's corresponding db id? Having a null value point to the `default` vpc might not be obvious for users specially because we don't have this behavior for sg?

dfellis: So the foreign key is allowed to be nullable onto the other table, but the key in the other table can't be nullable (unless you only allow a single null value) because it needs to be unique for the join. This is simply saying "I can't join tot anything"

And we're using this as a trick to make it easier to do if you just want to use the default VPC by having the mapper stitch it together in that situation.

depombo: I didn't understand why we didn't do this for security groups and once you mentioned on Discord that there can be more than one `default` sg in a single account+region if there is more than one vpc it made sense.

depombo (approved):

PR #659, opened 2022-04-13
Add extra waiter for ECS

aguillenv:


src/services/gateways/aws.ts
```diff
@@ -1104,6 +1104,20 @@ export class AWS {
         }
       },
     );
+    // Force retry for ~4 min to unlock security group
+    await createWaiter<ECSClient, DescribeServicesCommand>(
```
depombo, line 1108: I'm not sure I follow. when you say `unlock` do you mean wait until the sg gets deleted? also why not just poll the sg directly?

depombo: nvm just saw the PR description

aguillenv: No the issue we have deleting security groups is not with security groups itself, but it is that other resources like elb or ecs use security groups. Even when it's supposed that those resources are completely deleted and we are good to go to delete security groups, we are not. And it is because these other resources (elb and ecs in this case) are creating a network interface behind scenes that are "locking" or preventing the security group to be deleted. That network interface will be deleted eventually but for ecs I haven't found how to fully await until it is gone, for elb the issue is already solved the correct way.

dfellis (approved):
35 minutes for that test!!

aguillenv:
> 35 minutes for that test!!

4 service deletions in the test
- 901826ms
- 56104ms
- 93926ms
- 676381ms

:/ Sometimes, I do not know why it takes toooooo much time

PR #662, opened 2022-04-15
Build in graphql support with automatic schema loading via Postgraphile. Also organize the configuration for easier understanding

dfellis:
Postgraphile was easier to set up than I expected, much faster to spin up than I expected, and we have much more control over it than Hasura. Still limiting things to just the local environment for now, but we can think about how to expose this in a safe, controlled way. (It *should* still require Auth0 auth in those environments with how I set it up, but that's not tested.)

I also made the config nicer (and documented) in the process because the big blob of a single JSON object is hard to follow and not necessary.


dfellis:
Screenshot from 2022-04-15 10-39-53

dfellis:
Method calling is via GraphQL `mutation` and it also works out of the box (and converts `snake_case` to `camelCase` so it looks normal to GraphQL users

Screenshot from 2022-04-15 11-07-58

src/config/staging.ts
```diff
@@ -1,19 +1,28 @@
...
+    user: process.env.DB_USER ?? throwError('No DB User defined'),
+    password: process.env.DB_PASSWORD ?? throwError('No DB Password defined'),
+    port: 5432,
+    forceSSL: true,
+  },
+  logger: {
+    debug: true,
+    test: false,
+  },
+  auth0: {
```
depombo, line 19: maybe let's just do auth due to typos in the past between authO and auth0?

src/router/graphql.ts
```diff
@@ -0,0 +1,33 @@
...
+      {
+        externalUrlBase: `/v1/graphql/${db}`,
+        graphiql: withGraphiql,
+      });
+    return postgraphiles[db];
+  };
+
+  graphql.use('/:db', async (req, res, next) => {
+    const db = config.auth0 ?
+      // This will throw an error if the user is not allowed to access the specified DB
```
depombo, line 28: there is no try catch, does the error crash the process?

depombo (approved):

src/router/graphql.ts
dfellis: I assumed this would handle it? https://github.com/iasql/iasql-engine/blob/main/src/index.ts#L43-L52

It's a thrown error from an Express router and I thought Express grabbed the exceptions for us to route into that thing.

src/config/staging.ts
dfellis: Sure

src/router/graphql.ts
depombo: Got it, I think we should be fine then!

depombo (approved):

depombo (approved):

src/config/config.ts
```diff
@@ -1,17 +1,34 @@
...
+    test: boolean; // Whether or not a special test logger is enabled (bypass weirdness with Jest)
+  }
+  // Configuration for auth0 access control
+  auth?: { // Not including this sub-object implies it is not enabled
+    domain: string;
+    audience: string;
+  };
+  // Configuration for sentry error reporting
+  sentry?: { // Not including this sub-object implies it is not enabled
+    dsn: string;
```
aguillenv, line 26: shouldn't this be `dns`? 😄

dfellis: Nope. It's DSN. [Sentry Data Source Name](https://docs.sentry.io/product/sentry-basics/dsn-explainer/)

PR #672, opened 2022-04-18
Fix subnets from default vpc getting deleted

### LINKED ISSUE #671: Default subnets getting deleted when running clean script

aguillenv:
Resolves #671 

A race condition is occurring when checking the memo. It is being checked while a placeholder is in place so a empty object is being assigned.

What it should happen always:
`info vpc db object undefined`

What it is happening sometimes:
`info vpc db object {}`


src/modules/interfaces.ts
```diff
@@ -147,6 +147,11 @@ export class Crud<E> {
           logger.info(`Cache miss for ${this.entity?.name ?? ''} ${this.dest}`);
           ctx.memo[dest][entityName][id] = new (this.entity as new () => E)();
         } else {
+          // If object is empty it means it is a placeholder and it is not in the memo yet.
+          if (!Object.keys(ctx.memo[dest][entityName][id]).length) {
+            logger.info(`Cache miss for ${this.entity?.name ?? ''} ${this.dest}`);
+            return undefined;
+          }
```
dfellis, line 154: If this is staying in, should we just drop the placeholder logic? They aren't "getting through" to the users so they're just excessive complexity at this point?

dfellis (approved):

src/modules/interfaces.ts
aguillenv: I'll create a follow up task with this and try to tackle in a new PR.

depombo (approved):

PR #678, opened 2022-04-19
Sort dependencies and make lazy loader linear

aguillenv:


src/services/iasql.ts
```diff
@@ -496,6 +498,15 @@ export async function apply(dbId: string, dryRun: boolean, ormOpt?: TypeormWrapp
                 }
               }));
             }
+            return outArr;
+          })
+          .flat(9001);
+        const reversePromiseGenerators = records
```
dfellis, line 504: Nice splitting the deletes from the creates that way. Update-as-Replace is in a weird spot here, but I think it should eventually converge since the core of the algo is still the same.

src/services/lazy-dep.ts
```diff
@@ -21,17 +21,17 @@ export async function lazyLoader(promiseGenerators: (() => Promise<any>)[]) {
...
-    for (let i = 0; i < results.length; i++) {
-      if (results[i].status === 'fulfilled') {
-        continue;
+    for (const g of generatorsToRun) {
+      try {
+        await g();
+      } catch (e: any) {
+        logger.info(`error object ${JSON.stringify(e)}`)
+        const err = new Error((e ?? 'An unexpected error occurred'));
+        err.stack = e.stack ?? err.stack;
```
dfellis, line 31: Nit. I think this could be:
```suggestion
        const err = e ?? new Error('An unexpected error occurred');
```

dfellis: Though I see the weird structure was there before. I wonder where it came from and why.

dfellis (approved):

depombo (approved):

PR #682, opened 2022-04-21
Fixes for postgraphile in staging, and enable production

dfellis:
Before landing my dashboard change to allow module management in the GUI, I deployed the branch to staging and noticed that I could crash the engine, so this is the fix.

- Fix sslmode for postgraphile
- Enable in production now that it was verified with staging

### DIFF
--- a/src/router/graphql.ts
+++ b/src/router/graphql.ts
@@ -29,7 +29,7 @@ if (config.graphql) {
     postgraphiles[db] = postgraphiles[db] ?? {
       lastAccessMs: Date.now(),
       graphile: postgraphile(
-        `postgresql://${config.db.user}:${config.db.password}@${config.db.host}/${db}${config.db.forceSSL ? '?sslmode=require' : ''}`,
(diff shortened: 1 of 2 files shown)


src/router/graphql.ts
```diff
@@ -29,7 +29,7 @@ if (config.graphql) {
     postgraphiles[db] = postgraphiles[db] ?? {
       lastAccessMs: Date.now(),
       graphile: postgraphile(
-        `postgresql://${config.db.user}:${config.db.password}@${config.db.host}/${db}${config.db.forceSSL ? '?sslmode=require' : ''}`,
+        `postgresql://${config.db.user}:${config.db.password}@${config.db.host}/${db}${config.db.forceSSL ? '?sslmode=no-verify' : ''}`,
```
aguillenv, line 32: would this mode still force the ssl connection?

aguillenv (approved):

src/router/graphql.ts
dfellis: It forces an ssl connection even if the ssl cert is wrong (which it is, because RDS doesn't let us change it and is why our long-term plan is to run our own Postgres servers instead of using RDS). I thought the prior one would work, but apparently it meant "crash if the cert is bad."

PR #683, opened 2022-04-23
Remove the aws_account special casing: it's just another module, now

dfellis:


src/modules/iasql_functions@0.0.1/sql/create_fns.sql
```diff
@@ -10,9 +10,6 @@ declare
     _db_id text;
     _dblink_conn_count int;
 begin
-    if (select region from aws_account limit 1) is null then
```
aguillenv, line 13: Shouldn't we keep checking if we have something in `aws_account`?

test/modules/aws-ec2-integration.ts
```diff
@@ -11,7 +11,9 @@ const ubuntuAmiId = 'ami-0892d3c7ee96c0bf7';
 const apply = runApply.bind(null, dbAlias);
 const sync = runSync.bind(null, dbAlias);
 const query = runQuery.bind(null, dbAlias);
+const query2 = runQuery.bind(null, `${dbAlias}_sync`);
 const install = runInstall.bind(null, dbAlias);
+const install2 = runInstall.bind(null, `${dbAlias}_sync`);
```
aguillenv, line 16: Maybe `query_sync` and `install_sync` are more descriptive?

test/modules/aws-ecs-ecr-integration.ts
```diff
@@ -12,6 +12,8 @@ const apply = runApply.bind(null, dbAlias);
 const sync = runSync.bind(null, dbAlias);
 const query = runQuery.bind(null, dbAlias);
 const install = runInstall.bind(null, dbAlias);
+const query2 = runQuery.bind(null, dbAliasSidecar);
+const install2 = runInstall.bind(null, dbAliasSidecar);
```
aguillenv, line 16: Same

src/modules/iasql_functions@0.0.1/sql/create_fns.sql
dfellis: No? We would be unable to support module management and etc for GCP/Azure/etc if we keep this. It also never used these values for any reason, so it was just unnecessary gatekeeping.

aguillenv: I get your point, would need to exists a similar guard for each cloud provider and probably it does not makes sense. How it be handle the case where the `aws_account` modules is not installed or it has no records? that would break things or it is being handled in other part of the code already?

dfellis: > I get your point, would need to exists a similar guard for each cloud provider and probably it does not makes sense. How it be handle the case where the `aws_account` modules is not installed or it has no records? that would break things or it is being handled in other part of the code already?

So, the `aws_account` module *has* to be installed before you can install other aws modules since they depend on it. But on top of that, you can't successfully install any other aws module until there's a single row in the `aws_account` table with your credentials because the automatic sync-on-install for that module would fail.

It is still *possible* for someone to break it after-the-fact by deleting the `aws_account` record after installing other modules, but I don't think we should block all iasql stored proc access on that, especially if/when we add some sort of function for managing these "no cloud side" modules in a more general way.

src/router/graphql.ts
```diff
@@ -26,17 +26,26 @@ if (config.graphql) {
...
+          'public',
+          {
+            externalUrlBase: `/v1/graphql/${db}`,
+            graphiql: withGraphiql,
+            watchPg: true,
+          }),
+      };
+    } catch(e) {
+      // If postgraphile fails during construction, that error could crash the server. Instead
+      // we pass back an Express middleware that always fails
```
dfellis, line 43: I tried to connect to a local test database before I created it and crashed the server, so I added the fix to this PR to be quicker.

src/services/iasql.ts
```diff
@@ -86,14 +83,11 @@ export async function connect(
...
     const dbPass = dbGen[1];
     dbId = dbMan.genDbId(dbAlias);
-    const hasCredentials = !!awsAccessKeyId && !!awsSecretAccessKey;
     const metaDb = new IasqlDatabase();
     metaDb.alias = dbAlias;
     metaDb.pgUser = dbUser;
-    metaDb.region = awsRegion;
     metaDb.pgName = dbId;
     metaDb.directConnect = directConnect
-    metaDb.isReady = hasCredentials;
```
depombo, line 96: So removing this here will make the field in the database always default to true, which would mean that databases that use direct connect will be immediately `ready` in the dashboard when they are not.

src/modules/iasql_functions@0.0.1/sql/create_fns.sql
depombo: I added this here before the `is_ready` field existed in the metadata repository for `iasql_database`. I am okay with removing this logic because eventually we would have more clouds or providers other than AWS. Right now the `is_ready` field would not function properly though, so we need to remove it, revive it, introduce a special kind of provide module or perhaps provider credentials simple live in `iasql_metadata`? `aws_account` is not a normal module and neither would a `heroku_account`, `gcp_account`, etc? That can be a follow up once we need to add more clouds though and maybe we just remove `is_ready` from the metadata and derive it in the dashboard using graphql?

src/services/iasql.ts
dfellis: So, kinda?

This sort of metadata that can be derived from the actual database should probably be there instead of here. Two source of truth, y'know. Someone could delete the `aws_account` record and our UI would still show it as ready when it isn't.

Also saying kinda because if/when we have multiple module "trees" the "is ready" condition becomes more complex, and we very likely want to move it out of here entirely and push it to the dashboard (since the database is technically usable on connect).

src/modules/iasql_functions@0.0.1/sql/create_fns.sql
dfellis: I think deriving `is_ready` from a graphql query sounds like a plan to me. :)

test/modules/aws-cloudwatch-integration.ts
```diff
@@ -18,11 +18,15 @@ afterAll(async () => await execComposeDown(modules));
...
-    process.env.AWS_REGION ?? 'barf',
-    process.env.AWS_ACCESS_KEY_ID ?? 'barf',
-    process.env.AWS_SECRET_ACCESS_KEY ?? 'barf',
     'not-needed', 'not-needed').then(...finish(done)));
 
+  it('installs the aws_account module', install(['aws_account']));
+
+  it('inserts aws credentials', query(`
+    INSERT INTO aws_account (region, access_key_id, secret_access_key)
+    VALUES ('${process.env.AWS_REGION}', '${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}')
```
depombo, line 27: minor, but we need to change the `CONTRIBUTING.md` to reflect this change?

dfellis: Yep! It's "simpler" as an explanation, now, too. (Connect has nothing special about it, then you connect to the database and install `aws_account`, insert credentials, and then any other modules you want, but no weird curl syntax anymore!)

src/modules/iasql_functions@0.0.1/sql/create_fns.sql
dfellis: Done :)

src/services/iasql.ts
dfellis: Took care of it as suggested in the paired dashboard change

aguillenv (approved):

depombo (approved):

PR #697, opened 2022-04-26
Mark Route53 as a supported service

dfellis:


depombo (approved):

depombo:
Technically this is not true until we deploy prod?

dfellis:
> Technically this is not true until we deploy prod?

We can wait 'til then, I suppose? But it's a weird grey area. Anyone that checks out the code and runs `yarn docker-compose` is gonna get Route53 support.

PR #698, opened 2022-04-26
Improve iasql_install error messaging when a module is not found

### LINKED ISSUE #688: Better error message when installing module that doesn't exist

dfellis:
Resolves #688


dfellis:
Screenshot from 2022-04-26 18-00-21

Screenshot from 2022-04-26 18-17-29

Selects likely intended modules via [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) and deduplicates the results with [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) for a more intelligent response.

depombo (approved):

PR #699, opened 2022-04-26
identify user in sentry

depombo:


src/router/db.ts
```diff
@@ -17,16 +17,17 @@ db.get('/connect/:dbAlias', async (req, res) => {
       'dbAlias'
     ].filter(k => !req.params.hasOwnProperty(k)).join(', ')}`
   );
+  let uid, email;
```
dfellis, line 20: Why the `email` here when it doesn't appear to be used outside of the `try` block?

src/services/logger.ts
```diff
@@ -72,14 +72,23 @@ export function debugObj(e: any) {
...
     errStack = e.metadata.failures.map((f: Error) => f?.stack ?? f?.message).join('\n');
   }
-  if (config.sentry) err += `\nPlease provide the following error ID if reporting it to the IaSQL team: ${sentry.captureException(errStack)}`;
+  if (config.sentry) {
+    err += `\nPlease provide the following error ID if reporting it to the IaSQL team: ${sentry.captureException(errStack, {
+      // the id or the username need to be present
+      // https://docs.sentry.io/platforms/node/guides/express/enriching-events/identify-user/
+      user: {
+        id: uid,
+        username: dbId
```
dfellis, line 88: The 'username' seems weird?

src/router/db.ts
```diff
@@ -38,16 +39,17 @@ db.post('/connect', async (req, res) => {
       'dbAlias'
     ].filter(k => !req.body.hasOwnProperty(k)).join(', ')}`
   );
+  let uid, email;
```
dfellis, line 42: Same concern here, btw.

src/services/logger.ts
depombo: just going off the cleanest way to bucket events to a user or database. the scheduler doesn't have access to the uid so dubbing the dbId as the username, in amplitude we do something similar with the device id

dfellis (approved):
Now it makes more sense.

src/router/db.ts
aguillenv: because it is being used in the `catch` block with the `logUserErr` fn, I think?

aguillenv (approved):

src/router/db.ts
dfellis: > because it is being used in the `catch` block with the `logUserErr` fn, I think?

It wasn't when I first commented. Luisfer updated the PR.

depombo: Yes, I had forgotten to push some staged changes

dfellis (approved):

PR #703, opened 2022-04-27
use different auth tenant/account for staging

depombo:


src/config/production.ts
```diff
@@ -20,6 +20,9 @@ const config: ConfigInterface = {
     domain: 'https://auth.iasql.com/',
     audience: 'https://api.iasql.com', // id of this api in auth0
   },
+  cors: {
+    origin: 'https://api-staging.iasql.com'
```
dfellis, line 24: This one should just be:

```suggestion
    origin: 'https://api.iasql.com'
```

src/config/local.ts
```diff
@@ -15,6 +15,9 @@ const config: ConfigInterface = {
     debug: true,
     test: false,
   },
+  cors: {
+    origin: 'localhost:3000'
```
dfellis, line 19: This one should have the `http` right?

```suggestion
    origin: 'http://localhost:3000'
```

dfellis (approved):

dfellis (approved):

PR #706, opened 2022-04-28
Reorganize the modules

dfellis:


src/services/gateways/aws.ts
```diff
@@ -154,7 +154,8 @@ import {
   Route53Client
 } from '@aws-sdk/client-route-53'
 import { IAM } from '@aws-sdk/client-iam'
-import { RecordType } from '../../modules/aws_route53_hosted_zones@0.0.1/entity/resource_records_set'
+// WTF? Why is this circular reference here?
+import { RecordType } from '../../modules/0.0.1/aws_route53_hosted_zones/entity/resource_records_set'
```
aguillenv, line 158: Oh probably this was the auto import, RecordType probably should come from an AWS lib.

PR #707, opened 2022-04-28
Implement facsimile of the Alan module system to reduce relative path imports

dfellis:
# `..` dashed!

I got super annoyed by all of the relative path mangling in the prior PR when I moved the modules another level deep in the git tree, especially since [we came up with a superior system for Alan that would've avoided it](https://docs.alan-lang.org/module_resolution.html), so like any sane person I decided to do those tedious changes a second time!

But I figured out a trick with yarn's `link:` syntax to give myself a named alias to the `src/` directory, which I called `iasql`. So now if I want the config I can write `import config from 'iasql/config'` anywhere in the codebase and it'll work, for instance.

This was a really quick change and there's no one here to stop me today! Mwahahaha

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "typescript": "^4.6.3"
   },
   "dependencies": {
+    "iasql": "link:./src",
     "@amplitude/node": "^1.10.0",
     "@aws-sdk/client-cloudwatch-logs": "^3.76.0",
     "@aws-sdk/client-ec2": "^3.67.0",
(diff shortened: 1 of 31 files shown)


dfellis:
Ugh. This blew up because the source code is in Typescript but it is transpiled to Javascript, except for the symlinked code which remains typescript and then doesn't load properly.

Hmm.

dfellis:
With regret, I'm gonna have to close it. Attempts to fix this are super brittle. :(

aguillenv:
😢

PR #721, opened 2022-05-02
Begin development for v0.0.3

dfellis:


src/modules/index.ts
```diff
@@ -1,4 +1,5 @@
 export * as v0_0_1 from './0.0.1'
 export * as v0_0_2 from './0.0.2'
+export * as v0_0_3 from './0.0.3'
 export * as latest from './0.0.2' // Double-exported, but simplifies connect
```
aguillenv, line 4: When this will be updated?

aguillenv: Should we add the TODO if we will do it later?

aguillenv (approved):

src/modules/index.ts
dfellis: Oh, actually let's do that now.

dfellis:
```suggestion
export * as latest from './0.0.3' // Double-exported, but simplifies connect
```

aguillenv (approved):

PR #730, opened 2022-05-03
Remove console.log from where it doesn't belong

dfellis:
I realized that I had accidentally left some debug logs in one of my PRs, so I decided to remove them, but I also noticed a few legacy uses of `console.log` where it doesn't belong and fixed that. (The only places `console.log` should be allowed is in the `logger.ts` file which should be relatively obvious, and in the `scripts` as those files unfortunately cannot have a dependency on the `config`, at least at the moment)

### DIFF
--- a/test/modules/aws-route53-hosted-zones-integration.ts
+++ b/test/modules/aws-route53-hosted-zones-integration.ts
@@ -1,5 +1,6 @@
 import config from '../../src/config';
 import * as iasql from '../../src/services/iasql'
+import logger from '../../src/services/logger'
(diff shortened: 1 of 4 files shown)


test/modules/aws-route53-hosted-zones-integration.ts
```diff
@@ -1,5 +1,6 @@
 import config from '../../src/config';
 import * as iasql from '../../src/services/iasql'
+import logger from '../../src/services/logger'
```
aguillenv, line 3: Oh I think I left this ones by mistake while debugging.

aguillenv (approved):

PR #733, opened 2022-05-03
Fix IAM rate limiting bug

dfellis:


dfellis:
Screenshot from 2022-05-03 17-59-14

Broke it. Now to fix. :)

dfellis:
This is actually a different bug than the one we saw. The AWS API for all roles doesn't return all of them when there's more than 100 roles. We need to modify that function to use a different path.

I have an idea of what could work, but not sure if it really will. (There is always the possibility this is a bug on the AWS API side and we can't do anything except report it to them.)

aguillenv:
this PR is missing the pagination code for more than 100 roles, right?

aguillenv:
this PR is missing the pagination code for more than 100 roles, right?

dfellis:
> this PR is missing the pagination code for more than 100 roles, right?

Yes. I didn't get it implemented on Friday, will tackle soon.

PR #734, opened 2022-05-04
Add LogDNA logging to staging and production

dfellis:
The log ingestion key is not considered a secret, it only allows writing logs, but I can rotate this key and we can make it a secret instead if desired.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "@aws-sdk/client-rds": "^3.76.0",
     "@aws-sdk/client-route-53": "^3.82.0",
     "@aws-sdk/util-waiter": "^3.78.0",
+    "@logdna/logger": "^2.6.4",
     "@sentry/node": "^6.19.6",
     "@types/express": "^4.17.13",
     "@types/express-jwt": "^6.0.2",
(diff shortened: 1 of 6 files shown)


aguillenv:
> The log ingestion key is not considered a secret, it only allows writing logs, but I can rotate this key and we can make it a secret instead if desired.

They usually provide 2 keys, one that is "shareable" just to read/write logs and another one that is secret, not sure how logDna handles it.

aguillenv:
> The log ingestion key is not considered a secret, it only allows writing logs, but I can rotate this key and we can make it a secret instead if desired.

They usually provide 2 keys, one that is "shareable" just to read/write logs and another one that is secret, not sure how logDna handles it.

aguillenv (approved):
As a follow up we should add more loggers around the diffing on apply and sync to have more data, right?

dfellis:
> > The log ingestion key is not considered a secret, it only allows writing logs, but I can rotate this key and we can make it a secret instead if desired.
> 
> They usually provide 2 keys, one that is "shareable" just to read/write logs and another one that is secret, not sure how logDna handles it.

LogDNA actually has three types of keys. This one is write-only, another is read/write, and the third is read/write + API access.

PR #735, opened 2022-05-04
Don't 'Promise' more than you can handle

dfellis:
Despite some issues in reproducing some error conditions we've seen recently, most of them were caused by hitting the AWS API rate limiter, and that should only be possible by making API requests in parallel.

`Promise.all` requires an array of promises to `await` on, but for those promises to all be created they must have been kicked off, and therefore they're operating in parallel with each other. On top of that, this parallelization can cause double-requests (or more) for resources that aren't in our cache yet since the check for the resource happened before any of the requests responded. By switching to `for-await` loops instead, we linearize the code more, which should prevent the rate limiting errors (I hope) but also reduce wasteful duplicate queries by increasing the cache hit success rate.

### DIFF
--- a/src/modules/0.0.3/aws_route53_hosted_zones/index.ts
+++ b/src/modules/0.0.3/aws_route53_hosted_zones/index.ts
@@ -8,7 +8,7 @@ import * as metadata from './module.json'
 export const AwsRoute53HostedZoneModule: Module = new Module({
   ...metadata,
   utils: {
-    hostedZoneMapper: (hz: any, _ctx: Context) => {
+    hostedZoneMapper: (hz: any) => {
       const out = new HostedZone();
       if (!hz?.Id) throw new Error('No HostedZoneId defined');
       out.hostedZoneId = hz.Id;
@@ -47,7 +47,8 @@ export const AwsRoute53HostedZoneModule: Module = new Module({
       cloud: new Crud({
         create: async (hz: HostedZone[], ctx: Context) => {
           const client = await ctx.getAwsClient() as AWS;
-          return await Promise.all(hz.map(async (e) => {
+          const out = [];
+          for (const e of hz) {
             const createdHz = await client.createHostedZone(e.domainName);
             if (!createdHz?.Id) throw new Error('How this happen!?')
             // Re-get the inserted record to get all of the relevant records we care about
@@ -59,8 +60,9 @@ export const AwsRoute53HostedZoneModule: Module = new Module({
             newEntity.id = e.id;
             // Save the record back into the database to get the new fields updated
             await AwsRoute53HostedZoneModule.mappers.hostedZone.db.update(newEntity, ctx);
-            return newEntity;
-          }));
+            out.push(newEntity);
+          }
+          return out;
         },
         read: async (ctx: Context, ids?: string[]) => {
           const client = await ctx.getAwsClient() as AWS;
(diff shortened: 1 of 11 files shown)


src/services/gateways/aws.ts
```diff
@@ -210,21 +210,30 @@ export class AWS {
...
+  }
+
   async updateRoleDescription(name: string, description: string) {
     await this.iamClient.updateRole({
       RoleName: name,
       Description: description
     });
   }
 
-  async updateAttachedPoliciesArns(name: string, attachedPolicyArns: string[]) {
```
aguillenv, line 220: why is this deleted? it is not in use?

src/services/gateways/aws.ts
```diff
@@ -210,21 +210,30 @@ export class AWS {
     return role.Role?.Arn ?? '';
   }
 
+  async newRoleLin(
```
aguillenv, line 213: maybe we should delete the non-lin methods?

src/modules/0.0.3/aws_security_group/index.ts
```diff
@@ -152,8 +155,7 @@ export const AwsSecurityGroupModule: Module = new Module({
                 ctx.memo.db.SecurityGroup[sg.groupId] = sg;
               }
             }
-            return sg;
-          }));
+          }
```
aguillenv, line 158: there's no returning here

src/modules/0.0.3/aws_route53_hosted_zones/index.ts
```diff
@@ -163,8 +177,9 @@ export const AwsRoute53HostedZoneModule: Module = new Module({
             if (!newEntity) return;
```
aguillenv, line 177: Maybe here should be a `continue`?

src/services/gateways/aws.ts
dfellis: That would change the behavior of the prior module versions.

src/modules/0.0.3/aws_route53_hosted_zones/index.ts
dfellis: Oh, yes, you're right. I missed that one.

dfellis:
```suggestion
            if (!newEntity) continue;
```

src/services/gateways/aws.ts
dfellis: Oh, also, I am hoping to follow-up this PR with one that eliminates the AWS gateway file from usage in the `0.0.3` modules, so this would be rolled back in that case.

src/modules/0.0.3/aws_security_group/index.ts
```diff
@@ -180,53 +182,67 @@ export const AwsSecurityGroupModule: Module = new Module({
...
+            // is instead turned into *restoring* the value in the database to match the cloud value
+            if (e.groupName === 'default') {
+              // Because updates are based on the `groupId` matching but not some other property,
+              // we can be sure that the security group rules for the default security group are
+              // properly associated so we don't need to do anything about them here, just restore
+              // the other properties
+              const cloudRecord = ctx?.memo?.cloud?.SecurityGroup?.[e.groupId ?? ''];
+              cloudRecord.id = e.id;
+              await AwsSecurityGroupModule.mappers.securityGroup.db.update(cloudRecord, ctx);
+              ctx.memo.db.SecurityGroup[cloudRecord.groupId] = cloudRecord; // Force the cache
```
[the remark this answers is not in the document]
dfellis: I'm actually surprised that it somehow worked *without* this before.

The `apply` loop grabs basically the entire database and every cloud record it can get its hands on (right now, until we implement the "faster apply" task) and builds a diff of the two to figure out what changes it should make.

Once it is done applying those changes (like this one here), it drops *only the cloud cache* and re-builds it and does another check. Once it is satisfied that there are no differences between the two, it drops both caches and does another check in case there are discrepancies still with the actual database records.

We use that behavior here in the security groups module so you can change the description of a non-default security group without changing the name of the security group because AWS doesn't actually allow that in reality. We go through a fake security group name in one step that's stored only in the in-memory cache of the DB-side of things to get everything depending on that security group to migrate over to this temporary one, and then the cache is dropped and now there's no conflict on simply creating the "real" security group name again with the changed description.

But this path in here where we intentionally *block* changes to the default security group would not pick up on the restoration of the record because the cache still says "do the update" and that's what it was getting stuck on in an infinite loop.

I'm not sure *how* it was working before when we did operations in parallel, because it should have seen the "wrong" value in the cache and gotten into an infinite loop, too. Which just shows a big part of why we needed to get rid of these `Promise.all`s; they run in "parallel" (closer to reality than you'd expect since they're all operating on external APIs) and there's weird data state race conditions that I can't reason about, and this change eliminates it: the log order is the same as the apply order, and couple that with the apply algo we have I can walk through what each diff will do by hand if necessary. :)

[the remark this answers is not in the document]
aguillenv: hmm maybe with the parallel execution the cache was getting correctly updated for another path? definitely unexpected.

aguillenv (approved):

Issue #334, opened 2022-01-15
Create `aws_acm` module


dfellis:

aguillenv:
Closing this issue since PRs #740 and #736 are already merged. Follow up tasks have been created #744 and #745

PR #740, opened 2022-05-09
ACM import

aguillenv:
Part of #334 

- [x] aws_acm_list
- [x] aws_acm_import
- [ ] aws_acm_request


src/modules/0.0.3/aws_acm_import/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      entityId: (e: CertificateImport) => e.id?.toString() ?? '',
+      equals: () => true, // only database values
+      source: 'db',
+      cloud: new Crud({
+        create: async (es: CertificateImport[], ctx: Context) => {
+          const client = await ctx.getAwsClient() as AWS;
+          const textEncoder = new TextEncoder();
+          for (const e of es) {
+            const input: ImportCertificateCommandInput = {
+              Certificate: textEncoder.encode(e.body),
```
dfellis, line 23: Why is the certificate called `body` in the entity?

src/modules/0.0.3/aws_acm_import/index.ts
```diff
@@ -0,0 +1,42 @@
...
+              PrivateKey: textEncoder.encode(e.privateKey),
+            };
+            if (e.chain) {
+              input.CertificateChain = textEncoder.encode(e.chain);
+            }
+            const importedCertArn = await client.importCertificate(input);
+            if (!importedCertArn) throw new Error('Error importing certificate');
```
dfellis, line 30: Perhaps import failure should flag the record with a text column on the row? (And if so, then any record with that column not null should be ignored by this logic.)

dfellis:
Left a couple of comments. Unfortunately I think this one is gonna be held back for the next release.

src/modules/0.0.3/aws_acm_import/index.ts
aguillenv: oh I was just replicating the names they have in the interface. I can just change `body` to `certificate`.
image

src/modules/0.0.3/aws_acm_import/index.ts
aguillenv: Hmm I see your point. Could be interesting. Do you think it could mislead users thinking that since it does not errored everything worked fine?

dfellis: Well, blocking the `apply` with an error is probably better. I guess it'd be good if the error message had more information about what went wrong?

aguillenv: well that would be happen if `apply` fails. This specific line is just one of those `What!?` or `this should never happen` errors... I'm just making the check for sanity and to let typescript know that `importedCertArn` will be defined, but if import fails it will simply `throw`.

dfellis: > well that would be happen if `apply` fails. This specific line is just one of those `What!?` or `this should never happen` errors... I'm just making the check for sanity and to let typescript know that `importedCertArn` will be defined, but if import fails it will simply `throw`.

Well, so, if the cert definition provided by the user is bad, I would expect `importCertificate` to fail? And hopefully with a useful error message from AWS that the certificate is invalid in some way. I'd rather it be wrapped in a `try-catch` and then re-throw the error with any extra explanatory text we want to prepend to their message.

aguillenv: Yes `importCertificate` will fail and `throw`. The current errors from AWS are ok, let me share them and see if we think is necessary to add more detail, but I don't think so.

image

image

image

dfellis: That's pretty good.

There's a more generic change I want to make to annotate the `APPLY` errors with the table and ID value that triggered said error, but that's outside of the scope of this PR.

Approved!

dfellis (approved):

PR #741, opened 2022-05-09
Upgrade to v0.0.3 support

dfellis:


aguillenv (approved):
🚀

src/services/iasql.ts
```diff
@@ -1062,7 +1062,71 @@ export async function upgrade(dbId: string, dbUser: string) {
         }
       })();
       throw new Error('Upgrading. Please disconnect and reconnect to the database');
-    case 'v0_0_2': // TODO: Implement this upgrade before release of v0.0.3
+    case 'v0_0_2':
+      // The upgrade path here *should* work for all versions going forward. 0.0.1 was the exception
```
depombo, line 1066: would every version have a new path going forwards?

depombo (approved):

src/services/iasql.ts
dfellis: So hopefully we only need to minimally edit this one and attach a `case 'v0_0_3':` line at the top.

Once we drop `v0.0.1` from the repo, and if that pattern does hold, then we could simplify this quite a bit into a simple question of "is the version the latest version or not?" piece and drop the switch-case logic entirely.

PR #748, opened 2022-05-09
Easier install logic (auto-install dependencies and don't error if there's nothing to do)

dfellis:


dfellis:
## Step 1: Start with a blank-ish DB

Screenshot from 2022-05-09 13-53-00

## Step 2: Install fargate, get a bunch of other modules for free!

Screenshot from 2022-05-09 13-53-20

This PR also solves the "debounce" issue for install, but not uninstall, by accident, since trying to install something twice does not error anymore. But we should really figure out why the click is triggering twice in the dashboard instead of relying on this (since it still errors in the uninstall path).

aguillenv (approved):

depombo (approved):

PR #756, opened 2022-05-12
Add an audit log

dfellis:
- Add the new 'iasql_audit_log' table to the platform and fix the module migration generation to require zero config changes to run
- Write the audit trigger function and attach/detach it on install/uninstall
- Syntax fixes for the trigger function

### DIFF
--- a/src/modules/0.0.4/iasql_functions/sql/create_fns.sql
+++ b/src/modules/0.0.4/iasql_functions/sql/create_fns.sql
@@ -1,17 +1,18 @@
 -- TODO: Does this belong here or in a similar file in the iasql_platform module?
-create or replace function iasql_audit() returns trigger as $$
+create or replace function iasql_audit() returns trigger
 language plpgsql security definer
+as $$
 begin
   if (TG_OP = 'INSERT') then
     INSERT INTO iasql_audit_log (ts, "user", table_name, change_type, change)
-    VALUES (now(), user, TG_TABLE_NAME, 'INSERT', '{"change":' || to_json(NEW.*) || '}');
-  elif (TG_OP = 'DELETE') then
+    VALUES (now(), user, TG_TABLE_NAME, 'INSERT', ('{"change":' || to_json(NEW.*) || '}')::json);
+  elsif (TG_OP = 'DELETE') then
     INSERT INTO iasql_audit_log (ts, "user", table_name, change_type, change)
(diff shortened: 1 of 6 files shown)


dfellis:
Screenshot from 2022-05-12 18-03-16

Looks like this. Tracks all DB changes, when they happened, and the user that did it (the engine's user would indicate `sync` and `install` changes, while the "normal" user would be changes to be pushed at the cloud.

src/modules/0.0.4/iasql_platform/entity/index.ts
```diff
@@ -45,4 +46,42 @@ export class IasqlTables {
...
+
+@Entity()
+export class IasqlAuditLog {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column({
+    type: 'timestamp without time zone',
+  })
+  ts: Date;
```
aguillenv, line 65: should this be with timezone to save it in UTC?

src/services/iasql.ts
```diff
@@ -870,6 +870,16 @@ ${Object.keys(tableCollisions)
         return mt;
       }) ?? [];
       await orm.save(Modules.IasqlPlatform.utils.IasqlTables, modTables);
+      // For each table, we need to attach the audit log trigger, if the platform is >=0.0.4
+      if (!['v0_0_1', 'v0_0_2', 'v0_0_3'].includes(versionString)) {
+        for (const table of md.provides.tables) {
+          await queryRunner.query(`
+            CREATE TRIGGER ${table}_audit
+            AFTER INSERT OR UPDATE OR DELETE ON ${table}
+            FOR EACH ROW EXECUTE FUNCTION iasql_audit();
```
aguillenv, line 879: should this be the deferred one?

src/modules/0.0.4/iasql_functions/sql/create_fns.sql
```diff
@@ -1,3 +1,22 @@
+-- TODO: Does this belong here or in a similar file in the iasql_platform module?
+create or replace function iasql_audit() returns trigger
+language plpgsql security definer
+as $$
+begin
+  if (TG_OP = 'INSERT') then
```
aguillenv, line 6: does the `TG_OP` is a Postgres standard to define the trigger operation or where does it come from?

src/services/iasql.ts
aguillenv: Oh that quote answer my concern. Perfect!

depombo (approved):

PR #768, opened 2022-05-17
Move the graphile workers into a child process

dfellis:
- Convert the graphile worker scheduler into a child process
- We no longer have a single singleton for the metadata database access

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,7 +4,7 @@ import cors from 'cors';
 import express from 'express';
 import { inspect } from 'util';
 
-import * as scheduler from './services/scheduler'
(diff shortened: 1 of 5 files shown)


depombo:
hmm I am a bit concerned about whether this change affects error handling or not? perhaps we can test that somehow?

dfellis:
> hmm I am a bit concerned about whether this change affects error handling or not? perhaps we can test that somehow?

You're right that what will happen right now is that the error causes the caller to hang forever instead of erroring. I will have to improve that.

However, I am unsure how to test any failure here?

depombo (approved):

aguillenv (approved):

PR #771, opened 2022-05-18
Install *all* missing dependencies, not just the first level in the tree

dfellis:


dfellis:
> can we add a unit test for this to avoid regressions?

Sure. The reproduction step is super-simple.

depombo (approved):

PR #775, opened 2022-05-20
Implement simple s3 bucket support

### LINKED ISSUE #639: Create an `aws_s3` module

dfellis:
Resolves #639 but this is only the most basic of support. More features to be added over time


src/modules/0.0.5/aws_s3/index.ts
```diff
@@ -0,0 +1,53 @@
...
+        // TODO: With the model this simple it is actually impossible to really update this thing
+        // as the user can't change the create timestamp themselves, so just restore the record from
+        // the cloud cache and force it into the DB cache
+        update: async (es: Bucket[], ctx: Context) => {
+          for (const e of es) {
+            const cloudRecord = ctx?.memo?.cloud?.Bucket?.[e.name ?? ''];
+            await AwsS3Module.mappers.bucket.db.update(cloudRecord, ctx);
+            ctx.memo.db.Bucket[cloudRecord.name] = cloudRecord;
+          }
+        },
```
aguillenv, line 43: Shouldn't we return here the updated values?

aguillenv (approved):

src/modules/0.0.5/aws_s3/index.ts
dfellis: Hmm. Yeah. This probably only works coincidentally because we blow away the cloud cache on each loop in the `apply` and `sync` is only being tested in one path in the test suite.

src/modules/0.0.5/aws_s3/entity/index.ts
```diff
@@ -0,0 +1,24 @@
...
+  Entity,
+  PrimaryColumn,
+} from 'typeorm'
+
+import { cloudId, } from '../../../../services/cloud-id'
+
+@Entity()
+export class Bucket {
+  @PrimaryColumn({
+    nullable: false,
```
depombo, line 12: I think primary columns are nullable and unique by definition so this is a no-op

depombo (approved):
minor comment, looks good

PR #776, opened 2022-05-21
support tags in ec2 instance

depombo:


src/modules/0.0.5/aws_ec2/entity/instance.ts
```diff
@@ -13,7 +13,6 @@ import {
 import { SecurityGroup, } from '../../aws_security_group/entity';
 import { cloudId, } from '../../../../services/cloud-id'
 
-// TODO complete instance schema
```
depombo, line 16: I have a PR on top of this one where I add more fields, but somehow this change got added on this branch

src/modules/0.0.5/aws_ec2/index.ts
```diff
@@ -41,11 +44,12 @@ export const AwsEc2Module: Module = new Module({
           const client = await ctx.getAwsClient() as AWS;
           for (const instance of es) {
             if (instance.ami) {
-              const instanceId = await client.newInstance(
-                instance.name,
+              const instanceId = await client.newInstanceV2(
                 instance.instanceType,
                 instance.ami,
                 instance.securityGroups.map(sg => sg.groupId).filter(id => !!id) as string[],
+                instance.keyPairName,
```
depombo, line 51: this wasn't passed in before which makes me think this wasn't working when creating it via IaSQL?

src/services/gateways/aws.ts
```diff
@@ -184,7 +186,7 @@ type AWSConfig = {
 export const IASQL_EC2_TAG_NAME = 'IaSQL_Name';
 
 export class AWS {
-  private ec2client: EC2Client
+  private ec2client: EC2
   private ecrClient: ECRClient
```
depombo, line 190: when we move the AWS client usage to the mappers we should use the `EC2`/`ECR` objects instead of the `ECRClient`/`EC2Client`. They are more DRY / concise

dfellis (approved):
I still think a Tag table would work, but this is fine.

src/services/gateways/aws.ts
aguillenv: oh really?

src/modules/0.0.5/aws_ec2/entity/instance.ts
```diff
@@ -41,6 +36,12 @@ export class Instance {
   })
   keyPairName: string;
 
+  @Column({
+    type: 'json',
+    nullable: true,
+  })
+  tags: { [key: string]: string }
```
aguillenv, line 43: Since it is nullable it'd probably be:
```suggestion
  tags?: { [key: string]: string }
```

src/modules/0.0.5/aws_ec2/index.ts
```diff
@@ -28,11 +30,12 @@ export const AwsEc2Module: Module = new Module({
...
     instance: new Mapper<Instance>({
       entity: Instance,
-      equals: (a: Instance, b: Instance) => Object.is(a.name, b.name) &&
-        Object.is(a.instanceId, b.instanceId) &&
+      equals: (a: Instance, b: Instance) => Object.is(a.instanceId, b.instanceId) &&
         Object.is(a.ami, b.ami) &&
         Object.is(a.instanceType, b.instanceType) &&
         Object.is(a.keyPairName, b.keyPairName) &&
+        Object.is(Object.keys(a.tags)?.length, Object.keys(b.tags)?.length) &&
+        Object.keys(a.tags)?.every(ak => a.tags[ak] === b.tags[ak]) &&
```
aguillenv, line 38: Since tags could be null I thing this should be:
```suggestion
        Object.keys(a.tags ?? {}).every(ak => a.tags[ak] === b.tags[ak]) &&
```

src/services/gateways/aws.ts
```diff
@@ -184,7 +186,7 @@ type AWSConfig = {
 export const IASQL_EC2_TAG_NAME = 'IaSQL_Name';
 
 export class AWS {
-  private ec2client: EC2Client
+  private ec2client: EC2
```
aguillenv, line 189: Changing the client here would not affect the current ec2 methods that are expecting the old client? I mean shouldn't this be:
```suggestion
  private ec2client: EC2Client
  private newEc2Client: EC2
```

aguillenv (approved):

PR #787, opened 2022-05-24
add editable fields to ec2 instance

depombo:


src/modules/0.0.5/aws_ec2/index.ts
```diff
@@ -78,14 +89,34 @@ export const AwsEc2Module: Module = new Module({
           }
           return out;
         },
-        updateOrReplace: (_a: Instance, _b: Instance) => 'replace',
+        updateOrReplace: (prev: Instance, next: Instance) => {
+          if (!AwsEc2Module.utils.instanceEqReplaceableFields(prev, next)) return 'replace';
```
dfellis, line 93: Clever reuse here. I was wondering why you split equals in two.

src/modules/0.0.5/aws_ec2/index.ts
```diff
@@ -78,14 +89,34 @@ export const AwsEc2Module: Module = new Module({
...
+                await client.updateTags(e.instanceId as string, e.tags)
+              }
+              if (!Object.is(e.state, cloudRecord.state) && e.instanceId) {
+                if (e.state === 'running') {
+                  await client.startInstance(e.instanceId as string);
+                } else if (e.state === 'hibernated') {
+                  await client.stopInstance(e.instanceId as string, true);
+                } else if (e.state === 'stopped') {
+                  await client.stopInstance(e.instanceId as string);
+                }
```
dfellis, line 113: Should there be an `error` thrown in the `else` path if reached?

src/modules/0.0.5/aws_ec2/index.ts
```diff
@@ -78,14 +89,34 @@ export const AwsEc2Module: Module = new Module({
...
-            const created = await AwsEc2Module.mappers.instance.cloud.create([e], ctx);
-            await AwsEc2Module.mappers.instance.cloud.delete([cloudRecord], ctx);
-            out.push(created);
+            const isUpdate = AwsEc2Module.mappers.instance.cloud.updateOrReplace(cloudRecord, e) === 'update';
+            if (isUpdate) {
+              if (!AwsEc2Module.utils.instanceEqTags(e, cloudRecord) && e.instanceId && e.tags) {
+                await client.updateTags(e.instanceId as string, e.tags)
+              }
+              if (!Object.is(e.state, cloudRecord.state) && e.instanceId) {
+                if (e.state === 'running') {
```
dfellis, line 107: Also, should we use `State.RUNNING` here and etc for the other checks?

dfellis (approved):
Minor comments only

dfellis (changes requested):
Actually, unfortunately the test suite is failing consistently on this PR in the relevant tests, so this is not quite there, yet.

dfellis (approved):
Once the test suite passes go for it.

src/modules/0.0.5/aws_ec2/index.ts
```diff
@@ -78,14 +89,37 @@ export const AwsEc2Module: Module = new Module({
...
+              }
+              if (!Object.is(e.state, cloudRecord.state) && e.instanceId) {
+                if (cloudRecord.state === State.STOPPED && e.state === State.RUNNING) {
+                  await client.startInstance(insId);
+                } else if (cloudRecord.state === State.RUNNING && e.state === State.HIBERNATED) {
+                  await client.stopInstance(insId, true);
+                } else if (cloudRecord.state === State.RUNNING && e.state === State.STOPPED) {
+                  await client.stopInstance(insId);
+                } else {
+                  throw new Error(`Invalid instance state transition. From CLOUD state ${cloudRecord.state} to DB state ${e.state}`);
```
aguillenv, line 111: Not sure if I'd throw here, maybe just "restore" the cloud value and do nothing?

src/services/gateways/aws.ts
```diff
@@ -357,7 +357,97 @@ export class AWS {
...
+      },
+      input,
+      async (client, cmd) => {
+        try {
+          const data = await client.send(cmd);
+          for (const reservation of data?.Reservations ?? []) {
+            for (const instance of reservation?.Instances ?? []) {
+              if (instance.State?.Name !== 'stopped')
+                return { state: WaiterState.RETRY };
+            }
```
aguillenv, line 438: Shouldn't we be checking here if stopped or hibernated?

src/modules/0.0.5/aws_ec2/index.ts
```diff
@@ -27,18 +33,22 @@ export const AwsEc2Module: Module = new Module({
...
-        Object.is(Object.keys(a.tags ?? {})?.length, Object.keys(b.tags ?? {})?.length) &&
-        Object.keys(a.tags ?? {})?.every(ak => (a.tags ?? {})[ak] === (b.tags ?? {})[ak]) &&
-        Object.is(a.securityGroups?.length, b.securityGroups?.length) &&
-        a.securityGroups?.every(as => !!b.securityGroups?.find(bs => Object.is(as.groupId, bs.groupId))),
+      equals: (a: Instance, b: Instance) => Object.is(a.state, b.state) &&
+        AwsEc2Module.utils.instanceEqReplaceableFields(a, b) &&
+        AwsEc2Module.utils.instanceEqTags(a, b),
       source: 'db',
       cloud: new Crud({
         create: async (es: Instance[], ctx: Context) => {
```
aguillenv, line 54: Oh I think you have to return the created values on create if I'm not wrong

src/modules/0.0.5/aws_ec2/index.ts
depombo: this *should* never happen so would prefer to throw an error and if it comes up to often then address?

PR #788, opened 2022-05-24
release 0.0.5

depombo:


dfellis:
I would put the 0.0.6 stuff in a separate PR. It also needs a new entry in `src/modules/index.ts` for those to work, and is likely why the test suite is failing.

PR #792, opened 2022-05-25
Make ECS tests more stable and reduce execution times

### LINKED ISSUE #677: Fix ECS flakiness
https://github.com/iasql/iasql-engine/runs/6077943118?check_suite_focus=true

https://github.com/iasql/iasql-engine/runs/6079270998?check_suite_focus=true

aguillenv:
Resolves #677

### DIFF
--- a/src/services/gateways/aws.ts
+++ b/src/services/gateways/aws.ts
@@ -541,20 +541,31 @@ export class AWS {
         new DeleteSecurityGroupCommand(instanceParams),
       );
     } catch(e: any) {
-      // If it is a dependency violation we add the dependency to the error message in order to debug what is happening
       if (e.Code === 'DependencyViolation') {
-        const sgEniInfo = await this.ec2client.send(
-          new DescribeNetworkInterfacesCommand({
-            Filters: [
-              {
-                Name: 'group-id',
-                Values: [`${instanceParams.GroupId}`]
-              }
-            ]
-          })
-        );
-        const eniMessage = `Network interfaces associated with security group ${instanceParams.GroupId}: ${JSON.stringify(sgEniInfo.NetworkInterfaces)}`;
-        e.message = `${e.message} | ${eniMessage}`;
+        // Just await for 5 min on every dependency violation and retry
+        await new Promise(resolve => setTimeout(resolve, 5 * 60 * 1000));
+        try {
+          return await this.ec2client.send(
+            new DeleteSecurityGroupCommand(instanceParams),
+          );
+        } catch (e2: any) {
+          // If the dependency continues we add the dependency to the error message in order to debug what is happening
+          if (e2.Code === 'DependencyViolation') {
+            const sgEniInfo = await this.ec2client.send(
+              new DescribeNetworkInterfacesCommand({
+                Filters: [
+                  {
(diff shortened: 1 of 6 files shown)


src/services/gateways/aws.ts
```diff
@@ -541,20 +541,31 @@ export class AWS {
...
-              {
-                Name: 'group-id',
-                Values: [`${instanceParams.GroupId}`]
-              }
-            ]
-          })
-        );
-        const eniMessage = `Network interfaces associated with security group ${instanceParams.GroupId}: ${JSON.stringify(sgEniInfo.NetworkInterfaces)}`;
-        e.message = `${e.message} | ${eniMessage}`;
+        // Just await for 5 min on every dependency violation and retry
```
aguillenv, line 545:
```suggestion
        // Just wait for 5 min on every dependency violation and retry
```

test/helpers.ts
```diff
@@ -21,7 +21,8 @@ export async function execComposeDown(modules?: string[], region?: string) {
       logger.error(`Error cleaning db with error: ${e.message}`, e);
     }
   }
-  execSync('cd test && docker-compose down');
+  // Compose down and ensure everything gets tear down to avoid infinite test runs
+  execSync('cd test && docker-compose down --remove-orphans');
```
dfellis, line 25: I'm not convinced this will do that, since the issue (that I have seen) is a graphile worker still running and trying to talk to postgres. The test suite doesn't run inside of a docker container so that particular infinite loop would persist.

test/modules/aws-ecs-ecr-integration.ts
```diff
@@ -343,12 +351,6 @@ describe('ECS Integration Testing', () => {
     WHERE role_name = '${taskExecRoleName}';
   `, (res: any[]) => expect(res.length).toBe(0)));
 
-  it('check log group deletion', query(`
-    SELECT *
-    FROM log_group
-    WHERE log_group_name = '${logGroupName}';
-  `, (res: any[]) => expect(res.length).toBe(0)));
-
```
aguillenv, line 351: Removing this since log group is not in use in this test

aguillenv:
The flakiness I've seen in ECS are:
- Error creating a task definition due to not having a role available. This issue was discussed offline. After taking a deep look at the logs the only I can understand is happening there is a consistency issue with AWS APIs, I cannot see other reason. The weird is that a previous call seem to have the correct data and the next one does not!?? maybe they have different servers and we hit one where the data was not updated yet? Not sure. This is the [link to the error logs](https://github.com/iasql/iasql-engine/runs/6594732237?check_suite_focus=true)
- Database FK error between containers and public ECR repositories. I'm pretty sure this weirdness was caused by [this error in the code](https://github.com/iasql/iasql-engine/pull/792/files#diff-5a2a751b9ed3a84cb8da82204aa4251759312b0cfabcdbf37ab4e4d478ef92e2R14). It did not fail more often because the properties between `Repository` entity and `PublicRepository` are quite similar. This is the [link to the error logs](https://github.com/iasql/iasql-engine/runs/6594732379?check_suite_focus=true)

dfellis (approved):

depombo (approved):

PR #795, opened 2022-05-25
aws ec2 metadata module

depombo:


src/modules/0.0.6/aws_ec2_metadata/entity/instance_metadata.ts
```diff
@@ -0,0 +1,66 @@
...
+    onDelete: 'CASCADE',
+  })
+  @JoinColumn({
+    name: 'id'
+  })
+  id: number;
+
+  @Column()
+  @cloudId
+  instanceId: string;
```
aguillenv, line 34: hmmm but why aren't you joining by the `instanceId`? It is only possible to have values on this table if an instance already exists.

depombo: discussed offline. `instanceId` can be nullable in the `instance` table which violates the fk constraint

test/modules/aws-security-group-integration.ts
```diff
@@ -304,7 +304,7 @@ describe('Security Group install/uninstall', () => {
     true).then(...finish(done)));
 
   it('uninstalls the Security Group module', uninstall(
-    ['aws_rds', 'aws_ecs_fargate', 'aws_elb', 'aws_security_group', 'aws_ec2'],
+    ['aws_rds', 'aws_ecs_fargate', 'aws_elb', 'aws_security_group', 'aws_ec2', 'aws_ec2_metadata'],
```
dfellis, line 307: Is this change necessary?

test/modules/aws-vpc-integration.ts
```diff
@@ -131,7 +131,7 @@ describe('VPC install/uninstall', () => {
     true).then(...finish(done)));
 
   it('uninstalls the VPC module', uninstall(
-    ['aws_vpc', 'aws_ecs_fargate', 'aws_security_group', 'aws_rds', 'aws_elb', 'aws_ec2']));
+    ['aws_vpc', 'aws_ecs_fargate', 'aws_security_group', 'aws_rds', 'aws_elb', 'aws_ec2', 'aws_ec2_metadata']));
```
dfellis, line 134: Same here. This shouldn't be needed

dfellis (approved):
Two small comments. Would like them reverted as the EC2 module should continue to work without the metadata module

aguillenv (approved):

PR #798, opened 2022-05-27
Fix looping when no security group defined on EC2

aguillenv:
Right now if you create an instance and do not defined security groups it enters into a loop trying to stabilize. The `default` security group is added by AWS if no security group was defined. We need to update the database properly based on the newly created cloud object to avoid this looping.

### DIFF
--- a/test/modules/aws-ec2-integration.ts
+++ b/test/modules/aws-ec2-integration.ts
@@ -165,6 +165,37 @@ describe('EC2 Integration Testing', () => {
     FROM instance;
   `, (res: any[]) => expect(res.length).toBe(2)));
 
+  it('adds an ec2 instance with no security group', (done) => {
+    query(`
+      INSERT INTO instance (ami, instance_type, tags)
+      VALUES ('${amznAmiId}', 't2.micro', '{"name":"${prefix}-nosg"}');
+    `)((e?: any) => {
+      if (!!e) return done(e);
+      done();
+    });
+  });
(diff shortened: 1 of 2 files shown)


dfellis (approved):
Can we add a test for this? Create an EC2 instance and not define the security group and show that this works?

test/modules/aws-ec2-integration.ts
```diff
@@ -165,6 +165,37 @@ describe('EC2 Integration Testing', () => {
...
+  `, (res: any[]) => expect(res.length).toBe(1)));
+
+  it('check number of security groups for instance', query(`
+    SELECT *
+    FROM instance_security_groups
+    INNER JOIN instance ON instance.id = instance_security_groups.instance_id
+    WHERE tags ->> 'name' = '${prefix}-nosg';
+  `, (res: any[]) => expect(res.length).toBe(1)));
+
   it('deletes both ec2 instances', query(`
```
dfellis, line 199:
```suggestion
  it('deletes all ec2 instances', query(`
```

dfellis: However, this is also dangerous because it's not respecting the tag name prefix, so simultaneous runs on the same region will break unexpectedly.

aguillenv: indeed, I'll update it

dfellis (approved):

PR #800, opened 2022-05-27
EC2 + ELB

### LINKED ISSUE #783: EC2 to depend on ELB

aguillenv:
Resolves  #783


src/modules/0.0.6/aws_ec2/index.ts
```diff
@@ -121,5 +134,60 @@ export const AwsEc2Module: Module2 = new Module2({
...
+            return [];
+          }
+          const registredInstances = await client.getRegisteredInstances() ?? [];
+          const out = [];
+          for (const i of registredInstances) {
+            out.push(await AwsEc2Module.utils.registredInstanceMapper(i, ctx));
+          }
+          return out;
+        },
+        updateOrReplace: () => 'update',
```
dfellis, line 169: Technically for a trivial one like this you can omit this method.

src/services/iasql.ts
```diff
@@ -796,12 +796,12 @@ export async function install(moduleList: string[], dbId: string, dbUser: string
   // Check to make sure that all dependent modules are in the list
   let missingDeps: string[] = [];
   do {
-    missingDeps = mods
+    missingDeps = [...new Set(mods
```
aguillenv, line 799: This is needed because now ec2 installs elb as a dependency if not defined, and both share vpc as an inner dependency and it creates a duplicate in the mods array.

aguillenv:
I've decided to represent in this case the relationship between ELB `TargetGroup` and EC2 `Instance` as a junction table since a `TargetGroup` can have multiple `Instance`s registered and an `Instance` can be part of multiple `TargetGroup`s. I've added a constraint for correctness that checks that the target group related with the instance has the valid type.

Other options considered:
- Extract the `targetType` from the `TargetGroup` table and have a big junction table. This table would need to have `FK`s to all possible dependants (instance and ecs services for now but there are 2 more types). The reasons to discard this option are:
  - It would force to just create a target group only if a dependent object is in place, otherwise we cannot know the type
  - Since the type is mandatory for the creation, the `TargetGroup` table is the one that should have this field
  - Too complex table and with 3 `NULL`s per row once all types are in place.
- N junction tables and derivate the target group type property from them. I discarded this option basically for the same reasons as above:
  - It would force to just create a target group only if a dependent object is in place, otherwise we cannot know the type
  - Since the type is mandatory for the creation, the `TargetGroup` table is the one that should have this field
- Integrate the `TargetGroup` FK to the instance table (following the same pattern as ecs service + target group). Reasons:
  - Is not fully correct, it is a Many-to-Many relationship, not a One-to-Many
  - In this case we need to add a property that belongs to the relationship (the `port`) and could create confusion in the instance table.

I can think another option to implement in the future is to split the target group table to one table per target group type and then have one junction table with each target group table by type.

src/modules/0.0.6/aws_ec2/index.ts
```diff
@@ -125,5 +135,65 @@ export const AwsEc2Module: Module2 = new Module2({
...
+              Object.is(ri.targetGroup.targetGroupName, e.targetGroup.targetGroupName));
+            await AwsEc2Module.mappers.registeredInstance.db.update(relevantRegister, ctx);
+            out.push(relevantRegister);
+          }
+          return out;
+        },
+        read: async (ctx: Context, id?: string) => {
+          const client = await ctx.getAwsClient() as AWS;
+          if (id) {
+            // Just return empty array. Since theres no really a cloud id this should never be called.
```
dfellis, line 161: The faster apply change we want to do would fail on this logic. Why not just put the ID check at the end and filter to the relevant record instead? Same LOC and the function does what it should.

src/modules/0.0.6/aws_ec2/migration/1653932287040-aws_ec2.ts
```diff
@@ -1,24 +1,47 @@
...
         await queryRunner.query(`CREATE INDEX "IDX_fa3c179d5090cb1309c63b5e20" ON "instance_security_groups" ("instance_id") `);
         await queryRunner.query(`CREATE INDEX "IDX_b3b92934eff56d2eb0477a1d27" ON "instance_security_groups" ("security_group_id") `);
+        await queryRunner.query(`ALTER TABLE "registered_instance" ADD CONSTRAINT "FK_407298b1d222bfd78d4886a254d" FOREIGN KEY ("instance") REFERENCES "instance"("id") ON DELETE CASCADE ON UPDATE NO ACTION`);
+        await queryRunner.query(`ALTER TABLE "registered_instance" ADD CONSTRAINT "FK_bdcacf2c01109dd6ad3ffab7a83" FOREIGN KEY ("target_group") REFERENCES "target_group"("target_group_name") ON DELETE CASCADE ON UPDATE NO ACTION`);
         await queryRunner.query(`ALTER TABLE "instance_security_groups" ADD CONSTRAINT "FK_fa3c179d5090cb1309c63b5e20a" FOREIGN KEY ("instance_id") REFERENCES "instance"("id") ON DELETE CASCADE ON UPDATE CASCADE`);
         await queryRunner.query(`ALTER TABLE "instance_security_groups" ADD CONSTRAINT "FK_b3b92934eff56d2eb0477a1d27f" FOREIGN KEY ("security_group_id") REFERENCES "security_group"("id") ON DELETE CASCADE ON UPDATE CASCADE`);
+        await queryRunner.query(`
```
dfellis, line 17: Perhaps this should be put in a separate `.sql` file and included in here like how we do it for `iasql_functions` so regenerating the migration is a bit less painful, and VSCode/vim/etc can properly syntax highlight and code complete it?

dfellis (changes requested):
I unfortunately need to block on the cloud read suggestion since it blocks future work.

src/modules/0.0.6/aws_ec2/index.ts
aguillenv: Hmm because there's no really a unique I'd. It is a composite key in the database and the entityId custom function is a concatenation of id and target_group_name

depombo (approved):

PR #807, opened 2022-05-31
Ignore broken cloud records

### LINKED ISSUE #743: Cloud read ignore misconfigured resources instead of failing import/sync

dfellis:
Resolves #743

Places where the values provided by the AWS gateway are not present instead trigger the mappers to return undefined. Places where the Database does not yet have the value are still thrown errors to trigger a retry. This is still WIP as it needs to be thoroughly tested, but this is the approach I have waiting for review / double-checking

### DIFF
--- a/src/modules/0.0.9/aws_acm_list/index.ts
+++ b/src/modules/0.0.9/aws_acm_list/index.ts
@@ -1,21 +1,45 @@
+import { CertificateDetail, } from '@aws-sdk/client-acm'
+
 import { AWS, } from '../../../services/gateways/aws'
 import { Context, Crud2, Mapper2, Module2, } from '../../interfaces'
-import { Certificate } from './entity'
+import {
+  Certificate,
+  certificateTypeEnum,
+  certificateRenewalEligibilityEnum,
+  certificateStatusEnum,
+} from './entity'
 import * as metadata from './module.json'
+
+// Get Typescript to understand enum matching
+const isCertType = (t?: string): t is certificateTypeEnum => {
+  if (!t) return false;
+  return Object.values<string>(certificateTypeEnum).includes(t);
+}
+const isRenewalEligibility = (t?: string): t is certificateRenewalEligibilityEnum => {
+  if (!t) return false;
+  return Object.values<string>(certificateRenewalEligibilityEnum).includes(t);
+}
+const isStatusType = (t?: string): t is certificateStatusEnum => {
+  if (!t) return false;
+  return Object.values<string>(certificateStatusEnum).includes(t);
+}
 
 export const AwsAcmListModule: Module2 = new Module2({
   ...metadata,
   utils: {
-    certificateMapper: (e: any) => {
+    certificateMapper: (e: CertificateDetail) => {
       const out = new Certificate();
-      if (!e?.CertificateArn) throw new Error('No CertificateArn defined');
+      // To ignore faulty data in AWS, instead of throwing an error on bad data, we return
+      // undefined
+      if (!e?.CertificateArn) return undefined;
       out.arn = e.CertificateArn;
       out.certificateId = e.CertificateArn.split('/').pop();
-      out.certificateType = e.Type;
+      if (isCertType(e.Type)) out.certificateType = e.Type;
+      if (!e.DomainName) return undefined;
       out.domainName = e.DomainName;
       out.inUse = !!e.InUseBy?.length;
-      out.renewalEligibility = e.RenewalEligibility;
-      out.status = e.Status;
+      if (isRenewalEligibility(e.RenewalEligibility)) out.renewalEligibility = e.RenewalEligibility;
+      if (isStatusType(e.Status)) out.status = e.Status;
       return out;
     },
   },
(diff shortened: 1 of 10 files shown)


src/modules/0.0.7/aws_acm_list/index.ts
```diff
@@ -1,21 +1,46 @@
...
+    certificateMapper: (e: CertificateDetail) => {
       const out = new Certificate();
-      if (!e?.CertificateArn) throw new Error('No CertificateArn defined');
+      // To ignore faulty data in AWS, instead of throwing an error on bad data, we return undefined
+      // if (!e?.CertificateArn) throw new Error('No CertificateArn defined');
+      if (!e.CertificateArn) return undefined;
       out.arn = e.CertificateArn;
       out.certificateId = e.CertificateArn.split('/').pop();
-      out.certificateType = e.Type;
+      if (isCertType(e.Type)) out.certificateType = e.Type;
```
aguillenv, line 37: I guess this is needed instead of just `out.certificateType = e.Type as certificateTypeEnum;` because `e.Type` could be `null`/`undefined` or there is another reason?

src/modules/0.0.7/aws_ec2_metadata/index.ts
```diff
@@ -12,20 +12,25 @@ export const AwsEc2MetadataModule: Module2 = new Module2({
...
       out.spot = instance.InstanceLifecycle === 'spot';
-      out.cpuCores = instance.CpuOptions?.CoreCount as number;
-      const instanceType = await client.getInstanceType(instance.InstanceType ?? '');
-      out.memSizeMB = instanceType?.MemoryInfo?.SizeInMiB as number;
+      if (!instance.CpuOptions?.CoreCount) return undefined;
+      out.cpuCores = instance.CpuOptions?.CoreCount;
+      if (!instance.InstanceType) return undefined;
+      const instanceType = await client.getInstanceType(instance.InstanceType);
+      if (!instanceType?.MemoryInfo?.SizeInMiB) return undefined;
+      out.memSizeMB = instanceType?.MemoryInfo?.SizeInMiB;
```
aguillenv, line 33: nit: probably would make sense to have a "precondition" check at the beginning of the mapper checking if any of the `required` properties is wrong just `return undefined;` once.

aguillenv: Also, I agree replacing `throw`s and return `undefined`s instead, but I'm not entirely sure if in this case it should be the approach, I guess if we are 100% sure these properties always have to have a valid value it makes sense but if any of them could not be there we could default to something

aguillenv (commented):
So far seems to make sense. Are we going to keep this pattern for the cases where we request for a sub-entity in the mapper? the `db.read() ?? db.cloud()` cases? What I'm not completely sure on those is how it will behave in the case we asked for a misconfigured sub-entity, we get `undefined` and then used in the parent entity. In that case it will throw anyways and I think we cannot avoid that.

src/modules/0.0.7/aws_acm_list/index.ts
dfellis: No the new `isCertType` function returns true when the string is that enum type.

If I had to make that change my current PR wouldn't compile and the common test suite would fail.

aguillenv: ok, so checking that unexpected strings are not being assigned, got it

src/modules/0.0.7/aws_route53_hosted_zones/index.ts
```diff
@@ -18,8 +18,8 @@ export const AwsRoute53HostedZoneModule: Module2 = new Module2({
     },
     resourceRecordSetMapper: async (rrs: any, ctx: Context) => {
       const out = new ResourceRecordSet();
-      // if (!(rrs.Name && rrs.Type)) throw new Error('Wrong record from AWS');
-      if (!(rrs.Name && rrs.Type)) return undefined;
+      if (!(rrs.Name && rrs.Type)) throw new Error('Wrong record from AWS');
+      // if (!(rrs.Name && rrs.Type)) return undefined;
```
dfellis, line 22: @aguillenv since you wrote this module originally. Do you have any idea why *ignoring* this busted data from AWS would actually break things?

aguillenv: Those properties must always be present. This is just an extra check in case AWS glitch or something but those properties should always be present. The name because it must have at least the base domain name and all records have a type `A|AAAA|TXT|NS|...`

dfellis: Right, but I mean, if the record we get back from AWS is broken, my change (before commenting it out and reverting back) was just dropping the record entirely from consideration, and that was causing an *update* of the domain name to fail.

Allowing the bad data to come through and throw an error is fixing things, but I don't see why just ignoring the bad data isn't good enough here?

aguillenv: ok if I'm not wrong what is happening is the following:
- The replace is triggered. The hosted zone replace just creates the new one and let the next loop delete the previous on
- In the same first iteration is trying to create the new records and delete the old ones
- In the next loop it is trying to delete the old hosted zone, but it doing something weird trying to modify/update the records, but some of these ones belonged to the old one hosted zone and they do not exists anymore, so it fails in the update 

My guess is that it is passing with the old version because the `throw` forces a new loop where all hosted zones are already in place?

I think I need to review this logic. I'll do it since I still need to update this module for the elb+r53 integration

aguillenv (approved):

PR #814, opened 2022-06-01
R53 + ELB

### LINKED ISSUE #784: Route53 to depend on ELB

aguillenv:
Resolves #784

### DIFF
--- a/test/modules/aws-route53-hosted-zones-integration.ts
+++ b/test/modules/aws-route53-hosted-zones-integration.ts
@@ -21,8 +21,8 @@ const {
 
 const prefix = getPrefix();
 const dbAlias = 'route53test';
-const domainName = `${dbAlias}${prefix}.com.`;
-const replaceDomainName = `${dbAlias}${prefix}replace.com.`;
+const domainName = `${dbAlias}${prefix}.${process.env.AWS_REGION}.com.`;
+const replaceDomainName = `${dbAlias}${prefix}replace.${process.env.AWS_REGION}.com.`;
 const resourceRecordSetName = `test.${domainName}`;
 const aliasResourceRecordSetName = `aliastest.${domainName}`;
 const resourceRecordSetMultilineName = `test.multiline.${replaceDomainName}`;


aguillenv:
To add some context:
- A `hosted_zone` has multiple `resource_record_set`s
- Each `resource_record_set` has some properties but the main ones are:
  - `record_type`: A|AAAA|CNAME|...
  - `record`: plain text value representing ips, dns, ...

The idea is to support what AWS calls `AliasTarget` that is the structure they have to relate a `resource_record_set` with other resources like ELB, S3, CloudFront, Beanstalk and API Gateway (I think those are all of them for now).

The `AliasTarget` object have the following properties:
- `HostedZoneId`: It can be easily retrieved from the resource we need
- `DNS`: the dns we want to relate with
- `HealthCheck`: a boolean to enable a health check of the resource.

I can think 2 options to add this to the current model:

Option 1: 
- Having `n` `AliasTarget` tables, one per `AliasTarget` supported (`alias_target_elb`, `alias_target_s3`, ...).
- Add gradually `n` FK columns to the `resource_record_set` table.
- Add a constraint in `resource_record_set` table so only one between `record` and any of the new alias target can have a value and the others should be `NULL`.
- Each `AliasTarget` table would look like:
  - `id`
  - `hosted_zone_id`
  - `health_check`
  - FK to the resource it is representing

Option 2:
- Having only one `AliasTarget` table with `n` FKs to every target supported.
- The `resource_record_set` table would have only one FK to the `alias_target` table.
-  Add a constraint in `resource_record_set` table so only one between `record` and the new alias target can have a value and the other should be `NULL`.
- AliasTarget` table would look like:
  - `id`
  - `hosted_zone_id`
  - `health_check`
  - `n` FK columns, one per resource it is relating.


Option 1 would keep things isolated and each resource where it belongs
Option 2 seems to be more cleaner on the `resource_record_set` side.

aguillenv:
> To add some context:
> 
> * A `hosted_zone` has multiple `resource_record_set`s
> * Each `resource_record_set` has some properties but the main ones are:
>   
>   * `record_type`: A|AAAA|CNAME|...
>   * `record`: plain text value representing ips, dns, ...
> 
> The idea is to support what AWS calls `AliasTarget` that is the structure they have to relate a `resource_record_set` with other resources like ELB, S3, CloudFront, Beanstalk and API Gateway (I think those are all of them for now).
> 
> The `AliasTarget` object have the following properties:
> 
> * `HostedZoneId`: It can be easily retrieved from the resource we need
> * `DNS`: the dns we want to relate with
> * `HealthCheck`: a boolean to enable a health check of the resource.
> 
> I can think 2 options to add this to the current model:
> 
> Option 1:
> 
> * Having `n` `AliasTarget` tables, one per `AliasTarget` supported (`alias_target_elb`, `alias_target_s3`, ...).
> * Add gradually `n` FK columns to the `resource_record_set` table.
> * Add a constraint in `resource_record_set` table so only one between `record` and any of the new alias target can have a value and the others should be `NULL`.
> * Each `AliasTarget` table would look like:
>   
>   * `id`
>   * `hosted_zone_id`
>   * `health_check`
>   * FK to the resource it is representing
> 
> Option 2:
> 
> * Having only one `AliasTarget` table with `n` FKs to every target supported.
> * The `resource_record_set` table would have only one FK to the `alias_target` table.
> * Add a constraint in `resource_record_set` table so only one between `record` and the new alias target can have a value and the other should be `NULL`.
> * AliasTarget` table would look like:
>   
>   * `id`
>   * `hosted_zone_id`
>   * `health_check`
>   * `n` FK columns, one per resource it is relating.
> 
> Option 1 would keep things isolated and each resource where it belongs Option 2 seems to be more cleaner on the `resource_record_set` side.

Discussed on discord and we decided to go with option 2

depombo (approved):

PR #815, opened 2022-06-01
only warn if no graphile worker is found on disconnect

depombo:


src/services/scheduler.ts
```diff
@@ -135,7 +135,7 @@ export async function stop(dbId: string) {
     await conn.dropConn();
     delete workerRunners[dbId];
   } else {
-    throw new Error(`Graphile worker for ${dbId} not found`);
+    logger.warn(`Graphile worker for ${dbId} not found`);
```
dfellis, line 138: This change won't work. The error we're seeing is in the destructuring of `workerRunners[dbId]`

dfellis: Though we can keep it. If there's another error it's also fine.

dfellis: I can't leave a comment/suggestion on line 127, but this is what it should be:

```ts
const { runner, conn, } = workerRunners[dbId] ?? { runner: undefined, conn: undefined, };
```

dfellis (approved):

PR #816, opened 2022-06-02
exclude iasql_ tables from rec count and add op count to iasql_databases

depombo:


src/entity/index.ts
```diff
@@ -27,6 +27,12 @@ export class IasqlDatabase {
   })
   recordCount: number;
 
+  @Column({
+    type: 'int',
+    default: 0,
+  })
+  operationCount: number;
```
aguillenv, line 34: Maybe something like this could also work and reduce some code? https://www.postgresql.org/docs/current/ddl-generated-columns.html
Not sure if can be used cross tables.

aguillenv (approved):

src/entity/index.ts
depombo: oh that would be nice! unfortunately it only works for columns within the same row

dfellis (approved):

PR #828, opened 2022-06-06
analytics data model per db for better querying

depombo:


src/config/ci.ts
```diff
@@ -0,0 +1,30 @@
+import { ConfigInterface, } from './config';
+
+const config: ConfigInterface = {
+  http: {
+    port: 8088,
+  },
+  modules: {
+    latestVersion: '0.0.7',
```
aguillenv, line 8: Hmm probably this should be already 0.0.8?

package.json
```diff
@@ -16,9 +16,7 @@
     "start:dev:watch": "yarn build && yarn serve",
     "gen-sql": "IASQL_ENV=local ts-node ./node_modules/.bin/typeorm migration:generate -n",
     "migrate": "IASQL_ENV=local ts-node ./node_modules/.bin/typeorm migration:run",
-    "gen-module": "IASQL_ENV=local ./.module_migration.sh",
-    "docker-compose:dev": "IASQL_ENV=local docker-compose -f docker-compose.yml -f docker-compose.dev.yml up --build",
-    "docker-compose": "IASQL_ENV=local docker-compose up --build"
```
aguillenv, line 21: Not sure about removing this one. Are we sure this is not the one in use in the `CONTRIBUTING.md`?

src/config/index.ts
```diff
@@ -1,11 +1,12 @@
 import { ConfigInterface, throwError, } from './config';
 
+export const ENV = process.env.IASQL_ENV;
```
aguillenv, line 3: Maybe `IASQL_ENV`?

src/config/production.ts
```diff
@@ -34,7 +34,7 @@ const config: ConfigInterface = {
     release: process.env.SENTRY_RELEASE ?? throwError('No Sentry Release defined'),
   },
   telemetry: {
-    amplitudeKey: '968524573d0f8bd2e84460099bca9353'
+    amplitudeKey: '8fd6aa1d61c115e59e35b3adfd5dd41a'
```
aguillenv, line 37: Is this the same as local? How did we said we could differentiate them if they are in the same dashboard? Or we do mot need to?

src/services/telemetry.ts
```diff
@@ -11,33 +11,33 @@ type IasqlOperationType = typeof IasqlOperationType[keyof typeof IasqlOperationT
 
 const singleton = config.telemetry ? Amplitude.init(config.telemetry.amplitudeKey) : undefined;
 
-export async function logDbConnect(dbId: string, dbAlias: string, uid: string, email: string) {
+export async function logDbConnect(dbId: string, userProps: any) {
```
aguillenv, line 14: nit: maybe add `todo` to type this

aguillenv (approved):

src/config/production.ts
depombo: yes it is the same as local. we can differentiate by environment

src/services/telemetry.ts
depombo: hmm it varies per event, but maybe I can type it now

src/config/ci.ts
depombo: ah yes

package.json
depombo: This is what we use in contributing. Removing this mostly for us, but will add back

PR #834, opened 2022-06-08
pass in rest of userProps on db operation

depombo:


src/services/scheduler.ts
```diff
@@ -75,6 +75,10 @@ export async function start(dbId: string, dbUser:string) {
         }
         let output;
         let error;
+        const user = await MetadataRepo.getUserFromDbId(dbId);
+        const uid = !user ? undefined : user.id;
```
aguillenv, line 79:
```suggestion
        const uid = user?.id;
```

src/services/scheduler.ts
```diff
@@ -75,6 +75,10 @@ export async function start(dbId: string, dbUser:string) {
...
+        const email = !user ? undefined : user.email;
```
aguillenv, line 80:
```suggestion
        const email = user?.email;
```

src/services/scheduler.ts
```diff
@@ -75,6 +75,10 @@ export async function start(dbId: string, dbUser:string) {
...
+        const dbAlias = !user ? undefined : (user.iasqlDatabases[0].alias ?? undefined);
```
aguillenv, line 81:
```suggestion
        const dbAlias = user?.iasqlDatabases?.[0]?.alias;
```

aguillenv (approved):

aguillenv (approved):

PR #840, opened 2022-06-10
[ec2] add support for user data / cloud init script

depombo:


src/modules/0.0.8/aws_ec2/migration/1654893453604-aws_ec2.ts
```diff
@@ -14,26 +14,9 @@ export class awsEc21653932287040 implements MigrationInterface {
...
-            language plpgsql security definer
-            as $$
-            declare
-                _target_group_type target_group_target_type_enum;
-            begin
-                select target_type into _target_group_type
-                from target_group
-                where target_group_name = _target_group_name;
-                return _target_group_type = 'instance';
-            end;
```
depombo, line 28: what is this used for? hmm should we have it `iasql_functions`?

depombo: or have the migration (re)generation automatically regenerate this so we don't add a manual step?

aguillenv: this is a constraint to ensure that you are only allow to assign a target group of type `instance` to a ec2 instance. For now it has to be done manually, we do not have other way to add this automatically on the migrations generation and cannot be expressed with typeorm.

depombo: got it 👍

src/modules/0.0.8/aws_ec2/index.ts
```diff
@@ -11,13 +11,16 @@ export const AwsEc2Module: Module2 = new Module2({
   ...metadata,
   utils: {
     instanceMapper: async (instance: AWSInstance, ctx: Context) => {
+      const client = await ctx.getAwsClient() as AWS;
       const out = new Instance();
-      out.instanceId = instance.InstanceId;
+      out.instanceId = instance.InstanceId as string;
```
dfellis, line 16: Maybe instead of this assertion it early returns undefined if `InstanceId` somehow comes back null from AWS?

depombo: I am almost certain this should never come back `undefined`, it is just the AWS type weirdness

depombo: got it, that makes sense. I was mostly following the pattern in other places, but we should probably replace those cases with this patten

aguillenv (approved):

dfellis (approved):

PR #851, opened 2022-06-13
Make iasql_upgrade return a value instead of only communicate via error messages

### LINKED ISSUE #820: Modify `iasql_upgrade` to return a value instead of abusing the error handling

dfellis:
Resolves #820


dfellis:
Tested with *The Button* :tm:

Screenshot from 2022-06-13 16-26-43

depombo:
> Tested with _The Button_ ™️
> 
> Screenshot from 2022-06-13 16-26-43

We need to change the main for though?

PR #854, opened 2022-06-14
Fix the disconnect regression

dfellis:
I am not entirely sure why this change breaks the disconnect flow, but I have confirmed locally that this fixes it (by simply reverting the behavior in the disconnect flow only)

### DIFF
--- a/src/services/db-manager.ts
+++ b/src/services/db-manager.ts
@@ -65,9 +65,12 @@ export function grantPostgresRoleQuery(user: string) {
   `;
 }
 
-export function dropPostgresRoleQuery(user: string) {
+export function dropPostgresRoleQuery(user: string, ephemeral = false) {
+  // Ephemeral users need to drop their owned properties first. I do not understand why only them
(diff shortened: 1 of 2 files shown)


src/services/db-manager.ts
```diff
@@ -65,9 +65,12 @@ export function grantPostgresRoleQuery(user: string) {
   `;
 }
 
-export function dropPostgresRoleQuery(user: string) {
+export function dropPostgresRoleQuery(user: string, ephemeral = false) {
+  // Ephemeral users need to drop their owned properties first. I do not understand why only them
+  // and not the originally-created users for the database. Nor do I understand why dropping for
+  // the originally created users causes issues.
```
aguillenv, line 71: what's the error message from the database when running `DROP OWNED BY ${user}`?

dfellis: This one: https://sentry.io/organizations/iasql/issues/3260895664/?project=6327420&referrer=webhooks_plugin

It's very strange.

depombo (approved):

PR #855, opened 2022-06-14
Private NAT Gateway

aguillenv:
Part of #844


src/modules/0.0.9/aws_vpc/index.ts
```diff
@@ -184,5 +212,78 @@ export const AwsVpcModule: Module2 = new Module2({
         },
       }),
     }),
+    natGateway: new Mapper2<NatGateway>({
+      entity: NatGateway,
+      equals: (a: NatGateway, b: NatGateway) => Object.is(a.connectivityType, b.connectivityType)
+        // && Object.is(a.elasticIp, b.elasticIp) TODO: add elastic ip
```
dfellis, line 218: Do we want to hold this PR until the EIP PR is landed, or would you rather do a follow-up PR for this?

aguillenv: I'd like to do it in a follow up task after this one and #858 are merged

depombo (approved):

PR #856, opened 2022-06-14
Cleanup bad dbs from the metadata database on startup

dfellis:


aguillenv (approved):

dfellis:
> interesting approach. we definitely need this here. we should also probably write more defensible code in disconnect too since this will only run on startup? does not have to be part of this PR

So I already fixed the glitch in disconnect. This is to clear out the leftovers, but the point of this on-startup check is to handle *future* errors gracefully with the next deploy, since the failure was due to a crash *during* the disconnect logic.

PR #857, opened 2022-06-14
Only run the drop command on ephemeral accounts locally

dfellis:
After testing in staging and production, another anomaly has been detected where the resource dropping step necessary when running locally doesn't seem to apply and actually causes a crash.

The reason for this difference between the environments is unknown, but may be related to AWS's light forking of the OSS databases they run in RDS? In any case, this should eliminate the sentry report on this and prevent the postgres server from filling up with garbage users.

### DIFF
--- a/src/config/ci.ts
+++ b/src/config/ci.ts
@@ -15,6 +15,7 @@ const config: ConfigInterface = {
     port: 5432,
     forceSSL: false,
     sqlViaRest: true,
+    dropBeforeUserDeletion: true,
   },
   logger: {
     debug: true,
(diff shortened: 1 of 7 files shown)


depombo (approved):

dfellis:
So... it fails the way local was failing with this when I test in staging. There's something weirder going on here that I need to figure out to get the running sql via rest with restricted ephemeral accounts thing really working.

aguillenv (approved):

dfellis:
After a good night's rest and rereading the documentation, I found this: [DROP OWNED is often used to prepare for the removal of one or more roles. Because DROP OWNED only affects the objects in the current database, it is usually necessary to execute this command in each database that contains objects owned by a role that is to be removed.](https://www.postgresql.org/docs/10/sql-drop-owned.html)

`DROP OWNED` was being run under the wrong connection before.

PR #858, opened 2022-06-15
VPC Elastic IP

### LINKED ISSUE #848: Add elastic IP module, `aws_eip`

aguillenv:
Resolves #848


src/services/gateways/aws.ts
```diff
@@ -47,6 +47,10 @@ import {
   DescribeNetworkInterfacesCommand,
   EC2,
   Tag,
+  AllocateAddressCommand,
+  DescribeAddressesCommand,
+  ReleaseAddressCommand,
+  AllocateAddressCommandInput,
 } from '@aws-sdk/client-ec2'
```
depombo, line 54: can we use `this.EC2` which doesn't require these imports and then instantiating the objects?

aguillenv: ohh I can take a look at it 👍🏼

PR #864, opened 2022-06-15
Fix empty error logs

depombo:
Also DRY the connect handler

### DIFF
--- a/src/router/db.ts
+++ b/src/router/db.ts
@@ -38,6 +38,7 @@ async function connectHandler(req: any, res: any) {
     res.status(500).end(error);
     telemetry.logConnect({ uid, email }, { error }, dbId);
   }
+}
 
 db.get('/connect/:dbAlias', connectHandler);
 
--- a/src/router/db.ts
+++ b/src/router/db.ts
@@ -5,14 +5,14 @@ import * as dbMan from '../services/db-manager';
 import * as iasql from '../services/iasql'
 import MetadataRepo from '../services/repositories/metadata'
 import * as telemetry from '../services/telemetry'
-import logger, { logUserErr } from '../services/logger'
+import logger, { logErrSentry } from '../services/logger'
 import config from '../config'
 
 export const db = express.Router();
 
-db.get('/connect/:dbAlias', async (req, res) => {
+async function connectHandler(req: any, res: any) {
   logger.info('Calling /connect');
-  const {dbAlias} = req.params;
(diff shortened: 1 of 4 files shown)


src/services/scheduler.ts
```diff
@@ -164,9 +164,11 @@ export async function init() {
   if (!MetadataRepo.initialized) await MetadataRepo.init(); // Necessary in the child process
   const dbs: IasqlDatabase[] = await MetadataRepo.getAllDbs();
   const inits = await Promise.allSettled(dbs.map(db => start(db.pgName, db.pgUser)));
-  for (const bootstrap of inits) {
+  for (const [i, bootstrap] of inits.entries()) {
     if (bootstrap.status === 'rejected') {
-      logger.error(bootstrap.reason);
+      const db = dbs[i];
+      const user = await MetadataRepo.getUserFromDbId(db.pgName);
+      logUserErr(`Failed to bootstrap a db on startup. Reason: ${bootstrap.reason}`, user?.id, user?.email, db.alias);
```
depombo, line 171: this was one of the empty logs

src/router/db.ts
```diff
@@ -5,14 +5,14 @@ import * as dbMan from '../services/db-manager';
...
+import logger from '../services/logger'
 import config from '../config'
 
 export const db = express.Router();
 
-db.get('/connect/:dbAlias', async (req, res) => {
+async function connectHandler(req: any, res: any) {
   logger.info('Calling /connect');
-  const {dbAlias} = req.params;
+  const dbAlias = req.params.dbAlias ?? req.body.dbAlias;
```
aguillenv, line 15: I don't remember is `params` is always available for `req`, I think it is but just in case:
```suggestion
  const dbAlias = req.params?.dbAlias ?? req.body?.dbAlias;
```

src/services/iasql.ts
```diff
@@ -16,7 +16,7 @@ import * as dbMan from './db-manager'
 import * as scheduler from './scheduler-api'
 import MetadataRepo from './repositories/metadata'
 import config from '../config'
-import logger, { debugObj } from './logger'
+import logger, { debugObj, logErrSentry } from './logger'
```
aguillenv, line 19: why this import here? It does not seem in use

aguillenv (approved):

dfellis (approved):

dfellis (approved):

PR #865, opened 2022-06-17
DRY up aws gateway

depombo:


src/services/gateways/aws.ts
```diff
@@ -1045,19 +875,15 @@ export class AWS {
...
   async deleteLoadBalancer(arn: string) {
-    await this.elbClient.send(
-      new DeleteLoadBalancerCommand({ LoadBalancerArn: arn, })
-    );
+    await this.elbClient.deleteLoadBalancer({ LoadBalancerArn: arn, });
     // We wait it is completely deleted to avoid issues deleting dependent resources.
-    const input = new DescribeLoadBalancersCommand({ LoadBalancerArns: [arn], });
-    await createWaiter<ElasticLoadBalancingV2Client, DescribeLoadBalancersCommand>(
+    const input: DescribeLoadBalancersCommandInput = { LoadBalancerArns: [arn], };
+    await createWaiter<ElasticLoadBalancingV2, DescribeLoadBalancersCommandInput>(
```
aguillenv, line 886: Are you sure this is ok? For the waiters I think we still need to pass the old syntax

depombo: Yeah, it still fulfills the interface and it works!

aguillenv (approved):

src/services/gateways/aws_2.ts
```diff
@@ -0,0 +1,1912 @@
+import {
```
dfellis, line 1: Now there are two of them

dfellis (approved):

dfellis (approved):

PR #881, opened 2022-06-20
exclude spot instances to properly re-add them later

depombo:


aguillenv (approved):

src/services/gateways/aws_2.ts
```diff
@@ -421,7 +421,7 @@ export class AWS {
   }
 
   async getAMIs() {
-    return this.ec2client.describeImages({});
+    return await this.ec2client.describeImages({});
```
dfellis, line 424: These changes shouldn't affect anything since the `async` function returns a promise to the caller and you aren't intercepting errors. Basically identical behavior but with nested promises, now.

depombo: I just went back to the behaviour of the original gateway, but I see your point

aguillenv (approved):

PR #882, opened 2022-06-20
Verify disconnect works for old and busted databases

### LINKED ISSUE #860: Test that ensures old and unsupported versions are disconnected successfully

dfellis:
Resolves #860

Also fixes an issue with creating a new database with the wrong migrations if the `latest` variable in the Modules export doesn't match `latestVersion` in the environment in question.

TODO: Eliminate all uses in the codebase of the `latest` export, though the rest *probably* aren't as big of a deal.

### DIFF
--- a/src/config/ci.ts
+++ b/src/config/ci.ts
@@ -6,7 +6,7 @@ const config: ConfigInterface = {
   },
   modules: {
     latestVersion: '0.0.9',
-    oldestVersion: '0.0.5',
+    oldestVersion: '0.0.6',
   },
   db: {
     host: 'postgresql',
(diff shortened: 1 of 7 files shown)


src/config/ci.ts
```diff
@@ -6,7 +6,7 @@ const config: ConfigInterface = {
   },
   modules: {
     latestVersion: '0.0.9',
-    oldestVersion: '0.0.5',
+    oldestVersion: '0.0.6',
```
dfellis, line 9: We forgot to update these in the `prepare` PR and my new change just made that fatal, so that's good. (Ideally we don't need to set these values explicitly at all, but...)

depombo (approved):

PR #883, opened 2022-06-21
Force test exit and do not error on misconfigured resources

aguillenv:
The test suite is getting hanged up sometimes because the engine workers keep trying to connect with the database that is already down. Here is an example:https://github.com/iasql/iasql-engine/runs/6981051342?check_suite_focus=true#step:5:5695

We have to look deeper in that error, but for now we force jest to exit so we avoid long running tests that already failed.

Also, ignore some misconfigured resources instead of throwing an error.

### DIFF
--- a/src/modules/0.0.9/aws_elb/index.ts
+++ b/src/modules/0.0.9/aws_elb/index.ts
@@ -41,7 +41,7 @@ export const AwsElbModule: Module2 = new Module2({
           out.actionType = (a.Type as ActionTypeEnum);
           out.targetGroup =  await AwsElbModule.mappers.targetGroup.db.read(ctx, a?.TargetGroupArn) ??
             await AwsElbModule.mappers.targetGroup.cloud.read(ctx, a?.TargetGroupArn);
-          if (!out.targetGroup) throw new Error('Target groups need to be loaded first');
+          if (!out.targetGroup) return undefined;
         }
       }
       if (l.SslPolicy && l.Certificates?.length) {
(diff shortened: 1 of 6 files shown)


src/modules/0.0.9/aws_elb/index.ts
```diff
@@ -41,7 +41,7 @@ export const AwsElbModule: Module2 = new Module2({
           out.actionType = (a.Type as ActionTypeEnum);
           out.targetGroup =  await AwsElbModule.mappers.targetGroup.db.read(ctx, a?.TargetGroupArn) ??
             await AwsElbModule.mappers.targetGroup.cloud.read(ctx, a?.TargetGroupArn);
-          if (!out.targetGroup) throw new Error('Target groups need to be loaded first');
+          if (!out.targetGroup) return undefined;
```
dfellis, line 44: If the target group can be loaded out of order, wouldn't this change break in legitimate situations? That's why I didn't change this one.

dfellis (approved):
LGTM assuming my concern is unjustified.

src/modules/0.0.9/aws_elb/index.ts
aguillenv: Hmmm I see your point. But, if it is in the wrong order it means it will not find it on `db` but if that's true, it will be `undefined` and we will default to the `cloud` path where it should exists if any, right?

dfellis: That makes sense, but then I wonder what this was guarding against in the first place?

aguillenv: Well the possibility that for some reason the target group is not  in `cloud` either. That mean something is wrong and misconfigured. Because a listener cannot be created without a target group to forward to.

dfellis: Okay, so we enforce the target group via non-nullable foreign key relations in the database, right?

So any load balancer in the cloud without a target group is some misconfiguration on the other side, then? There is no situation where a race condition from creating both of them in the not-too-distant past would cause one API to return null while the other succeeds?

In that case I don't have an issue with this, but if there is some race condition situation possible, can we add a singular delayed retry, say 10-20 seconds later, to try to catch that?

depombo (approved):

PR #884, opened 2022-06-21
Prepare for development of v0.0.10

dfellis:


src/config/production.ts
```diff
@@ -6,7 +6,7 @@ const config: ConfigInterface = {
   },
   modules: {
     latestVersion: '0.0.9',
-    oldestVersion: '0.0.6',
+    oldestVersion: '0.0.7',
```
aguillenv, line 9: shouldn't we do this with the production release? otherwise if we need to do a hot fix or a deploy during the week we could stop giving support to this version?

PR #886, opened 2022-06-23
VPC endpoint gateway

### LINKED ISSUE #868: Research and implement VPC endpoint

aguillenv:
Resolves #868 

In AWS exists 3 different types of VPC endpoints (Gateway, Interface and Gateway LoadBalancer). Each of them use different properties of the API. I've decided to break them into different entities to avoid having several check constraints in the table like "if type A field B must exist but not field C", that with 3 different types I think will be difficult to follow.

I've implemented the 'Gateway' type since it is the simpler case and the one we really need to demonstrate https://twitter.com/t3rabytes/status/1534885118048878592?s=10&t=P5TnAfGTEXbezn9lTa_Dtg

### DIFF
--- a/src/modules/0.0.10/aws_vpc/index.ts
+++ b/src/modules/0.0.10/aws_vpc/index.ts
@@ -21,10 +21,7 @@ import {
   NatGatewayState,
   ElasticIp,
   EndpointGateway,
-  DnsRecordIpType,
-  IpAddressType,
   EndpointGatewayService,
-  EndpointGatewayState,
 } from './entity'
 import { Context, Crud2, Mapper2, Module2, } from '../../interfaces'
 import * as metadata from './module.json'
@@ -104,6 +101,7 @@ export const AwsVpcModule: Module2 = new Module2({
       if (!out.vpc) return undefined;
       out.policyDocument = eg.PolicyDocument;
       out.state = eg.State;
+      out.routeTableIds = eg.RouteTableIds;
       if (eg.Tags?.length) {
         const tags: { [key: string]: string } = {};
         eg.Tags.filter((t: any) => !!t.Key && !!t.Value).forEach((t: any) => {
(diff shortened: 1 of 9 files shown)


test/modules/aws-vpc-integration.ts
```diff
@@ -62,70 +77,85 @@ describe('VPC Integration Testing', () => {
...
-  it('applies the public nat gateway with no existing elastic ip change', apply());
-
-  it('checks public nat gateway with no existing elastic ip count', query(`
-    SELECT * FROM nat_gateway WHERE tags ->> 'Name' = '${pubNg2}';
-  `, (res: any) => expect(res.length).toBe(1)));
-
-  it('checks public nat gateway with no existing elastic ip count', query(`
-    SELECT * FROM elastic_ip WHERE tags ->> 'Name' = '${pubNg2}';
-  `, (res: any) => expect(res.length).toBe(1)));
+  describe('Elastic IP and nat gateway creation', () => {
```
aguillenv, line 80: I've created these inner `describe`s to separate the logic in the tests. NATs adds too much time to test execution. Having it in separate `describe`s let us comment what we do not want to execute then and quickly just run the new ones and make sure it is working without waiting that long locally.

src/modules/0.0.10/aws_vpc/index.ts
```diff
@@ -398,5 +428,149 @@ export const AwsVpcModule: Module2 = new Module2({
...
+          // Update vpc endpoint
+          // !Object.is(a.policyDocument, b.policyDocument)
+          // !Object.is(a.routeTableIds, b.routeTableIds)
+          // Update tags
+          // !AwsVpcModule.utils.eqTags(a.tags, b.tags)
+          // Restore
+          // !Object.is(a.state, b.state)
+          // Replace
+          // !Object.is(a.vpc?.vpcId, b.vpc?.vpcId)
+          // !Object.is(a.service, b.service)
```
dfellis, line 508: Delete this?

aguillenv: Oh that was my guide to see what need to be done in each case. I'll delete it

dfellis (approved):

PR #892, opened 2022-06-23
Unify IAM mappers + Update nulls after save 

### LINKED ISSUE #885: Typeorm hook seem not doing the work to change `null` to `undefined`

aguillenv:
Resolves #885


dfellis (approved):

src/services/typeorm.ts
```diff
@@ -130,4 +131,19 @@ export class TypeormWrapper {
   createQueryRunner() {
     return this.connection.createQueryRunner();
   }
+
+  updateNulls(value: any): void {
+    if (!value) return;
+    if (Array.isArray(value)) {
+      for (const v of value) {
+        Object.keys(v).forEach((k: string) => {
```
aguillenv, line 139:
```suggestion
        Object.keys(v ?? {}).forEach((k: string) => {
```

src/modules/0.0.10/aws_iam/index.ts
```diff
@@ -35,7 +37,7 @@ export const AwsIamModule: Module2 = new Module2({
         // we are trusting aws won't change it from under us
         Object.is(a.assumeRolePolicyDocument, b.assumeRolePolicyDocument) &&
         Object.is(a.attachedPoliciesArns?.length, b.attachedPoliciesArns?.length) &&
-        a.attachedPoliciesArns?.every(as => !!b.attachedPoliciesArns?.find(bs => Object.is(as, bs))),
+        ((!a.attachedPoliciesArns && !b.attachedPoliciesArns) || !!a.attachedPoliciesArns?.every(as => !!b.attachedPoliciesArns?.find(bs => Object.is(as, bs)))),
```
depombo, line 40: we should probably have a global util for comparing the equality two lists? We don’t check if `a` and `b` are undefined in other mappers

aguillenv: It is true, what we do in other similar cases is default to `false`. I'll replicate that behaviour here.

dfellis (approved):

PR #894, opened 2022-06-24
Part 1: Macro-ify AWS client usage

### LINKED ISSUE #861: Dissolve AWS Gateway

dfellis:
Part of #861

### DIFF
--- a/src/modules/0.0.10/aws_acm_import/index.ts
+++ b/src/modules/0.0.10/aws_acm_import/index.ts
@@ -1,14 +1,14 @@
-import { ImportCertificateCommandInput, paginateListCertificates, } from '@aws-sdk/client-acm'
+import { ACM, ImportCertificateCommandInput, paginateListCertificates, } from '@aws-sdk/client-acm'
 import { AWS, paginateBuilder, } from '../../../services/aws_macros'
 import { Context, Crud2, Mapper2, Module2, } from '../../interfaces'
 import { AwsAcmListModule } from '../aws_acm_list'
 import { CertificateImport } from './entity'
 import * as metadata from './module.json'
 
-const getCertificatesSummary = paginateBuilder(paginateListCertificates, 'CertificateSummaryList');
+const getCertificatesSummary = paginateBuilder<ACM>(paginateListCertificates, 'CertificateSummaryList');
 
 // TODO: Can I macro this somehow?
-async function importCertificate(client: any, input: ImportCertificateCommandInput) {
+async function importCertificate(client: ACM, input: ImportCertificateCommandInput) {
   const res = await client.importCertificate(input);
   const arn = res.CertificateArn ?? '';
   let certificates: string[] = [];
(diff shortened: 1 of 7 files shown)


src/modules/0.0.10/aws_acm_import/index.ts
```diff
@@ -1,10 +1,28 @@
...
+async function importCertificate(client: any, input: ImportCertificateCommandInput) {
+  const res = await client.importCertificate(input);
+  const arn = res.CertificateArn ?? '';
+  let certificates: string[] = [];
+  let i = 0;
+   // Wait for ~1min until imported cert is available
+  do {
+    const start = Date.now();
+    await new Promise(r => setTimeout(r, 2000));
+    while (Date.now() - start < 2000); // Sleep for 2s
```
aguillenv, line 20: oh you re copying here the fixed hot-loops issue

src/services/aws_macros.ts
```diff
@@ -0,0 +1,102 @@
...
+  pageName?: string,
+  pageSize = 25
+): ((client: any) => Promise<any[]>) {
+  if (pageName) {
+    return async (client: any) => {
+      const vals = [];
+      const paginator = paginateFn({
+        client,
+        pageSize,
+      }, {});
```
aguillenv, line 67: that `{}` should be also a variable input for this function. I think it is used somewhere to filter.

src/services/aws_macros.ts
```diff
@@ -0,0 +1,102 @@
...
+      }
+      return vals;
+    };
+  }
+}
+
+export function crudBuilder(
+  methodName: string,
+  argMapper: (args: any[]) => any,
+  retFormatter?: (arg0: any) => any
```
aguillenv, line 93: what does `ret` stands for here?

aguillenv: oh a return like function?

aguillenv (commented):
looking good 🚀

src/modules/0.0.10/aws_acm_import/index.ts
dfellis: :facepalm:

src/services/aws_macros.ts
dfellis: I was only going to make it available *if* we have a usage that exploits it. If not, then we can do this and simplify the macro functions.

src/services/aws_macros.ts
dfellis: It formats the returned value, so you get a response from AWS, then do some minor transform to either get a sub-object or mutate in some way or the other.

src/modules/0.0.10/aws_acm_list/index.ts
```diff
@@ -62,11 +88,12 @@ export const AwsAcmListModule: Module2 = new Module2({
         read: async (ctx: Context, id?: string) => {
           const client = await ctx.getAwsClient() as AWS;
           if (id) {
-            const rawCert = await client.getCertificate(id);
+            // const rawCert = await client.getCertificate(id);
```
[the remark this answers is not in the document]
dfellis:
```suggestion
```

aguillenv (approved):

PR #906, opened 2022-06-28
Convert the ECR module to the new AWS macros

dfellis:


src/services/aws_macros.ts
```diff
@@ -88,7 +88,7 @@ export function paginateBuilder<T>(
 }
 
 export function crudBuilder<T>(
-  methodName: string,
+  methodName: keyof T,
```
depombo, line 91: Nice

depombo (approved):

PR #907, opened 2022-06-29
EC2 + IAM integration

### LINKED ISSUE #838: Add `aws_iam` as a dependency for `aws_ec2`

aguillenv:
Resolves #838


src/modules/0.0.11/aws_iam/index.ts
```diff
@@ -46,13 +62,23 @@ export const AwsIamModule: Module2 = new Module2({
           for (const role of es) {
             const roleArn = await client.newRoleLin(
               role.roleName,
-              role.assumeRolePolicyDocument,
+              JSON.stringify(role.assumeRolePolicyDocument),
```
dfellis, line 65: Is this actually necessary, or was it a change you made while debugging that may not be necessary.

If it is necessary, how did we not catch this issue before, basically?

src/modules/0.0.11/aws_iam/index.ts
```diff
@@ -46,13 +62,23 @@ export const AwsIamModule: Module2 = new Module2({
...
+            try {
+              const allowEc2Service = AwsIamModule.utils.allowEc2Service(role);
+              if (allowEc2Service) {
+                await client.createInstanceProfile(role.roleName);
+                await client.attachRoleToInstanceProfile(role.roleName);
+              }
+            } catch (err: any) {
+              // Do not throw if for any reason this does not work
+              logger.warn(err.message ?? 'Error attaching instance profile on role create');
+            }
```
dfellis, line 81: Why do we not error out in this scenario?

src/modules/0.0.11/aws_iam/index.ts
```diff
@@ -92,8 +118,18 @@ export const AwsIamModule: Module2 = new Module2({
...
+              try {
+                const allowEc2Service = AwsIamModule.utils.allowEc2Service(e);
+                if (allowEc2Service) {
+                  await client.createInstanceProfile(e.roleName);
+                  await client.attachRoleToInstanceProfile(e.roleName);
+                }
+              } catch (err: any) {
+                // Do not throw if for any reason this does not work
+                logger.warn(err.message ?? 'Error attaching instance profile on role update');
+              }
```
dfellis, line 132: Same here, this seems weird to me?

src/services/gateways/aws_2.ts
```diff
@@ -1868,4 +1868,30 @@ export class AWS {
...
+      InstanceProfileName: roleName,
+    });
+  }
+
+  async detachRoleToInstanceProfile(roleName: string) {
+    await this.iamClient.removeRoleFromInstanceProfile({
+      InstanceProfileName: roleName,
+      RoleName: roleName,
+    });
+  }
```
dfellis, line 1896: Can you convert these to the macro form?

src/modules/0.0.11/aws_iam/index.ts
aguillenv: oh no, this is because before the `role.assumeRolePolicyDocument ` was a `string` and now I've changed it to be `jsonb`. The AWS API still need it to be a `string`, though.

dfellis: Ahh, got it.

src/services/gateways/aws_2.ts
aguillenv: I thought about it but I wasn't sure if do it in this PR or keep that change in a separate PR

src/modules/0.0.11/aws_iam/index.ts
aguillenv: Here because if the role already exists, and we are updating adding something else to the policy, the attachment already exists and would fail on duplication. But, If we are changing the policy role to support now ec2, we want to create the attachment.

aguillenv: but now that I think of it it is not being handled the case where it was ec2 and now it is being something else, we need to detach the instance profile. I'll work on that

src/modules/0.0.11/aws_iam/index.ts
aguillenv: hmmm thought it would make sense after I did the update. But yeah I'll remove this, same with delete

src/modules/0.0.11/aws_iam/index.ts
```diff
@@ -15,27 +105,42 @@ export const AwsIamModule: Module2 = new Module2({
...
     },
     roleNameFromArn: (arn: string, _ctx: Context) => {
-      const roleName = arn.split(':role/')[1];
+      // Role name is always the last element
       // Regular role example - arn:aws:iam::547931376551:role/AWSECSTaskExecution
       // AWS service role example - arn:aws:iam::547931376551:role/aws-service-role/ecs.amazonaws.com/AWSServiceRoleForECS
-      return roleName.split('/').length === 3 ? roleName.split('/')[2] : roleName;
-    }
+      // EC2 role instance profile ARN example - arn:aws:iam::257682470237:instance-profile/test-role
+      return  arn.split('/').pop();
```
depombo, line 123: much better

depombo (approved):

dfellis (commented):
Mostly style suggestions

src/modules/0.0.11/aws_iam/index.ts
```diff
@@ -1,10 +1,99 @@
...
+    Description: description
+  }),
+);
+
+const detachRolePolicy = crudBuilder<IAM>(
+  'detachRolePolicy',
+  (roleName: string, policyArn: string) => ({ RoleName: roleName, PolicyArn: policyArn, }),
+);
+const detachRolePolicies = (client: IAM, roleName: string, policyArns: string[]) => mapLin(
+  Promise.resolve(policyArns),
```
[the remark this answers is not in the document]
aguillenv: Yes, since the `mapLin` receives is a `Promise<any []>` I'm passing the resolved promise directly. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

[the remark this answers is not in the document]
dfellis: Oh, why not modify `mapLin` to also accept a bare array? eg, `arrProm: Promise<any[] | undefined> | any[]`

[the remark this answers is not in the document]
aguillenv: Hmm well I thought it was easier this way, and no need to modify the function signature and add a new branch in the `mapLin` code for the `array` case or the `Promise` case.

dfellis (approved):

src/modules/0.0.11/aws_iam/index.ts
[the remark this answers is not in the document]
aguillenv: oh I see. ok if no change need it, it will be more natural the `mapLin` invocation

PR #912, opened 2022-06-29
Convert 'aws_s3' to the macros and create 'crudBuilder2' for better type safety

dfellis:


src/modules/0.0.11/aws_s3/index.ts
```diff
@@ -16,16 +32,16 @@ export const AwsS3Module: Module2 = new Module2({
...
         read: async (ctx: Context, id?: string) => {
           const client = await ctx.getAwsClient() as AWS;
-          const allBuckets = await client.getBuckets();
+          const allBuckets = await getBuckets(client.s3Client);
           return allBuckets
-            .filter(b => !id || b.Name === id)
-            .filter(b => !!b.Name)
-            .map(b => {
+            .filter((b: any) => !id || b.Name === id)
+            .filter((b: any) => !!b.Name)
```
depombo, line 43: hmm can we actually type this instead of using `any`?

dfellis: Oh, we shouldn't need that anymore. That was actually the impetus that got me to start fixing `crudBuilder`!

depombo (approved):

src/modules/0.0.11/aws_cloudwatch/index.ts
```diff
@@ -21,10 +16,10 @@ const getLogGroups = paginateBuilder<CloudWatchLogs>(
 const getLogGroup = async (client: CloudWatchLogs, groupName: string) => (
   await getLogGroups(client)
 ).find(lg => lg.logGroupName === groupName);
-const deleteLogGroup = crudBuilder<CloudWatchLogs>(
+const deleteLogGroup = crudBuilderFormat<CloudWatchLogs, 'deleteLogGroup', undefined>(
   'deleteLogGroup',
-  (logGroupName: string) => ({ logGroupName, }),
-  (_lg: DeleteLogGroupCommandOutput) => undefined,
+  (logGroupName) => ({ logGroupName, }),
+  (_lg) => undefined,
```
aguillenv, line 22: nit: This one (and all the ones returning  `undefined`) could just be `crudBuilder2` instead of `crudBuilderFormat`, right?

dfellis: So I used the format path here (which was baked into `crudBuilder`) to mimic the exact behavior of the AWS Gateway function it is replacing.

I don't mind making a change like that, but I have been trying to keep behavior-modifying changes to a minimum in this PR (only making them where the improved type safety tells me there's a bug in the code we missed before).

PR #915, opened 2022-06-30
Macro-ify RDS

dfellis:


aguillenv (approved):

src/modules/0.0.11/aws_rds/index.ts
```diff
@@ -1,6 +1,20 @@
...
+  DBInstance,
+  paginateDescribeDBInstances,
+  DeleteDBInstanceMessage,
+  paginateDescribeDBParameters,
+  paginateDescribeDBParameterGroups,
+} from '@aws-sdk/client-rds'
+import { createWaiter, WaiterState } from '@aws-sdk/util-waiter'
 
-import { AWS, } from '../../../services/gateways/aws_2'
+import { AWS, crudBuilder2, crudBuilderFormat, paginateBuilder, mapLin, } from '../../../services/aws_macros'
```
depombo, line 17: what is the difference between `crudBuilder2` and `crudBuilderFormat` or when should each be used?

depombo (approved):

src/modules/0.0.11/aws_rds/index.ts
dfellis: That was in the prior PR from last night. :)

`crudBuilder2` only accepts two parameters, the name of the AWS method to use, and conversion of the input arguments into the `input` parameter to feed into the AWS method.

`crudBuilderFormat` has a third parameter which takes the response of the AWS method and reformats it into whatever form you want. I could not get that piece to be typesafe without a third generic parameter, though, which is the return type for the resulting function, so it's not "auto inferred" the way `crudBuilder2` is.

Personally I'd just stick with `crudBuilder2` for a more compact representation that also gives you results you can look up in [the AWS JS SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/), but `crudBuilderFormat` is good for replicating behaviors we had already added in the AWS Gateway library.

PR #916, opened 2022-06-30
Add gp ebs support to `aws_ec2` + Let ec2 tests run multi region

aguillenv:
Resolves: #914


src/modules/0.0.11/aws_ebs/entity/general_purpose_volume.ts
```diff
@@ -0,0 +1,114 @@
...
+    enum: VolumeState,
+  })
+  state?: VolumeState;
+
+  @ManyToOne(() => Instance, instance => instance.id, {
+    eager: true,
+    nullable: true,
+  })
+  @JoinColumn({ name: 'attached_instance_id', })
+  attachedInstance?: Instance;
```
depombo, line 74: Interesting, I assumed the relationship here would be inverted and ec2 would own the fk relationship. This looks good too. I didn’t realize it was possible to attach multiple volumes / hard drives to a single instance, but I suppose that makes sense. Is it possible to attach a single volume to multiple instances? Not that we should support it on the first pass, mostly curious.

src/modules/0.0.11/aws_ebs/aws_helper.ts
```diff
@@ -0,0 +1,169 @@
+import { CreateVolumeCommandInput, DescribeVolumesCommandInput, EC2, paginateDescribeVolumes, Tag, Volume } from '@aws-sdk/client-ec2';
```
depombo, line 1: nit: probably just `aws`?

depombo:
I’m curious - why did you decide to make this a separate module to ec2? nothing else can really own an EBS volume, right? might make more sense in the ec2 module?

src/modules/0.0.11/aws_ebs/entity/general_purpose_volume.ts
dfellis: No, you definitely can't attach a single volume to multiple instances. You can detach a volume from one and re-attach to another, but AWS won't auto-NFS these volumes or anything like that (since that's a completely different layer of the stack in most OSes and more importantly varies between OSes; they can't just emulate a hard drive on a PCI bus and be done with it).

aguillenv: Actually, one specific type of volume (not the general purpose I've mapped here) can be attached to multiple instances, but I think is to a specific instance type (nitro instances or something like that).

That was one of the reason I've decided to split the volume types and here just mapped the general purpose ones.

aguillenv:
> I’m curious - why did you decide to make this a separate module to ec2? nothing else can really own an EBS volume, right? might make more sense in the ec2 module?

Hmm that's fair. My idea was to have them separate and handle ebs volumes if you know what you are doing and you want to do it.

I'll check out if ebs can only be attached to other resources other than ec2 instances. That could be a good point to take that decision.

aguillenv:
> > I’m curious - why did you decide to make this a separate module to ec2? nothing else can really own an EBS volume, right? might make more sense in the ec2 module?
> 
> Hmm that's fair. My idea was to have them separate and handle ebs volumes if you know what you are doing and you want to do it.
> 
> I'll check out if ebs can only be attached to other resources other than ec2 instances. That could be a good point to take that decision.

In the end, I've followed your suggestion @depombo and include the ebs entity into the aws_ec2 module. Ebs could not be linked to anything else and it also made less weird some of the integration between the two entities.

src/modules/0.0.11/aws_ec2/entity/general_purpose_volume.ts
```diff
@@ -0,0 +1,114 @@
...
+  CREATING = "creating",
+  DELETED = "deleted",
+  DELETING = "deleting",
+  ERROR = "error",
+  IN_USE = "in-use",
+}
+
+@Unique('Unique_gp_instance_device_name', ['instanceDeviceName', 'attachedInstance'])
+@Entity()
+export class GeneralPurposeVolume {
```
depombo, line 34: curious, why not just volume?

test/modules/aws-ec2-integration.ts
```diff
@@ -3,10 +3,9 @@ import * as iasql from '../../src/services/iasql'
 import { getPrefix, runQuery, runInstall, runUninstall, runApply, finish, execComposeUp, execComposeDown, runSync, } from '../helpers'
 
 const dbAlias = 'ec2test';
-// specific to us-west-2, varies per region
-const region = 'us-west-2'
-const amznAmiId = 'ami-06cffe063efe892ad';
-const ubuntuAmiId = 'ami-0892d3c7ee96c0bf7';
+const region = process.env.AWS_REGION;
+const amznAmiId = 'resolve:ssm:/aws/service/ami-amazon-linux-latest/amzn2-ami-hvm-x86_64-gp2';
+const ubuntuAmiId = 'resolve:ssm:/aws/service/canonical/ubuntu/server/20.04/stable/current/amd64/hvm/ebs-gp2/ami-id';
```
depombo, line 8: wait, this works across regions!?! that is amazing!

depombo: we should update the docs

depombo (approved):

test/modules/aws-ec2-integration.ts
aguillenv: yes!

aguillenv: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/finding-an-ami.html

PR #921, opened 2022-07-01
Fix the 'runSql' bug triggered by a user this morning

dfellis:
It was trivial enough that there's no real point in opening a ticket for it.


dfellis:
1. Every connection gets a unique name, even if it *shouldn't* collide with anything else.
2. The `finally` block is run after the `try` block *and* the `catch` block, so we can't assume every line of the `try` block was executed, so run the methods only if the connection was actually created.

aguillenv (approved):

PR #922, opened 2022-07-01
Improve security group stability

dfellis:
I have noticed since the security group AWS client was converted it has become less stable. I don't know why as nothing logically changed with the AWS query logic, but I spotted one place where we don't check the output correctly, so I have added a guard there, which will hopefully fix the stability problems we're seeing.


dfellis:
It's failing on the cache clearing step because the cache is already empty? https://github.com/iasql/iasql-engine/runs/7151827876?check_suite_focus=true#step:5:739

aguillenv (approved):

PR #926, opened 2022-07-02
WIP: Fix macro input typing

dfellis:


src/services/aws_macros.ts
```diff
@@ -100,22 +100,22 @@ export function paginateBuilder<T>(
...
 ) {
-  return async (client: T, ...args: any[]): Promise<V> => retFormatter(
+  return async (client: T, ...args: A): Promise<V> => retFormatter(
     (await (client[methodName] as T[U] extends Function ? T[U] : any)(argMapper(...args))) as PromiseReturnType<T[U]>,
     ...args
   );
 }
 
-export function crudBuilder2<T, U extends keyof T>(
+export function crudBuilder2<T, U extends keyof T, A extends any[] = any[]>(
```
aguillenv, line 114: oh, this was de-prioritized right? I found out the crudBuilder2 was not typing the input object while developing memory db

dfellis: Yeah :(

Maybe we can bump the priority back up, but it requires some deep Typescript type system knowledge to do, so not sure who wants to tackle it.

PR #938, opened 2022-07-05
Add proper support for multiple query statements

dfellis:
It was discovered that TypeORM's `query` method breaks if there are multiple SQL statements in the query. They are run correctly, but no response is returned, as it must not be able to handle the response provided by the postgres driver. So this PR just uses the `pg` library directly and does some light formatting to make the response more closely match the behavior of the `query` function before, which is also generally more intuitive than the metadata-filled objects `pg` returns by default.

### DIFF
--- a/src/services/iasql.ts
+++ b/src/services/iasql.ts
@@ -3,6 +3,7 @@
 // meta operations within the module code itself, if desirable.
 import { exec as execNode, } from 'child_process'
 import { promisify, } from 'util'
+import pg from 'pg'
 const exec = promisify(execNode);
 
 import * as levenshtein from 'fastest-levenshtein'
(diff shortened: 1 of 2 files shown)


src/services/iasql.ts
```diff
@@ -179,16 +180,31 @@ export async function runSql(dbAlias: string, uid: string, sql: string) {
     connMain = await createConnection({ ...dbMan.baseConnConfig, database, name: pass, });
     await connMain.query(dbMan.newPostgresRoleQuery(user, pass, database));
     await connMain.query(dbMan.grantPostgresRoleQuery(user));
-    connTemp = await createConnection({ ...dbMan.baseConnConfig, database, name: user, });
+    connTemp = new pg.Client({
```
depombo, line 183: I think it is possible to get the bare pool/connection from `pg` from the createconnection like we do here https://github.com/iasql/iasql-engine/blob/main/src/services/scheduler.ts#L29.  It would be ideal so we consistently use the same config, but not blocking.

depombo (commented):
Do we want to add a unit test to make sure we don’t have regressions?

src/services/iasql.ts
dfellis: So this *has* to be a separate connection (see the old code being replaced) so I didn't think it was that big of a deal to cut out the middleman here, too.

dfellis:
> Do we want to add a unit test to make sure we don’t have regressions?

What did you have in mind here? One query of each "kind" (error, single response, multiple response)?

Also would you mind if it's in the `aws_account` integration test, instead? A unit test without a database in the loop would probably not actually exercise the part(s) of the logic that could actually break things, like a `pg` library update changing their own response.

src/services/iasql.ts
depombo: I understand it it is a separate connection. I meant consistency in configuring and getting a connection since we do it in a few places across the codebase. That was the purpose of the db manager at least? The TypeORM bug is unfortunate. Like this is totally fine though. Mostly a suggestion

depombo:
> > Do we want to add a unit test to make sure we don’t have regressions?
> 
> What did you have in mind here? One query of each "kind" (error, single response, multiple response)?

Yeah, particularly the first and the last

> 
> Also would you mind if it's in the `aws_account` integration test, instead? A unit test without a database in the loop would probably not actually exercise the part(s) of the logic that could actually break things, like a `pg` library update changing their own response.

That sounds great!

src/services/iasql.ts
dfellis: To be clear, I was also worried about Typeorm still returning a "tampered" object here, so I went with the clearest way to do this change. :)

depombo (approved):

PR #940, opened 2022-07-06
Fix the flaky ec2 test by reverting EBS change

dfellis:
Unfortunately, I believe the recent EBS PR is breaking the test suite


aguillenv (approved):

aguillenv:
I'm pretty confident [this PR](https://github.com/iasql/iasql-engine/pull/941)  stabilizes the test suite and theres no need to revert all of this, in any case it would need to be reverted only the multi region change for the ec2 test suite, nothing else.

PR #941, opened 2022-07-06
Stabilize ec2 multiregion tests

aguillenv:


package.json
```diff
@@ -8,7 +8,7 @@
     "clean": "rm -rf dist",
     "lint": "tslint --project .",
     "test": "IASQL_ENV=test jest -i",
-    "coverage": "IASQL_ENV=test jest -i --coverage --forceExit",
+    "coverage": "IASQL_ENV=test jest -i --coverage --forceExit --bail",
```
aguillenv, line 11: Added this to make the tests fail on the first error and do not wait until all the suite is done to see if something failed. I believed it would help us, but we can revert if we do not like it

test/modules/aws-ec2-integration.ts
```diff
@@ -4,6 +4,8 @@ import { getPrefix, runQuery, runInstall, runUninstall, runApply, finish, execCo
 
 const dbAlias = 'ec2test';
 const region = process.env.AWS_REGION;
+// t3.micro is availvaile in almost every az except one of us-east-1 region
+const instanceType = region === 'us-east-1' ? 't2.micro' : 't3.micro';
```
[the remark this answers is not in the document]
aguillenv: Yes, we could use the API, we could create a js client or a bash script. I have more or less the script but maybe it is easier to run the js client?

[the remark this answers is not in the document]
aguillenv: I'll go with the client

test/modules/aws-ec2-integration.ts
```diff
@@ -79,24 +110,28 @@ describe('EC2 Integration Testing', () => {
 
   it('inserts aws credentials', querySync(`
     INSERT INTO aws_account (region, access_key_id, secret_access_key)
-    VALUES ('us-east-1', '${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}')
+    VALUES ('us-east-1', '${accessKeyId}', '${secretAccessKey}')
   `));
 
+  test('two plus two is four', () => {
+    expect(2 + 2).toBe(5);
+  });
```
dfellis, line 118: uhhh

package.json
aguillenv: ohhh ok, so `--bail` work between suites, but here we want to stop tests within the same suite (https://github.com/facebook/jest/issues/2626). The solution for us should go in something like this https://thewebdev.info/2022/02/24/how-to-stop-test-suite-after-first-fail-with-jest/

I'll remove that option

dfellis (approved):

dfellis (approved):

test/modules/aws-ec2-integration.ts
```diff
@@ -1,9 +1,49 @@
...
+const getAvailabilityZones = async () => {
+  return await ec2client.describeAvailabilityZones({
+    Filters: [
+      {
+        Name: 'region-name',
+        Values: [region,],
+      },
+    ],
+  })
+}
```
depombo, line 27: not for this PR, but we might want to make this a table in `aws_account`?

test/modules/aws-ec2-integration.ts
```diff
@@ -1,9 +1,49 @@
...
+        Name: 'location',
+        Values: availabilityZones,
+      },
+      {
+        Name: 'instance-type',
+        Values: ['t2.micro', 't3.micro'],
+      }
+    ],
+  });
+}
```
depombo, line 43: out of the scope of this PR, but this makes me think it might make sense to introduce an instance type table with fk to the availability zone table in `aws_account` I mention in https://github.com/iasql/iasql-engine/pull/941/files#r915165321?

depombo (approved):

test/modules/aws-ec2-integration.ts
aguillenv: Hmmm instance type table should exist in aws ec2 module, I think. AvailabilityZone table not sure is aws_account or if it should be part of  vpc?

depombo: yes, an `instance type` table would be in ec2. I suggested `availability_zone` in `aws_account` seems you mentioned different accounts have different availability zones

PR #953, opened 2022-07-08
Turn the AvailabilityZone enum into a table

### LINKED ISSUE #942: Make Availability Zone a table instead of an `enum`
The availability zones available change based on the account.
image

dfellis:
Resolves #942


dfellis:
Screenshot from 2022-07-08 17-56-27

Somehow this PR has made it possible to create two Route53 hosted zones with the same name in our test suite which then breaks the assumption that the domain name is unique.

1. Apparently you can have two hosted zones with different IDs that target the same domain name, so perhaps we should remmove that unique constraint?
2. I am at a loss as to *how* this change has caused this.

dfellis:
I suspect something weird with the equality checking now that TypeORM turns the availability zone into an object with a single property, but I'm not sure where to correct this. Will return to this after a break.

src/modules/0.0.12/aws_ec2/index.ts
```diff
@@ -15,7 +15,6 @@ import * as metadata from './module.json'
 import { AwsElbModule } from '../aws_elb'
```
aguillenv, line 15: I think the equals method fro instance should be updated here

dfellis: Thank you for the hint :)

src/modules/0.0.12/aws_rds/entity/rds.ts
```diff
@@ -32,6 +32,7 @@ export class RDS {
   })
   allocatedStorage: number;
 
+  // TODO: Tie this to the AvailabilityZone entity in the future
```
depombo, line 35: we should probably create issues for this TODOs?

depombo (approved):

src/modules/0.0.12/aws_rds/entity/rds.ts
dfellis: Sounds good.

PR #954, opened 2022-07-09
initialize sentry within the scheduler child processes

### LINKED ISSUE #896: surface install errors other than the "Collision of tables…" to sentry

depombo:
Resolves #896 and surfaces errors to sentry for all iasql functions which were not getting surfaced.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,8 @@
     "gen-sql": "IASQL_ENV=ci ts-node ./node_modules/.bin/typeorm migration:generate -n",
     "migrate": "IASQL_ENV=ci ts-node ./node_modules/.bin/typeorm migration:run",
     "gen-module": "IASQL_ENV=ci ./.module_migration.sh",
+    "docker-compose:debug": "IASQL_ENV=ci docker-compose -f docker-compose.yml -f docker-compose.debug.yml up --build",
+    "docker-compose": "IASQL_ENV=local docker-compose up --build",
     "version": "./.version.sh"
   },
   "repository": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,11 +14,9 @@
     "serve:debug": "node --inspect=0.0.0.0:9229 --max_old_space_size=8192 dist/index.js",
     "start": "yarn serve",
     "start:debug": "nodemon -e ts -w ./src -x yarn serve:debug",
-    "gen-sql": "IASQL_ENV=local ts-node ./node_modules/.bin/typeorm migration:generate -n",
-    "migrate": "IASQL_ENV=local ts-node ./node_modules/.bin/typeorm migration:run",
-    "gen-module": "IASQL_ENV=local ./.module_migration.sh",
-    "docker-compose:debug": "IASQL_ENV=local docker-compose -f docker-compose.yml -f docker-compose.debug.yml up --build",
-    "docker-compose": "IASQL_ENV=local docker-compose up --build",
```
dfellis, line 21: Please restore these two.

depombo: I had talked with Alejandro about removing them last week to avoid dirtying metrics for local usage, but had never made the PR

dfellis: I use `yarn docker-compose` *all* the time. I'd prefer we fix the metrics issue rather than remove it, because I'd just run it, anyways, when testing locally?

depombo: It is not a metrics issue, we simply have to turn telemetry off by using `IASQL_ENV=ci docker-compose up —build` or a command that wraps it. The commands are restored anyway

dfellis (approved):

PR #957, opened 2022-07-11
Update CONTRIBUTING.md

murtaza-ghadiali:
Windows command added


depombo (approved):

dfellis:
Hey @murtaza-ghadiali it looks like our test suite currently doesn't work correctly for forked PRs. We'll need to look into that, but definitely don't want to gate this PR on it when it has nothing to do with the test suite, so I will merge it in now! :)

PR #966, opened 2022-07-12
Make the AvailabilityZone records quasi-immutable

dfellis:
This should hopefully resolve the E2E test failure we saw last night, but can only confirm after this is merged


dfellis:
Actually, just realized that I can manually deploy this to staging, so nevermind on "can't confirm before landing."

depombo (approved):

PR #968, opened 2022-07-13
pass `Error` to logErrSentry fn

depombo:
Fixes mislabeled sentry exceptions for dbs that fail to startup

### DIFF
--- a/src/services/logger.ts
+++ b/src/services/logger.ts
@@ -92,7 +92,7 @@ export function debugObj(e: any) {
 // returns the sentry error id
 export function logErrSentry(e: any, uid?: string, email?: string, dbAlias?: string): string {
   if (!(e instanceof DepError) && !(e instanceof Error)) {
-    singleton.error(`Invalid error type ${typeof e} when logging to sentry`)
+    singleton.error(`Invalid error type ${typeof e} when logging to sentry. e = ${JSON.stringify(e)}`)
   }
   let message = e?.message ?? '';
   let err = e;


dfellis (approved):
How did the type safety fail here?

depombo:
> How did the type safety fail here?

It uses `any` since it can be the error returned by the lazy loader or this hmmm

dfellis:
> > How did the type safety fail here?
> 
> It uses `any` since it can be the error returned by the lazy loader or this hmmm

Then I have an idea, just a sec.

src/services/scheduler.ts
```diff
@@ -173,7 +173,7 @@ export async function init() {
       const user = await MetadataRepo.getUserFromDbId(db.pgName);
       const message = `Failed to bootstrap db ${db.pgName} on startup. Reason: ${bootstrap.reason}`;
       logger.error(message);
-      logErrSentry(message, user?.id, user?.email, db.alias);
+      logErrSentry({message}, user?.id, user?.email, db.alias);
```
dfellis, line 176:
```suggestion
      logErrSentry(message instanceof Object ? message : { message, }, user?.id, user?.email, db.alias);
```

depombo: well this message is always a string, I can add some type safety in the `logErrSentry` signature

src/services/logger.ts
```diff
@@ -90,10 +91,13 @@ export function debugObj(e: any) {
 // TODO is there a way to DRY that?
 // returns the sentry error id
 export function logErrSentry(e: any, uid?: string, email?: string, dbAlias?: string): string {
+  if (!(e instanceof DepError) && !(e instanceof Error)) {
+    singleton.error(`Invalid error type ${typeof e} when logging to sentry`)
+  }
```
aguillenv, line 96: If we fail if is not `DepError` or `Error` why don't we:
```suggestion
export function logErrSentry(e: Error | DepError, uid?: string, email?: string, dbAlias?: string): string {
```

aguillenv: Or if we want to specifically caught this kind of errors we could show the content of the error to not lose information:
```suggestion
export function logErrSentry(e: any, uid?: string, email?: string, dbAlias?: string): string {
  if (!(e instanceof DepError) && !(e instanceof Error)) {
    singleton.error(`Invalid error type ${typeof e} when logging to sentry. e = ${JSON.stringify(e)}`)
  }
```

depombo: Catch statements are of unknown type unfortunately so we would have to cast it at the call site every time we call this function. Instead I am checking the type here.

aguillenv: oh thats true!! well maybe my second suggestion can give us some extra idea of what happened if we send the wrong thing at some point?

dfellis (approved):

PR #969, opened 2022-07-13
add readonly integration test with all modules

depombo:


test/readonly/test.ts
```diff
@@ -0,0 +1,69 @@
...
+  it('adds a new repository', query(`
+    INSERT INTO repository (repository_name, scan_on_push, image_tag_mutability)
+    VALUES ('test', false, 'MUTABLE');
+  `));
+
+  it('fails to apply', (done) => {
+    query(`
+      select * from iasql_apply();
+    `)((_e?: any) => done());  // Ignore failure
+  });
```
dfellis, line 62: This doesn't actually confirm that it failed, though?

.github/workflows/ci.yml
```diff
@@ -56,6 +56,29 @@ jobs:
...
+      - name: Use Node.js 16.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16.x
+
+      - name: Install dependencies
+        run: yarn
+
+      - name: Run common tests
+        run: yarn coverage **/common/*
```
aguillenv, line 74: this should be:
```suggestion
        run: yarn coverage **/readonly/*
```

aguillenv: another option is that the read only test could live inside the `common` directory, we remove this new ci job and update the common one adding the read only credentials?

.github/workflows/ci.yml
```diff
@@ -56,6 +56,29 @@ jobs:
...
+
+      - name: Use Node.js 16.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16.x
+
+      - name: Install dependencies
+        run: yarn
+
+      - name: Run common tests
```
aguillenv, line 73:
```suggestion
      - name: Run read-only tests
```

aguillenv (approved):

dfellis (approved):

PR #981, opened 2022-07-13
Auto-import module versions

dfellis:
Part of the work to make version bumping fully automated is for the code to automatically determine which versions should even be loaded.

### DIFF
--- a/src/modules/index.ts
+++ b/src/modules/index.ts
@@ -9,6 +9,6 @@ import config from '../config'
 // tslint:disable-next-line:no-var-requires
 export const latest = require(`./${config.modules.latestVersion}`);
 
-export const modules = fs.readdirSync(__dirname, 'utf8').filter(m => semver.valid(m)).map(v => [`v${v.replace(/\./g, '_')}`, require(v)]);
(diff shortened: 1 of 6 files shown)


src/modules/index.ts
```diff
@@ -1,13 +1,17 @@
...
-export * as v0_0_9 from './0.0.9'
-export * as v0_0_10 from './0.0.10'
-export * as v0_0_11 from './0.0.11'
-export * as v0_0_12 from './0.0.12'
-export * as v0_0_13 from './0.0.13'
+export const modules = Object.fromEntries(
+  fs.readdirSync(__dirname, 'utf8').filter(m => semver.valid(m)).map(v => [`v${v.replace(/\./g, '_')}`, require(`./${v}`)])
+);
+modules.latest = latest;
+
```
aguillenv, line 16: Do we want to add the oldest here? I do not remember if we use it somewhere

dfellis: I actually want to remove this `latest` bit and migrate everything to using the config-defined latest version, since technically this could do weird stuff in production.

PR #983, opened 2022-07-14
Schedule cleanup job

### LINKED ISSUE #975: Create CI job to wipe all AWS accounts and remove `execComposeDown` in tests

aguillenv:
Closes #975


.github/workflows/module-cleanup.yml
```diff
@@ -0,0 +1,94 @@
...
+        env:
+          AWS_REGION: ${{ matrix.region }}
+          AWS_ACCESS_KEY_ID: ${{ secrets[env.MATRIX_AWS_ACCESS_KEY_ID] }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets[env.MATRIX_AWS_SECRET_ACCESS_KEY] }}
+
+      - name: Run cleaning script
+        if: ${{ !inputs.global }}
+        run: |
+          echo "\nDelete all records..."
+          psql postgres://postgres:test@localhost:5432/iasql -f test/sql/delete_records.sql -c "select * from iasql_apply();"
```
dfellis, line 80: Minor question: Why not shove the `select * from iasql_apply()` into the `.sql` files?

test/sql/delete_records.sql
```diff
@@ -14,14 +14,7 @@ BEGIN
...
         raise notice 'logging table %', aux_tables_array[table_elem];
-        -- public_repository and role are special because they are region agnostic so we append a region to their names
-        IF aux_tables_array[table_elem] = 'public_repository' THEN
-          -- raise notice '%', format('DELETE FROM public_repository WHERE repository_name LIKE ''%s''', '%' || aws_region);
-          EXECUTE format('DELETE FROM public_repository WHERE repository_name LIKE ''%s''', '%' || aws_region);
-        ELSIF aux_tables_array[table_elem] = 'role' THEN
-          -- raise notice '%', format('DELETE FROM role WHERE role_name LIKE ''%s''', '%' || aws_region);
-          EXECUTE format('DELETE FROM role WHERE role_name LIKE ''%s''', '%' || aws_region);
-        ELSE
+        IF aux_tables_array[table_elem] != 'public_repository' AND aux_tables_array[table_elem] != 'role' THEN
```
dfellis, line 17: We should make a task to add some metadata in the `iasql_tables` table on if a table is global or not, so we can automatically filter between these two files.

dfellis (approved):

.github/workflows/module-cleanup.yml
dfellis: That works for me. :)

test/sql/delete_records.sql
depombo: 💯

.github/workflows/scheduled-cleanup.yml
```diff
@@ -0,0 +1,125 @@
...
+    outputs:
+      set-test-module-names: ${{ steps['set-test-module-names'].outputs['test-module-names'] }}
+    steps:
+      - uses: actions/checkout@v2
+      - run: yarn
+      - id: set-test-module-names
+        name: Set modules tests names
+        run: echo "::set-output name=test-module-names::$(npx jest **/modules/* --listTests --json | jq -c 'map(split("/") | .[-1] | split(".") | .[0] | gsub( "-"; "_") | ascii_upcase | tostring)')"
+
+  # Job by region since we pass the 256 job limit running them all at once
```
depombo, line 29: just a note that if AWS adds regions we need to update a few things manually in our codebase

test/sql/delete_records.sql
dfellis: We're OSS. I think a Dockerhub image updated with each release makes sense to me?

depombo: Would it be worth it / possible to have a docker image with dependencies installed that still uses the latest codebase to shave more time off ?

dfellis: Hmm... I don't know if dockerhub would allow that. I'm not even sure how you get an account and/or how many images are allowed for a free account (if they even do that anymore?)

I'm pretty sure each image needs a version tag, but it might just be a string, which would make it *possible* to do what you're suggesting. But again, need to read up on the details here.

However, I don't know how useful that would be, though. For CI we would want to build from scratch no matter what, and for this cleanup job, I don't think it would matter if the image is a few days old?

PR #986, opened 2022-07-14
Write new prepare and drop scripts and update the version script to require no arguments

dfellis:
Tested locally with the git commit and checkout commands commented out


aguillenv (approved):
Would be good to add the steps of how deploy process would look like in the README or somewhere once all the changes we want to do are in place.

dfellis:
> Would be good to add the steps of how deploy process would look like in the README or somewhere once all the changes we want to do are in place.

Agreed, though this is only partway there with the next PR being my intended last one. I'll address this comment in that PR.

depombo (approved):

PR #988, opened 2022-07-14
Do not clean on compose down to avoid interference between concurrent runs

aguillenv:


test/helpers.ts
```diff
@@ -99,47 +89,6 @@ export function runQuery(databaseName: string, queryString: string, assertFn?: (
...
-  });
-  logger.info(`Connection created...`);
-  await conn.query(`
-    INSERT INTO aws_account (region, access_key_id, secret_access_key)
-    VALUES ('${awsRegion}', '${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}')
-  `);
-  logger.info('AWS Account set up');
-  await iasql.install(modules, dbAlias, config.db.user);
-  logger.info(`Modules ${modules.join(', ')} installed...`);
-  const delQuery = fs.readFileSync(`${__dirname}/sql/delete_records.sql`, 'utf8');
```
depombo, line 127: this sql is used in tutorial integration tests. it is worth checking if it still makes sense to have it be different to the fn in `iasql_platform`?

depombo: I think so

depombo (approved):

PR #1001, opened 2022-07-18
Handle misconfigured target group

aguillenv:
Solving this issue https://sentry.io/organizations/iasql/issues/3433358736/?project=6106929&referrer=webhooks_plugin


dfellis (approved):
LGTM. We won't accidentally delete a target group that we think is misconfigured.

This could cause issues with the cleanup script, though, being unable to see these target groups to trigger a deletion of them.

aguillenv:
> LGTM. We won't accidentally delete a target group that we think is misconfigured.
> 
> This could cause issues with the cleanup script, though, being unable to see these target groups to trigger a deletion of them.

I guess if it is misconfigured is not safe to be managed, that would include deletion, but I see your point

Issue #963, opened 2022-07-11
Higher level ECS modules without custom module/mappers


depombo:
Some research is necessary to figure out if it makes more sense to use a `VIEW`s or perhaps `RULE`s. Whatever format we pick will set the stage for advanced IaSQL users. It is common within eng orgs using IaC to create higher level modules built on top of the ones provided e.g. the DevOps/infra person in a 10-20 person eng team creates a `backend`/`microservice` abstraction that the software engineers use to spin up their services. This higher level abstraction will for example use a k8 pod with a set number of instances, an RDS instance, load balancer, etc. The software engineer can overwrite the module, but mostly they take it off the shelf and provide a few input params. Ideally we can offer our users a way to do this without having to write a custom IaSQL module, but simply do it with SQL functionality like an updatable `VIEW`

aguillenv:
I was taking a look and found earlier this week an example of [updateable views](https://www.percona.com/blog/2019/01/10/postgresql-updatable-views-performing-schema-updates-with-minimal-downtime/)

could be tricky though

depombo:
I think updatable views do not work across multiple tables. Perhaps the way forward would be with two-way triggers or somehow with rules hmm

depombo:
Proof of concept of two-way triggers at work using a very simple example!

```sql
drop table if exists example;
drop table if exists example_simplified;

create table example_simplified(id integer UNIQUE NOT NULL, description text);
create table example(id integer UNIQUE NOT NULL, description text, serial integer);

create or replace function trigger_on_example_simplified()
returns trigger language plpgsql as $$
begin
    INSERT INTO example (id, description) VALUES (new.id, new.description)
    ON CONFLICT (id)
    WHERE id = new.id
    DO UPDATE SET description = new.description;
    RETURN new;
end
$$;

create trigger trigger_on_example_simplified
after insert or update on example_simplified
for each row when (pg_trigger_depth() = 0)
execute procedure trigger_on_example_simplified();

create or replace function trigger_on_example()
returns trigger language plpgsql as $$
begin
    INSERT INTO example_simplified (id, description) VALUES (new.id, new.description)
    ON CONFLICT (id)
    WHERE id = new.id
    DO UPDATE SET description = new.description;
    RETURN new;
end
$$;

create trigger trigger_on_example
after insert or update on example
for each row when (pg_trigger_depth() = 0)
execute procedure trigger_on_example();

insert into example_simplified (id, description)
values (1, 'John');

select *
from example;
[...]

PR #1003, opened 2022-07-18
EC2 RFC

### LINKED ISSUE #964: EC2 schema RFC

aguillenv:
Resolves #964 

Easy to read: https://github.com/iasql/iasql-engine/blob/ec2-rfc/rfcs/001%20-%20EC2%20schema.md


rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,110 @@
...
+- Launch templates
+- Spot instances
+- Dedicated hosts
+- Capacity reservations
+- Auto Scaling Groups
+- Elastic IP integration
+- Placement groups
+- Network interfaces
+
+No implicaiton in the `instance` table:
```
dfellis, line 71:
```suggestion
No implication in the `instance` table:
```

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,110 @@
...
+
+- Instance type - Availability zone relationship
+- AMIs management
+- EBS volume snapshots and life cycle
+
+The `Saving plans`, `Reserved instances` or `Scheduled instances` are not included in the list above since they seem to be budgeting solutions that need interaction with the user rather than infrastructure-related modules' features that could have any implication in the model.
```
dfellis, line 77: I don't agree with this. If a user knows the EC2 instance they are going to spin up will be permanently up for years at a time, we should let them specify that so they can pay a significantly lower price for it.

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,110 @@
...
+
+Let's start with the features without impact on the `instance` table. These features will be part of the `aws_ec2` module with their mappers. The Instance type table by availability zones will be a table with a composite primary key. It technically has a minimum impact in the `instance` table that will be a special FK to the composite primary key, but no UX model impact.
+
+The features with direct implications in the `instance` table can be divided into two groups. The first group will not only affect the `aws_ec2` module, and the second group will.
```
dfellis, line 81: This sentence reads awkwardly. Perhaps:

```suggestion
The features with direct implications in the `instance` table can be divided into two groups: one group affects only the `aws_ec2` module, the impacts more modules.
```

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,110 @@
...
+
+Launch templates, auto-scaling group requests and spot instances requests can be linked to an instance using the tags AWS assign with a specific pattern. Otherwise will be hard to link all the relationships due to the async nature of ASG and spot instances.
+
+### Alternatives Considered
+
+Create different modules for the services that can manage AWS instances.
+Create different tables within the same `aws_ec2` module for the services that can manage AWS instances.
+These two were discarded since they will increase the complexity of the model. Users will need to find all the tables that can contain `instance` information.
+
+Add iasql module's capabilities.
```
dfellis, line 97: I don't understand this option. What do you mean by this?

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,110 @@
...
+This option is not viable since it will be really hard to maintain and make it compatible through versions. It will imply the need for several opinionated decisions in behaviour that could not be expected by the users. Also, new users may not understand what capabilities are and add all of them every time.
+
+## Expected Semver Impact
+
+Patch version
+
+## Affected Components
+
+`aws_ec2` module
```
dfellis, line 106: I thought it said other modules, such as `aws_vpc` would also be impacted?

rfcs/001 - EC2 schema.md
aguillenv: Oh I can rephrase better but it is the idea I suggested on Wednesday about having the modules dynamically add or remove columns/tables based on "capabilities" selected by the user.

rfcs/001 - EC2 schema.md
aguillenv: definitely!

rfcs/001 - EC2 schema.md
aguillenv: Yes, I agree with what you are saying, but the way they work it is at cost/budget level and not resources. Let me give an example. When you purchase a Reserved instance you are not buying an instance for x period of time, you are buying the "spot" (similar when you reserved a spot in a parking lot) for an instance that have those specifications you are defining. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/apply_ri.html

dfellis: Got it, but I would say that could justifiably be a different table in that case, not that we should ignore it entirely.

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,113 @@
...
+- Scheduled instances
+- Capacity reservations
+- Auto Scaling Groups
+- AMIs management
+- EBS volume snapshots and life cycle
+- Elastic IP integration
+- Placement groups
+- Network interfaces
+
+Almost all the features listed will need a direct integration with the `instance` table, needing to add at least eight more columns to the model. Those columns are only for integration, but the `instance` model still needs more properties to be defined to be complete. Some of those missing properties are: instance auto-recovery, shutdown behaviour, stop - hibernate behaviour, termination protection, stop protection, detailed cloudWatch monitoring, elastic GPU, credit specification, tenancy, RAM disk ID, kernel ID or Metadata accessible (I might be missing others).
```
depombo, line 50: Wouldn’t some, or most, of the properties described at the end be part of the `instance_metadata` table instead of the `instance` table?

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,113 @@
...
+
+No implication in the `instance` table:
+
+- Instance type - Availability zone relationship
+- AMIs management
+- EBS volume snapshots and life cycle
+
+The `Saving plans`, `Reserved instances` or `Scheduled instances` are not included in the list above since they seem to be budgeting solutions that need interaction with the user rather than infrastructure-related modules' features that could have any implication in the model.
+
+Let's start with the features without impact on the `instance` table. These features will be part of the `aws_ec2` module with their mappers. The Instance type table by availability zones will be a table with a composite primary key. It technically has a minimum impact in the `instance` table that will be a special FK to the composite primary key, but no UX model impact.
```
depombo, line 79: how would eventually going multi-region affect this part of the proposal? would it affect other parts of this proposal?

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,113 @@
...
+
+The features with direct implications in the `instance` table can be divided into two groups: one group affects only the `aws_ec2` module, the other impacts more modules.
+
+Let's start with the first group, Placement groups and Network interfaces. These features will be part of their respective modules (probably `aws_vpc`?). Then, they will need to be integrated with the `instance` table in the `aws_ec2` module with an FK nullable column.
+
+The rest of the features will need to be implemented in the `aws_ec2` module and integrated with the `instance` table in the `aws_ec2` module with an FK nullable column.
```
depombo, line 85: so currently `registered_instances` has the FK outside the `instance` table. would it be better to keep the UX of things consistent or do we have a strong reason to keep that different to these tables? do we anticipate any other tables like `registered_instances`?

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,113 @@
...
+
+Spot instances and Auto Scaling Group can be updated at any point in time by the cloud. These will imply the development of the "safer sync faster apply" feature as a pre-requisite for the two.
+
+Launch templates, auto-scaling group requests and spot instances requests can be linked to an instance using the tags AWS assign with a specific pattern. Otherwise will be hard to link all the relationships due to the async nature of ASG and spot instances.
```
depombo, line 89: I know for spot instance requests, instances come with a nullable spot request id field. Would that not be preferable to a tag? maybe it should live in `instance_metadata` since it is technically not mutable?

rfcs/001 - EC2 schema.md
aguillenv: I don't think so since all of them are possible inputs for instance creation

rfcs/001 - EC2 schema.md
aguillenv: Hmm good point. It will depend on how we implement that. For instance, assuming it will be a `region` column in all tables, this will work, we just need to add the new column to the composite key and thats all. At first glance I do not see other parts of this proposal affected by the multi-region change.

rfcs/001 - EC2 schema.md
aguillenv: Well I think that specific case could not be modeled otherwise, is a `many-to-many` relationship, where an instance can be registered with multiple target groups and target groups can have multiple ec2 instances registered.

The network interfaces would be a similar case to ebs volumes, where an instance can have several ENIs but one ENI is only assigned to an instance, so the FK to the instance will be in the network interface table.

rfcs/001 - EC2 schema.md
aguillenv: Ok so, some  points here. 

1. I will rephrase better, but what I'm saying is that from the tags that AWS assign, is how we can know if an instance belongs to an ASG or a spot fleet or was launched based on a launch template. Based on this tag we can look and assign the respective FK.

2. The field you are saying for spot instance request id it is a tricky one and I think we should not trust it. When you ask for a spot fleet request, for example, it is created a spot  instance request that creates the instance, but that spot instance request (that is the one that comes in the instance field as spot request id) is then deleted from the spot requests table, because it belongs to the fleet, it was not a single request. What I want to say is that sometimes those ids can be ephemeral.

3. From this comment I've being thinking and you are completely right that some fields, specifically the FKs to spot requests and ASG should belong to `instance_metadata` and not to `instance` table.

I will take a look again and maybe draw some diagrams showing the relationships.

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,136 @@
...
+- Placement groups
+- Key Pairs
+- Network interfaces
+
+The `instance` table still needs more advanced options to be defined. Some of those missing properties are: instance auto-recovery, shutdown behaviour, stop - hibernate behaviour, termination protection, stop protection, detailed cloudWatch monitoring, elastic GPU, credit specification, tenancy, RAM disk ID, kernel ID or Metadata accessible (I might be missing others). Additionally, some of the features listed above will need to be added to the `instance` table as FKs.
+
+The current EC2 model looks like this:
+
+![Current EC2 model](./assets/001/current_schema.png)
+Note: Could be also accessed via https://dbdocs.io/alejandro/iasql
```
depombo, line 58: I'm curious, is this different from https://dbdocs.io/iasql/iasql?table=instance&schema=public&view=table_structure ?

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,136 @@
...
+
+We will review the needed features, see their implications in the model and get a final version of the ECS schema.
+
+- Creating an instance using the UI/API lets you add some advanced configuration that usually is not used but is there. The screenshot below shows part of the extra configuration. At least the following columns will need to be added to the `instance` table: instance auto-recovery, shutdown behaviour, stop - hibernate behaviour, termination protection, stop protection, detailed cloudWatch monitoring, elastic GPU, credit specification, tenancy, RAM disk ID, kernel ID or Metadata accessible.
+
+![EC2 instance advance properties](./assets/001/ec2_advance_properties.png)
+
+- Instance types depend on availability zones. We need to create a table relating them using a composite key. We also need to add the `availability_zone` column to the `instance` table to relate with the new `instance_type` table using the composite key.
```
depombo, line 70: Once/if this is accepted, we should create a task and tackle this one soon-ish?

depombo (approved):

rfcs/001 - EC2 schema.md
depombo: https://github.com/iasql/iasql-engine/issues/963 is prioritized this week which we can also use here for a `instance_simplified` type of table to provide a simpler UX that still allows peeking down the abstraction layer easily if/when the user needs to

aguillenv: > I'm curious, is this different from https://dbdocs.io/iasql/iasql?table=instance&schema=public&view=table_structure ?

It is basically the same. I just picked that model and did some trim leaving just ec2 and tables directly related to it

rfcs/001 - EC2 schema.md
aguillenv: yes, this would be easy to do

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,136 @@
...
+The `instance` table still needs more advanced options to be defined. Some of those missing properties are: instance auto-recovery, shutdown behaviour, stop - hibernate behaviour, termination protection, stop protection, detailed cloudWatch monitoring, elastic GPU, credit specification, tenancy, RAM disk ID, kernel ID or Metadata accessible (I might be missing others). Additionally, some of the features listed above will need to be added to the `instance` table as FKs.
+
+The current EC2 model looks like this:
+
+![Current EC2 model](./assets/001/current_schema.png)
+Note: Could be also accessed via https://dbdocs.io/alejandro/iasql
+
+## Proposal
+
+After considering alternatives like splitting everything into more modules or having different `instance` tables around based on what service creates them, the proposal is to keep the schema growing in this same module. The relationships, though, will be added where they belong and rely on the instance metadata for the no needed ones for the CRUD of the `instance` table.
```
dfellis, line 62: Nit:

```suggestion
After considering alternatives like splitting everything into more modules or having different `instance` tables around based on what service creates them, the proposal is to keep the schema growing in this same module. The relationships, though, will be added where they belong and rely on the instance metadata for the optional fields that are unnecessary for the CRUD of the `instance` table.
```

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,136 @@
...
+
+![EC2 instance advance properties](./assets/001/ec2_advance_properties.png)
+
+- Instance types depend on availability zones. We need to create a table relating them using a composite key. We also need to add the `availability_zone` column to the `instance` table to relate with the new `instance_type` table using the composite key.
+
+- Launch templates will be a new table and need to be an `instance` table input column `launch_template_id`. If we insert other values to the `instance`, these will override the launch template ones. To know if an instance was created using one of them we check the tags and look for the id to link the FK.
+
+- Spot requests can be created manually, using launch templates or can be created by other entities like ASG or the same instance. If an instance was created using a spot request should not affect the `instance` table, so the relationship between them will be part of the `instance_metadata`. To be able to know if an instance was created by a spot request we check the tags and look for the id to link them.
```
dfellis, line 74: The phrasing here is a bit ambiguous. Does "If an instance was created using a spot request should not affect the `instance` table mean:

1. That spot instances *don't* show up in the `instance` table and therefore do not affect it.
2. Spot instances *do* show up in the `instance` table, but do not affect the schema of the `instance` table.

dfellis: BTW, I am assuming it's (2) here, but I first read it as (1).

dfellis (approved):
Approved assuming my reading of one sentence is correct.

rfcs/001 - EC2 schema.md
aguillenv: Yes, it is the second. They do show up but if they were created or not by a spot request will not affect the schema of the instance table in terms of CRUD operations. Maybe this rephrase explain it better?

```suggestion
- Spot requests can be created manually, using launch templates or can be created by other entities like ASG or the same instance. The relationship between the `instance` and the spot request will be part of the `instance_metadata` since the spot request does not affect the `instance` schema in terms of CRUD operations. To be able to know if an instance was created by a spot request we check the tags and look for the id to link them.
```

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,136 @@
...
+The current EC2 model looks like this:
+
+![Current EC2 model](./assets/001/current_schema.png)
+Note: Could be also accessed via https://dbdocs.io/alejandro/iasql
+
+## Proposal
+
+After considering alternatives like splitting everything into more modules or having different `instance` tables around based on what service creates them, the proposal is to keep the schema growing in this same module. The relationships, though, will be added where they belong and rely on the instance metadata for the optional fields that are unnecessary for the CRUD of the `instance` table.
+
+We will review the needed features, see their implications in the model and get a final version of the ECS schema.
```
yrobla, line 64: should be EC2 not ECS ?

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,136 @@
...
+
+![Proposed schema](./assets/001/proposed_schema.png)
+
+Note: This schema does not have all the possible relationships, but the most important ones with the `instance` table.
+Note 2: Could be also accessed via https://dbdocs.io/alejandro/proposed-iasql?view=relationships
+
+Some of the features exposed do apply changes automatically to the cloud, like spot instances and Auto Scaling Groups. These will imply the development of the "safer sync faster apply" feature as a pre-requisite.
+
+we can start with instance type, launch templates, dedicated hosts, capacity reservation, ebs snapshots, placement groups, key pairs and network interfaces.
+
```
yrobla, line 112: I have one comment about instanceType. The types of instances are defined, and right now it is just an string. We could create a table and only allow to insert the defined values there.

rfcs/001 - EC2 schema.md
dfellis: That rewrite clears it up!

rfcs/001 - EC2 schema.md
dfellis: Yep, that's right. Should be

```suggestion
We will review the needed features, see their implications in the model and get a final version of the EC2 schema.
```

rfcs/001 - EC2 schema.md
dfellis: I definitely think that's a good idea.

It could also be done via the `aws_ec2_metadata` module to attach that safety onto the ec2 module. So it's still a string, but we attach a foreign key relationship to an `instance_type` table if the metadata module is installed so users can opt-in on the improved type safety.

aguillenv: Yes, and also instance types depend on availability zones. This is explained in one of the first points of this proposal (https://github.com/iasql/iasql-engine/pull/1003/files#diff-35d35a757122d48b71d944384346f61d9c234056b348aa48f7072db57cdbd715R70).

This will be a table where the source of truth will always be the cloud. We have handled these cases by just deleting new records if a user creates one, and on updates/deletes just restore the records from the cloud.

yrobla (approved):

rfcs/001 - EC2 schema.md
```diff
@@ -0,0 +1,136 @@
...
+## Current Status
+
+### Proposed
+
+2022-07-18
+
+### Accepted
+
+YYYY-MM-DD
```
aguillenv, line 11:
```suggestion
2022-08-04
```

PR #1010, opened 2022-07-18
emit trigger only when changing src/ test/

depombo:


.github/workflows/main.yml
```diff
@@ -3,6 +3,7 @@ name: MAIN
 on:
   push:
     branches: [main]
+    paths: ['src/**', 'test/**']
```
dfellis, line 6: Can we add the `package.json` and `yarn.lock` files,  too?

dfellis (approved):

Issue #722, opened 2022-05-02
VPC improvements


aguillenv:
Revive the work in [this PR](https://github.com/iasql/iasql-engine/pull/661)

PR #1011, opened 2022-07-19
Issue #722: Wait for vpc to be created

yrobla:
Modify vpc creation to wait until state is available.
Also modify method equality to check for cidr, state
and default.

Closes-Issue: #722


aguillenv (approved):
Looking good! I'd suggest add the necessary changes to update path and some testing. Feel free to do it in follow up PRs

depombo (approved):

PR #1012, opened 2022-07-19
Fix the response output when executing insert/update/delete sql commands

dfellis:
The engine provided a useless empty set when running non-SELECT commands before. The new logic detects this scenario and returns the affected record count in the same form as a query result, instead, to be more useful to end-users.

This resolves iasql/dashboard#147
This also resolves iasql/dashboard#171


aguillenv (approved):

depombo:
I think this also fixes https://github.com/iasql/dashboard/issues/171 ? 2 for 1!

PR #1015, opened 2022-07-19
ECS + Subnets

aguillenv:
Part of #1005

### DIFF
--- a/src/modules/0.0.14/aws_ecs_fargate/sql/create_custom_constraints.sql
+++ b/src/modules/0.0.14/aws_ecs_fargate/sql/create_custom_constraints.sql
@@ -1,5 +1,5 @@
 -- Create service subnets constraint
-create or replace function check_service_subnets(_subnets integer[]) returns boolean
+create or replace function check_service_subnets(_subnets text[]) returns boolean
 language plpgsql security definer
 as $$
 declare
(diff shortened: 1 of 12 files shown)


src/modules/0.0.14/aws_ecs_fargate/sql/create_custom_constraints.sql
```diff
@@ -12,3 +12,29 @@ begin
...
+        service_row record;
+    BEGIN
+        FOR service_row IN
+            SELECT * FROM service
+        LOOP
+            IF OLD.subnet_id = any(service_row.subnets) THEN
+                RAISE EXCEPTION 'subnet_id % is being used by service %', OLD.subnet_id, service_row.name;
+            END IF;
+        END LOOP;
+        IF (TG_OP = 'DELETE') THEN
```
dfellis, line 27: Am I reading this correctly that deletes are basically banned at the database level here?

aguillenv: not sure what do you mean?  I'm executing this function on a `BEFORE DELETE OR UPDATE` trigger, the only difference is that on `DELETE` the trigger needs to return the `OLD` record and for `UPDATE`s need to return the `NEW` record

aguillenv: Oh no no, it is blocking `UPDATE`s and `DELETE`s if the subnet id in question is present in any of the service `subnets` column

dfellis (approved):

depombo (approved):

PR #1018, opened 2022-07-20
Switch the load balancer availability zone list to a join table

### LINKED ISSUE #1004: Add AZ joins

dfellis:
Resolves #1004

There are zero tests involving availability zones on load balancers and it appears to be a read-only listing from AWS in the code here.


aguillenv:
> Resolves #1004
> 
> There are zero tests involving availability zones on load balancers and it appears to be a read-only listing from AWS in the code here.

Yes, this one was quite weird, uses the subnets to create the loadbalancer but then just return the availability zones, thats why it is just read

dfellis:
Closing this out for a different approach.

PR #1021, opened 2022-07-20
Issue #1019: Add support for tags at vpc module

### LINKED ISSUE #1019: Add support for tags at vpc module
Include addition and update of tags

yrobla:
Include adding and updated tags

Resolves #1019

### DIFF
--- a/src/modules/0.0.14/aws_vpc/index.ts
+++ b/src/modules/0.0.14/aws_vpc/index.ts
@@ -45,6 +45,7 @@ import {
 } from './entity'
 import { Context, Crud2, Mapper2, Module2, } from '../../interfaces'
 import * as metadata from './module.json'
+import isEqual from 'lodash.isequal'
 
 const createSubnet = crudBuilder2<EC2, 'createSubnet'>('createSubnet', (input) => input);
 const getSubnet = crudBuilderFormat<EC2, 'describeSubnets', AwsSubnet | undefined>(
@@ -58,39 +59,39 @@ const deleteSubnet = crudBuilder2<EC2, 'deleteSubnet'>('deleteSubnet', (input) =
 async function createVpc(client: EC2, input: CreateVpcCommandInput) {
   const res = await client.createVpc(input);
   const describeInput: DescribeVpcsCommandInput = {
-    VpcIds: [res.Vpc?.VpcId ?? '']
+    VpcIds: [res.Vpc?.VpcId ?? ''],
   };
   let out;
   await createWaiter<EC2, DescribeVpcsCommandInput>(
-    {
-      client,
-      // all in seconds
-      maxWaitTime: 300,
-      minDelay: 1,
-      maxDelay: 4,
-    },
-    describeInput,
-    async (cl, cmd) => {
-      const data = await cl.describeVpcs(cmd);
-      try {
-        out = data.Vpcs?.pop();
-        // If it is not a final state we retry
-        if ([VpcState.PENDING].includes(out?.State as VpcState)) {
-          return { state: WaiterState.RETRY };
+      {
+        client,
+        // all in seconds
+        maxWaitTime: 300,
+        minDelay: 1,
+        maxDelay: 4,
+      },
+      describeInput,
+      async (cl, cmd) => {
+        const data = await cl.describeVpcs(cmd);
+        try {
+          out = data.Vpcs?.pop();
+          // If it is not a final state we retry
+          if ([VpcState.PENDING].includes(out?.State as VpcState)) {
+            return {state: WaiterState.RETRY};
+          }
+          return {state: WaiterState.SUCCESS};
+        } catch (e: any) {
+          throw e;
         }
-        return { state: WaiterState.SUCCESS };
-      } catch (e: any) {
-        throw e;
-      }
-    },
+      },
   );
   return out;
 }
 
 const getVpc = crudBuilderFormat<EC2, 'describeVpcs', AwsVpc | undefined>(
-  'describeVpcs',
-  (id) => ({ VpcIds: [id], }),
-  (res) => res?.Vpcs?.[0],
+    'describeVpcs',
+    (id) => ({VpcIds: [id],}),
+    (res) => res?.Vpcs?.[0],
 );
 const getVpcs = paginateBuilder<EC2>(paginateDescribeVpcs, 'Vpcs');
 const deleteVpc = crudBuilder2<EC2, 'deleteVpc'>('deleteVpc', (input) => input);
@@ -292,11 +293,7 @@ async function deleteNatGateway(client: EC2, id: string) {
 async function updateTags(client: EC2, resourceId: string, tags?: { [key: string] : string }) {
   let tgs: Tag[] = [];
   if (tags) {
-    tgs = Object.keys(tags).map(k => {
-      return {
-        Key: k, Value: tags[k]
-      }
-    });
+    tgs = Object.entries(tags).map(([Key, Value]) => ({ Key, Value}));
   }
   // recreate tags
   await client.deleteTags({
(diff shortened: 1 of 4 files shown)


src/modules/0.0.14/aws_vpc/index.ts
```diff
@@ -543,7 +567,11 @@ export const AwsVpcModule: Module2 = new Module2({
               await AwsVpcModule.mappers.vpc.db.update(cloudRecord, ctx);
               out.push(cloudRecord);
             } else {
-              await AwsVpcModule.mappers.vpc.cloud.create(e, ctx);
+              const res = await AwsVpcModule.mappers.vpc.cloud.create(e, ctx);
+              if (!AwsVpcModule.utils.eqTags(e.tags, cloudRecord.tags) && e.vpcId && e.tags) {
+                await updateTags(client.ec2client, res.vpcId, e.tags);
+              }
+
```
aguillenv, line 574: oh wait you should not be doing this here, this is the replace path. Also, please add the `updateOrReplace` mapper function to identify the cases (we forgot in the previous PRs). Then, the tags' update and the current restore should go through the `update` path.

src/modules/0.0.14/aws_vpc/index.ts
```diff
@@ -359,6 +359,12 @@ export const AwsVpcModule: Module2 = new Module2({
       out.cidrBlock = vpc.CidrBlock;
       out.state = vpc.State as VpcState;
       out.isDefault = vpc.IsDefault ?? false;
+      const tags: { [key: string]: string } = {};
+      (vpc.Tags || []).filter(t => !!t.Key && !!t.Value).forEach(t => {
```
dfellis, line 363: Nit. This filter would technically filter out a key or value that is an empty string. Probably fine for the key, but I could see someone legitimately using just the presence of a key in a HashSet-style and not care about the value, and this would drop it.

Perhaps this would be better:

```suggestion
      (vpc.Tags || []).filter(t => t.hasOwnProperty('Key') && t.hasOwnProperty('Value')).forEach(t => {
```

src/modules/0.0.14/aws_vpc/index.ts
```diff
@@ -536,15 +564,24 @@ export const AwsVpcModule: Module2 = new Module2({
           const client = await ctx.getAwsClient() as AWS;
           const out = [];
           for (const e of es) {
-            const cloudRecord = ctx?.memo?.cloud?.Vpc?.[e.vpcId ?? ''];
+            const cloudRecord = ctx?.memo?.cloud?.Vpc?.[e.vpcId ?? ''];
```
aguillenv, line 567:
```suggestion
            const cloudRecord = ctx?.memo?.cloud?.Vpc?.[e.vpcId ?? ''];
```

src/modules/0.0.14/aws_vpc/index.ts
```diff
@@ -495,19 +501,41 @@ export const AwsVpcModule: Module2 = new Module2({
...
       equals: (a: Vpc, b: Vpc) => {
-        const result = Object.is(a.cidrBlock, b.cidrBlock) && Object.is(a.state, b.state) && Object.is(a.isDefault, b.isDefault);
+        const result = Object.is(a.cidrBlock, b.cidrBlock) && Object.is(a.state, b.state) && Object.is(a.isDefault, b.isDefault) &&
+          (AwsVpcModule.utils.eqTags(a.tags, b.tags));
         return result;
       },
       source: 'db',
       cloud: new Crud2({
+        updateOrReplace: (a: Vpc, b: Vpc) => {
+          if (Object.is(a.cidrBlock, b.cidrBlock)) return "replace";
```
aguillenv, line 511: Hmm I think here it is the opposite, right? 
```suggestion
          if (!Object.is(a.cidrBlock, b.cidrBlock)) return "replace";
```

aguillenv: Please also include the `isDefault` property for the `replace` path

src/modules/0.0.14/aws_vpc/index.ts
```diff
@@ -419,8 +425,13 @@ export const AwsVpcModule: Module2 = new Module2({
       if (serviceName.includes('s3')) return EndpointGatewayService.S3;
       if (serviceName.includes('dynamodb')) return EndpointGatewayService.DYNAMODB;
     },
-    eqTags: (a: { [key: string]: string }, b: { [key: string]: string }) => Object.is(Object.keys(a ?? {})?.length, Object.keys(b ?? {})?.length) &&
-      Object.keys(a ?? {})?.every(ak => (a ?? {})[ak] === (b ?? {})[ak]),
+    eqTags: (a: { [key: string]: string }, b: { [key: string]: string }) => {
+      // remove possible `updated` key
+      delete a.updated;
+      delete b.updated;
```
aguillenv, line 431: oh why are you doing this?

dfellis: Agree here. This shouldn't be necessary.

src/modules/0.0.14/aws_vpc/index.ts
```diff
@@ -495,20 +506,45 @@ export const AwsVpcModule: Module2 = new Module2({
...
         create: async (es: Vpc[], ctx: Context) => {
           // TODO: Add support for creating default VPCs (only one is allowed, also add constraint
           // that a single VPC is set as default)
           const client = await ctx.getAwsClient() as AWS;
           const out = [];
           for (const e of es) {
+            let tgs: Tag[] = [];
+            if (e.tags !== undefined && e.tags !== null) {
+              const tags: {[key: string]: string} = e.tags;
+              tags.owner = 'iasql-engine';
```
dfellis, line 528: Why is this here?

src/modules/0.0.14/aws_vpc/index.ts
```diff
@@ -495,20 +506,45 @@ export const AwsVpcModule: Module2 = new Module2({
...
+              tgs = Object.keys(tags).map(k => {
+                return {
+                  Key: k, Value: tags[k]
+                }
+              });
```
dfellis, line 533: Could also be

```suggestion
              tgs = Object.entries(tags).map(([Key, Value]) => ({ Key, Value, }));
```

src/modules/0.0.14/aws_vpc/index.ts
```diff
@@ -543,8 +579,17 @@ export const AwsVpcModule: Module2 = new Module2({
...
             } else {
-              await AwsVpcModule.mappers.vpc.cloud.create(e, ctx);
-              out.push(e);
+              const isUpdate = Object.is(AwsVpcModule.mappers.vpc.cloud.updateOrReplace(cloudRecord, e), 'update');
+              if (!isUpdate) {
+                // if CIDR is different we do the create
+                const newVpc = await AwsVpcModule.mappers.vpc.cloud.create(e, ctx);
+                out.push(newVpc);
+              } else {
+                if (!AwsVpcModule.utils.eqTags(e.tags, cloudRecord.tags) && e.vpcId && e.tags) {
```
dfellis, line 588: I would remove this last `&& e.tags` it should still update the tags if the user has deleted all of them.

```suggestion
                if (!AwsVpcModule.utils.eqTags(e.tags, cloudRecord.tags) && e.vpcId) {
```

test/modules/aws-vpc-integration.ts
```diff
@@ -101,6 +105,25 @@ describe('VPC Integration Testing', () => {
     AND state='pending';
   `, (res: any) => expect(res.length).toBe(0)));
 
+  it('tries to update vpc tags', query(`
+  UPDATE vpc SET tags = '{"name": "${prefix}-2", "updated": true}' WHERE cidr_block='192.${randIPBlock}.0.0/16';
```
dfellis, line 109:
```suggestion
  UPDATE vpc SET tags = '{"name": "${prefix}-2"}' WHERE cidr_block='192.${randIPBlock}.0.0/16';
```

dfellis (approved):

aguillenv (approved):

PR #1023, opened 2022-07-21
Add a check function to validate the load balancer's availability zones list

### LINKED ISSUE #1004: Add AZ joins

dfellis:
Resolves #1004

### DIFF
--- a/src/modules/0.0.14/aws_elb/sql/create_custom_constraints.sql
+++ b/src/modules/0.0.14/aws_elb/sql/create_custom_constraints.sql
@@ -4,7 +4,7 @@ as $$
 declare
   _number_of_records integer;
 begin
-  select count(*) from availability_zone where name = any(_availability_zones) into _number_of_records
+  select count(*) from availability_zone where name = any(_availability_zones) into _number_of_records;
   if _number_of_records != array_length(_availability_zones) then
     raise exception 'Load balancer % includes one or more invalid availability zones', _load_balancer_name;
   end if;
--- a/src/modules/0.0.14/aws_elb/sql/create_custom_constraints.sql
+++ b/src/modules/0.0.14/aws_elb/sql/create_custom_constraints.sql
@@ -5,7 +5,7 @@ declare
   _number_of_records integer;
 begin
   select count(*) from availability_zone where name = any(_availability_zones) into _number_of_records;
-  if _number_of_records != array_length(_availability_zones) then
+  if _number_of_records != array_length(_availability_zones, 1) then
     raise exception 'Load balancer % includes one or more invalid availability zones', _load_balancer_name;
   end if;
   return true;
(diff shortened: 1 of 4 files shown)


src/modules/0.0.14/aws_elb/sql/create_custom_constraints.sql
```diff
@@ -0,0 +1,14 @@
+create or replace function check_load_balancer_availability_zones(_load_balancer_name varchar, _availability_zones varchar[]) returns boolean
+language plpgsql security definer
+as $$
+declare
+  _number_of_records integer;
+begin
+  select count(*) from availability_zone where name = any(_availability_zones) into _number_of_records;
```
aguillenv, line 7: oh interesting this works, i've always done select x into y from.... not the whole query and then put that in the variable.

src/modules/0.0.14/aws_elb/sql/create_custom_constraints.sql
```diff
@@ -0,0 +1,14 @@
...
+  if _number_of_records != array_length(_availability_zones, 1) then
+    raise exception 'Load balancer % includes one or more invalid availability zones', _load_balancer_name;
+  end if;
+  return true;
+end;
+$$;
+ALTER TABLE load_balancer ADD CONSTRAINT check_load_balancer_availability_zones CHECK (check_load_balancer_availability_zones(load_balancer_name, availability_zones));
```
aguillenv, line 14: since this check will be called on insert, would not be possible that theres no `load_balancer_name`, would that constraint fail first that this one?

dfellis: It shouldn't because the `any(NULL)` should return no records to count and the length of it is also 0.

src/modules/0.0.14/aws_elb/sql/create_custom_constraints.sql
dfellis: Well, the whole query here is a singular value because of the `count`

dfellis: Oh, I see what you mean now wrt to the syntax. The whole `into` syntax is a bit weird, though.

src/modules/0.0.14/aws_elb/sql/create_custom_constraints.sql
aguillenv: Oh I was meaning since the load balancer name is being passed as an argument, but it is true that if nor there is `NULL` so all good

aguillenv (approved):

src/modules/0.0.14/aws_elb/sql/create_custom_constraints.sql
```diff
@@ -0,0 +1,14 @@
+create or replace function check_load_balancer_availability_zones(_load_balancer_name varchar, _availability_zones varchar[]) returns boolean
```
yrobla, line 1: can you add some comments on the file to explain the purpose of it? for me as a newcomer is still difficult to understand the purpose. This one is pretty straightforward but i am sure that there will be other functions that will not

PR #1024, opened 2022-07-21
`aws_lambda` basics

### LINKED ISSUE #638: Create an `aws_lambda` module

aguillenv:
Resolves #638


src/modules/0.0.14/aws_lambda/entity/lambda_function.ts
```diff
@@ -0,0 +1,135 @@
...
+  provided = "provided",
+  providedal2 = "provided.al2",
+  python27 = "python2.7",
+  python36 = "python3.6",
+  python37 = "python3.7",
+  python38 = "python3.8",
+  python39 = "python3.9",
+  ruby25 = "ruby2.5",
+  ruby27 = "ruby2.7"
+}
```
dfellis, line 53: This changes relatively frequently. Can we query it into a table instead?

dfellis: I'm also surprised at how far behind Java is on this list. The newest here is Java 11, but [Java 18](https://en.wikipedia.org/wiki/Java_version_history) has been out for ~3 months.

src/modules/0.0.14/aws_lambda/entity/lambda_function.ts
```diff
@@ -0,0 +1,135 @@
...
+  @Column({
+    default: '$LATEST',
+  })
+  version?: string;
+
+  @Column({ nullable: true, })
+  description?: string;
+
+  @Column({ nullable: true, })
+  zipB64?: string;
```
dfellis, line 73: Is this always present in the database, or is it an upload-only flag? If the latter, can we make a task to add a boolean, defaulted to false, to let users decide if they want to keep the payload in the DB?

src/modules/0.0.14/aws_lambda/index.ts
```diff
@@ -0,0 +1,204 @@
...
+        create: async (es: LambdaFunction[], ctx: Context) => {
+          const client = await ctx.getAwsClient() as AWS;
+          const out = [];
+          for (const e of es) {
+            const input: CreateFunctionCommandInput = {
+              FunctionName: e.name,
+              Description: e.description,
+              Code: {
+                ZipFile: base64ToUint8Array(e.zipB64 ?? ''),
+              },
```
dfellis, line 103: We should probably skip and/or drop the creation of a lambda function without the base64-encoded zip file defined.

Also, add a TODO here to validate that the b64 contents are actually a zip file?

src/modules/0.0.14/aws_lambda/index.ts
```diff
@@ -0,0 +1,204 @@
...
+import { LambdaFunction } from './entity'
+import { AwsIamModule } from '../aws_iam'
+
+const base64ToUint8Array = (base64: string) => {
+  const buf = Buffer.from(base64, 'base64');
+  const bytes = new Uint8Array(buf);
+  return bytes;
+}
+
+const lambdaFunctionMapper = async (fn: GetFunctionResponse, ctx: Context) => {
```
dfellis, line 34: I don't see `zipB64` in this function. Is that on purpose?

src/modules/0.0.14/aws_lambda/entity/lambda_function.ts
aguillenv: this is an enum itself in aws sdk, I can check if theres an api call that we can do to get this info

src/modules/0.0.14/aws_lambda/entity/lambda_function.ts
aguillenv: it is upload only because on `read` we cannot retrieve this data back.

src/modules/0.0.14/aws_lambda/index.ts
aguillenv: yes, because we do not receive the b64 from the cloud

src/modules/0.0.14/aws_lambda/entity/lambda_function.ts
aguillenv: I've done a quick search in the API and couldn't find an endpoint to retrieve this information. Either we can maintain this enum or directly use the one provided by AWS SDK? I did not use the last one to be consistent with other modules.

src/modules/0.0.14/aws_lambda/index.ts
aguillenv: The options I see here on create are either `throw` if no zip, or `db.delete` the row. I'd go with the `throw` option at least for now to let users know what's wrong.

aguillenv: Regarding the TODO suggestion 👍🏼

src/modules/0.0.14/aws_iam/index.ts
```diff
@@ -180,6 +180,10 @@ export const AwsIamModule: Module2 = new Module2({
             const newRole = await AwsIamModule.utils.roleMapper(rawRole, ctx);
             await AwsIamModule.mappers.role.db.update(newRole, ctx);
             out.push(newRole);
+            // TODO: find a way to fix this better?
+            // wait 10 seconds to let aws get the role ready to use :/
+            // https://stackoverflow.com/a/37438525
+            await new Promise(r => setTimeout(r, 10000));
```
dfellis, line 186: :shrug:

test/modules/aws-lambda-integration.ts
```diff
@@ -0,0 +1,254 @@
...
+  //   WHERE role_name = '${lambdaFunctionRoleName}';
+  // `, (res: any[]) => expect(res.length).toBe(0)));
+
+  // it('applies the lambda function role removal', apply());
+
+  // it('check lambda function role does not exist', query(`
+  //   SELECT *
+  //   FROM role 
+  //   WHERE role_name = '${lambdaFunctionRoleName}';
+  // `, (res: any[]) => expect(res.length).toBe(0)));
```
dfellis, line 210: Why are these commented out?

src/modules/0.0.14/aws_iam/index.ts
aguillenv: horrible I know. Not sure how to fix it properly, it also increases exponentially IAM tests run time. I'll try to set a lower number and see what happens :/

test/modules/aws-lambda-integration.ts
aguillenv: it should not!!

src/modules/0.0.14/aws_iam/index.ts
dfellis: Is there a waiter to wait for the IAM role to be ready to use?

dfellis: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-iam/globals.html#waitforroleexists Maybe this would work?

aguillenv: The issue is that the role in question seems to be defined, but the error is `The role defined for the function cannot be assumed by Lambda`. The assume role part is done by passing the `AssumeRolePolicyDocument` property creating the role. I've seen that waiter and another for policies, I'll try to run both of thems and see if that works.

aguillenv: Nope, the waiter does not work and I've found this https://stackoverflow.com/questions/20156043/how-long-should-i-wait-after-applying-an-aws-iam-policy-before-it-is-valid. Apparently there's no way no know this :/

dfellis: Alternative solution. In the Lambda module, this is a "known" possible error, and just have an internal try-catch-sleep loop for up to say, 10 iterations with 5 second sleeps?

aguillenv: oh definitely better and do not impact IAM module

dfellis (approved):

depombo (approved):

PR #1025, opened 2022-07-21
002 - Custom Module Functions RFC

dfellis:


rfcs/002 - Custom Module Functions RFC.md
```diff
@@ -0,0 +1,80 @@
...
+
+This has always been a sore point architecturally as encapsulation of information is broken, and it is not truly possible to adjust the behavior of the functions in question without affecting prior module versions at the same time. It is also very difficult (though not impossible) to introduce new functions because 4 different files across the repository must be updated in sync to do so correctly, and can only be done in a way that would be considered a SemVer minor update, major is completely not allowed.
+
+It is proposed to deprecate/drop the `iasql_operation` entity and associated enum, and create a new graphile worker job type, called `rpc` and a new entity named `iasql_rpc` with no enum. Instead two fields would be added as an "addressing system" of module name and method name, which the module would implement within an `rpc` sub-object where all properties are the method names and the values are async functions that take a `Context` object as the first argument and then any arguments from the SQL side are added as follow-up parameters, all as strings similar to the current `params` array.
```
yrobla, line 40: Due to my background i have a very strong opinion about microservices. One approach could be dividing this single iasql engine inside smaller pieces, that have a very specific focus. The worker could be just one specific microservice, that is able to talk with others to requests specific actions. This framework adds better isolation and would be easier to develop and maintain, but would need some orchestration layer for communication (typically kubernetes)

dfellis: I also prefer microservices but hate Kubernetes. ;)

I don't think we're yet at the point where splitting things up like that makes sense, though.

1. There's still a tight coupling between the engine and the modules, with the interface between them still being in flux as we try to improve things there (as you yourself have seen the abstraction is not quite there yet for module development). Breaking that out will complicate iteration speed when we can't really afford that at the moment.
2. The "rpc" here is a remote procedure call in the classic sense, but it is between the Postgres database and the IaSQL engine as an RPC between two monoliths, essentially. The reason for this is that Postgres *is* our interface for end-users and the engine is an implementation detail that they don't need to pay much attention to, so as much of the API as possible is pushed into the database. This was chosen for a variety of reasons that should also become a technical blogpost ;) but the biggest one is that the fewer demands you place on the client to use your product, the more places they will be able to. You can theoretically do all of your cloud management with just a SQL client like ArcType this way, if you wanted to.
3. The operational simplicity of the engine as it currently stands is a huge bonus. Your own onboarding went smoother, I think, than if we had a hand-written runbook on how to set up a local microk8-or-whatever cluster and get the microservices running locally. (I dislike Kubernetes because it has tried to be all things to all people and is not opinionated enough to actually be any easier to manage than a classic nginx reverse proxy + app pool of hosts + db write main & read replicas that the vast majority of backend companies really ought to be using. If K8s came with a solid set of defaults and said "swap these out only if you know what you're doing" it would be a much better product.)

Microservices are great at 1) more efficiently allocating compute resources and 2) enforcing strict interface boundaries. We do not yet have an issue with (1) and we have not fully figured out what (2) should be, so I just think it's premature to move in that direction.

rfcs/002 - Custom Module Functions RFC.md
```diff
@@ -0,0 +1,80 @@
...
+
+This would also allow us to move the majority of the `src/services/iasql.ts` functions into the `iasql_functions` module, allowing behavior changes to happen between version bumps of IaSQL and properly segmenting scope again. We would also [tackle this task](https://github.com/iasql/iasql-engine/issues/1016) on the new `iasql_rpc` entity instead of the `iasql_operation` entity.
+
+On a similar note, the current hand-editing of the migration files to insert custom pure-SQL functions is error prone when the migration is regenerated on entity changes. I propose that we have a similar `sql` sub-type on the modules similar to the old `migration` type that automatically runs whatever custom SQL was given to it after the migration up and before the migration down, to prevent this, treating it sort of like a second migration step, but more ergonomic than having to hand write a TypeORM migration file.
```
aguillenv, line 46: Yes, this is necessary. It would be something like
```ts
sql: {
  afterMigrationUp,
  beforeMigrationDown
}
```

rfcs/002 - Custom Module Functions RFC.md
aguillenv, line 40: So on module installation the `rpc` object will be read and make all the necessary inserts into the `iasql_rpc` table automatically for us?

dfellis: So there shouldn't be any need to insert anything into the `iasql_rpc` table. What I meant was that it would auto-generate plpgsql functions to be added after all migration/sql commands have been run on install and similar delete commands on uninstall.

These would use a generic plpgsql function kinda like [`iasql_cloud_manipulation`](https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.14/iasql_functions/sql/create_fns.sql#L75-L96) that is given metadata about the RPC function to call and returns the output in a generalized way that can then be consumed in a mostly-automatic way on the other side.

rfcs/002 - Custom Module Functions RFC.md
dfellis: Yep :) Though I would probably use terminology like `afterInstall` and `beforeUninstall`

rfcs/002 - Custom Module Functions RFC.md
aguillenv: so the `iasql_rpc` table would still have an "operation" id and the output and error columns?

aguillenv: I think what I'm not seeing is when the fn similar to `iasql_cloud_manipulation` will be called

dfellis: When the user invokes the RPC function in Postgres it would internally call the function that actually implements the job logic and wait for the response, while the job on the engine side would use the metadata provided to it to actually invoke the function implemented by the module.

aguillenv: I think I follow it now!!!

aguillenv (approved):

rfcs/002 - Custom Module Functions RFC.md
yrobla: Ok we have quite different point of view about Kubernetes :) But makes sense at least for the moment while there is no huge need to scale and distribute load

depombo (approved):
This will be great to have and I can't think of a better way to address the problems posed.

yrobla (approved):

rfcs/002 - Custom Module Functions RFC.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Current Status
+
+### Proposed
+
+2022-07-21
+
+### Accepted
+
+YYYY-MM-DD
```
dfellis, line 11:
```suggestion
2022-08-15
```

PR #1041, opened 2022-07-26
Fix the drop and develop scripts

dfellis:
We realized that the drop and develop scripts weren't updating the config files, so this fixes that.

### DIFF
--- a/.develop.sh
+++ b/.develop.sh
@@ -13,6 +13,16 @@ VERSION=$1
 # The version to clone
 LATESTVERSION=`ts-node src/scripts/latestVersion.ts`
 
+# The oldest version still supported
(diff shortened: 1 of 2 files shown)


.develop.sh
```diff
@@ -13,6 +13,16 @@ VERSION=$1
 # The version to clone
 LATESTVERSION=`ts-node src/scripts/latestVersion.ts`
 
+# The oldest version still supported
+LASTVERSION=`ts-node src/scripts/oldestVersion.ts`
```
aguillenv, line 17: In a future pass over these scripts, we might want to standardize the name to `oldestVersion`. I think is not only here but in some other places.

aguillenv (approved):

PR #1048, opened 2022-07-28
Add elasticache coverage

yrobla:
Start by adding cache_cluster entity with a set of minimum fields


src/modules/0.0.15/aws_elasticache/entity/cache_cluster.ts
```diff
@@ -0,0 +1,52 @@
...
+      enum: Engine,
+      default: Engine.REDIS
+    })
+    engine: Engine;
+
+    @Column()
+    numNodes?: number;
+
+    @Column()
+    port?: number;
```
aguillenv, line 50: if these fields are `nullable` please add the option in the column decorator

src/modules/0.0.15/aws_elasticache/entity/cache_cluster.ts
```diff
@@ -0,0 +1,52 @@
...
+  
+    @Column({
+      nullable: true,
+      comment: 'Unique identifier provided by AWS once the instance is provisioned',
+    })
+    @cloudId
+    clusterId?: string;
+    
+    @Column()
+    nodeType: string;
```
aguillenv, line 31: should this be an `enum`?

aguillenv (commented):
I haven't checked the API but those fields make sense to me. Just  a couple of comments

src/modules/0.0.15/aws_elasticache/entity/cache_cluster.ts
```diff
@@ -1,43 +1,42 @@
...
+    type: "enum",
+    enum: Engine,
+    default: Engine.REDIS,
+  })
+  engine: Engine;
+
+  @Column({
+    nullable: true,
+  })
+  numNodes?: number;
```
aguillenv, line 43: if this one can be nullable, remember adding the fn to convert from null to undefined like this one https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.15/aws_ec2/entity/instance.ts#L99-L107

src/modules/0.0.15/aws_elasticache/index.ts
```diff
@@ -0,0 +1,275 @@
...
+                if (cluster.engine !== cloudRecord.engine) {
+                  cluster.engine = cloudRecord.engine;
+                  await AwsElastiCacheModule.mappers.cacheCluster.db.update(
+                    cluster,
+                    ctx
+                  );
+                  out.push(cluster);
+                } else {
+                  // if the cluster ID is different, no needed to delete it
+                  if (cluster.clusterId===cloudRecord.clusterId) {
```
aguillenv, line 219: just as a comment, this check is not necessary, it should never come through the update path since it will be detected as create/delete in the main apply path.

aguillenv (approved):

src/modules/0.0.15/aws_elasticache/index.ts
yrobla: mm from my understanding i thought it will fall under update, because equals compares node type, num nodes and engine. But actually i'm thinking that for this case, equals needs to be based on clusterId directly... i can have multiple clusters with same engine, nodes ... and it doesn't mean that is the same. What do you think?

dfellis: `equals` determines if two things that are identified as the same thing are actually equal or not. `@cloudId` (usually) is what is used to decide what things should be compared and have CRUD operations potentially performed on them.

aguillenv: So the properties you are comparing in the `mapper.equals` method are the ones that will trigger any `mapper.update` call if there's any difference between objects with the same id. In this case, `clusterId` is the cloudId, meaning is what we use to identify cloud resources as unique. When an apply occurs it first matches based on cloudIds in order to see what is only defined in AWS (then need to be deleted), what is only in the DB (then it needs to be created) and for objects with the same cloudId call the `mappers.equals`, see if they have differences and then apply the `mapper.update` if necessary.

dfellis (approved):

aguillenv (approved):

PR #1054, opened 2022-08-01
Handle invalid elastic ips on nat gateways

### LINKED ISSUE #1038: NAT gateway with invalid EIP is failing

aguillenv:
Resolves #1038

### DIFF
--- a/src/modules/0.0.15/aws_vpc/index.ts
+++ b/src/modules/0.0.15/aws_vpc/index.ts
@@ -646,6 +646,7 @@ export const AwsVpcModule: Module2 = new Module2({
             }
             if (e.elasticIp) {
               input.AllocationId = e.elasticIp.allocationId;
+              if (!input.AllocationId) throw new Error('Elastic ip need to be created first');
             } else if (!e.elasticIp && e.connectivityType === ConnectivityType.PUBLIC) {
               const elasticIp = new ElasticIp();
               // Attach the same tags in case we want to associate them visualy through the AWS Console
--- a/src/modules/0.0.15/aws_vpc/index.ts
+++ b/src/modules/0.0.15/aws_vpc/index.ts
@@ -364,9 +364,13 @@ export const AwsVpcModule: Module2 = new Module2({
       const out = new NatGateway();
       out.connectivityType = nat.ConnectivityType as ConnectivityType;
       const natPublicAddress = nat.NatGatewayAddresses?.filter(n => !!n.AllocationId).pop();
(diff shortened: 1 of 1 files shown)


src/modules/0.0.15/aws_vpc/index.ts
```diff
@@ -364,9 +364,13 @@ export const AwsVpcModule: Module2 = new Module2({
...
       const natPublicAddress = nat.NatGatewayAddresses?.filter(n => !!n.AllocationId).pop();
-      if (natPublicAddress) {
-        out.elasticIp = await AwsVpcModule.mappers.elasticIp.db.read(ctx, natPublicAddress.AllocationId) ??
-          await AwsVpcModule.mappers.elasticIp.cloud.read(ctx, natPublicAddress.AllocationId);
+      if (natPublicAddress?.AllocationId) {
+        try {
+          out.elasticIp = await AwsVpcModule.mappers.elasticIp.db.read(ctx, natPublicAddress.AllocationId) ??
+            await AwsVpcModule.mappers.elasticIp.cloud.read(ctx, natPublicAddress.AllocationId);
+        } catch (error: any) {
+          if (error.Code === 'InvalidAllocationID.NotFound') return undefined;
```
dfellis, line 372: This just makes the Nat Gateway go away if the IP is invalid. Is that what we want?

aguillenv: Yes, right? That is the misconfiguration, any other case could be a transient error or another issue and we might fail.

yrobla: why is that code repeated twice? don't get it...

yrobla (approved):

src/modules/0.0.15/aws_vpc/index.ts
dfellis: But if you simultaneously create a new elastic IP and a nat gateway in new records in the database, wouldn't it reasonably return this error that we should retry the nat gateway construction on in that case?

aguillenv: Hmmm no, it should not be a problem, because if both are created at the same time, the NAT Gateway will fail if the ID is not there yet (committing that change explicitly after this comment). Here, mapping from the cloud, we make sure that if the cloudRecord has an `allocationId`, that id has to exist, otherwise is a misconfiguration.

dfellis (approved):

depombo (approved):

PR #1061, opened 2022-08-02
Class-Based Modules, Part 1

dfellis:
Part of #862 

The first step in switching to a class-based module system. Proposed new base class and implementation in the AwsAccount module. This conversion *should* work in a piecemeal basis as I made sure the new `ModuleBase` base class implements the `ModuleInterface` interface, but we'll let the test suite decide that. ;)

This is called "Part 1" not just because it's an incomplete conversion, but because there may be further changes/mutations to the pattern shown in this PR. The elimination of `utils` and `mappers` (sorta on the latter) flattens things and should improve Typescript's ability to infer type info, but if there are other [Mapper Footguns](https://github.com/iasql/docs/blob/bd7d9b7ac9af1fc4db3066131c56c61a1f8dd34b/docs/reference/apply_and_sync.md) that I can eliminate with module definition changes, then I hope to get them done this week, too.

### DIFF
--- a/src/modules/0.0.16/aws_account/index.ts
+++ b/src/modules/0.0.16/aws_account/index.ts
@@ -1,50 +1,48 @@
 import { AWS, } from '../../../services/aws_macros'
 import { AwsAccountEntity, } from './entity'
-import { Context, Crud2, Mapper2, Module2, } from '../../interfaces'
-import * as metadata from './module.json'
+import { Context, Crud2, Mapper2, ModuleBase, } from '../../interfaces'
+import * as metadata from './module.json' // TODO: Eliminate this?
 
-export const AwsAccount: Module2 = new Module2({
-  ...metadata,
-  provides: {
-    context: {
-      // This function is `async function () {` instead of `async () => {` because that enables the
-      // `this` keyword within the function based on the objec it is being called from, so the
-      // `getAwsClient` function can access the correct `orm` object with the appropriate creds and
-      // read out the right AWS creds and create an AWS client also attached to the current context,
-      // which will be different for different users. WARNING: Explicitly trying to access via
-      // `AwsAccount.provides.context.getAwsClient` would instead use the context *template* that is
-      // global to the codebase.
-      async getAwsClient() {
-        const orm = this.orm;
-        if (this.awsClient) return this.awsClient;
-        const awsCreds = await orm.findOne(AwsAccount.mappers.awsAccount.entity);
-        this.awsClient = new AWS({
-          region: awsCreds.region,
-          credentials: {
-            accessKeyId: awsCreds.accessKeyId,
-            secretAccessKey: awsCreds.secretAccessKey,
-          },
-        });
-        return this.awsClient;
-      },
-      awsClient: null, // Just reserving this name to guard against collisions between modules.
+class AwsAccount extends ModuleBase {
+  __dirname = __dirname;
+  dependencies = metadata.dependencies;
+  context: Context = {
+    // This function is `async function () {` instead of `async () => {` because that enables the
+    // `this` keyword within the function based on the objec it is being called from, so the
+    // `getAwsClient` function can access the correct `orm` object with the appropriate creds and
+    // read out the right AWS creds and create an AWS client also attached to the current context,
+    // which will be different for different users. WARNING: Explicitly trying to access via
+    // `AwsAccount.provides.context.getAwsClient` would instead use the context *template* that is
+    // global to the codebase.
+    async getAwsClient() {
+      const orm = this.orm;
+      if (this.awsClient) return this.awsClient;
+      const awsCreds = await orm.findOne(awsAccount.awsAccount.entity);
+      this.awsClient = new AWS({
+        region: awsCreds.region,
+        credentials: {
+          accessKeyId: awsCreds.accessKeyId,
+          secretAccessKey: awsCreds.secretAccessKey,
+        },
+      });
+      return this.awsClient;
     },
-  },
-  mappers: {
-    awsAccount: new Mapper2<AwsAccountEntity>({
-      entity: AwsAccountEntity,
-      equals: (_a: AwsAccountEntity, _b: AwsAccountEntity) => true,
-      source: 'db',
-      cloud: new Crud2({
-        create: async (_e: AwsAccountEntity[], _ctx: Context) => { /* Do nothing */ },
-        read: (ctx: Context, id?: string) => ctx.orm.find(AwsAccountEntity, id ? {
-          where: {
-            id,
-          },
-        } : undefined),
-        update: async (_e: AwsAccountEntity[], _ctx: Context) => { /* Do nothing */ },
-        delete: async (_e: AwsAccountEntity[], _ctx: Context) => { /* Do nothing */ },
-      }),
+    awsClient: null, // Just reserving this name to guard against collisions between modules.
+  };
+  awsAccount = new Mapper2<AwsAccountEntity>({
+    entity: AwsAccountEntity,
+    equals: (_a: AwsAccountEntity, _b: AwsAccountEntity) => true,
+    source: 'db',
+    cloud: new Crud2({
+      create: async (_e: AwsAccountEntity[], _ctx: Context) => { /* Do nothing */ },
+      read: (ctx: Context, id?: string) => ctx.orm.find(AwsAccountEntity, id ? {
+        where: {
+          id,
+        },
+      } : undefined),
+      update: async (_e: AwsAccountEntity[], _ctx: Context) => { /* Do nothing */ },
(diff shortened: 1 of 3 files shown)


src/modules/0.0.16/aws_account/index.ts
```diff
@@ -1,50 +1,49 @@
...
-          where: {
-            id,
-          },
-        } : undefined),
-        update: async (_e: AwsAccountEntity[], _ctx: Context) => { /* Do nothing */ },
-        delete: async (_e: AwsAccountEntity[], _ctx: Context) => { /* Do nothing */ },
-      }),
+    awsClient: null, // Just reserving this name to guard against collisions between modules.
+  };
+  awsAccount = new Mapper2<AwsAccountEntity>({
```
depombo, line 33: we can probably get rid of the original `Mapper`?

dfellis: Haven't we already? The issue is that these all need to keep the changed name (or we have to edit every prior LOC in the module system referencing it).

src/modules/0.0.16/aws_account/index.ts
```diff
@@ -1,50 +1,49 @@
...
+    entity: AwsAccountEntity,
+    equals: (_a: AwsAccountEntity, _b: AwsAccountEntity) => true,
+    source: 'db',
```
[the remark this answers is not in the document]
dfellis: We used to, and I think it might make sense in the `metadata` modules to do it, but actually until "automatic mode" it's not great because restoring changes made incorrectly by the user is delayed until the next time they `sync`, which may be never.

src/modules/0.0.16/aws_account/index.ts
```diff
@@ -1,50 +1,49 @@
...
+class AwsAccount extends ModuleBase {
+  constructor() { super(); super.init(); }
+  dirname = __dirname;
+  dependencies = metadata.dependencies;
+  context: Context = {
+    // This function is `async function () {` instead of `async () => {` because that enables the
+    // `this` keyword within the function based on the objec it is being called from, so the
+    // `getAwsClient` function can access the correct `orm` object with the appropriate creds and
+    // read out the right AWS creds and create an AWS client also attached to the current context,
+    // which will be different for different users. WARNING: Explicitly trying to access via
```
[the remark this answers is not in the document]
dfellis: Do we need to? No one is ever going to access that way. This is a warning I copy-pasted during the refactor, but it's clear to access the context via the `ctx` variable passed in?

src/modules/0.0.16/aws_account/index.ts
dfellis: Screenshot from 2022-08-02 13-46-16

Confirmed that only `Mapper2` remains.

src/modules/0.0.16/aws_account/index.ts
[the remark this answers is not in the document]
depombo: got it, mostly a suggestion to have to replace the comment

src/modules/0.0.16/aws_account/index.ts
depombo: got it

depombo (approved):

src/modules/interfaces.ts
```diff
@@ -372,3 +372,95 @@ export class Module2 {
...
+  name: string;
+  version: string;
+  dependencies: string[];
+  provides: {
+    entities: { [key: string]: any, };
+    tables: string[];
+    functions: string[];
+    context?: Context;
+  };
+  context?: Context;
```
aguillenv, line 387: What is the difference between these 2 context?

src/modules/interfaces.ts
```diff
@@ -372,3 +372,95 @@ export class Module2 {
...
+    if (files.length !== 1) throw new Error('Cannot determine which file is the migration');
+    const migration = require(`${migrationDir}/${files[0]}`);
+    // Assuming TypeORM migration files
+    const migrationClass = migration[Object.keys(migration)[0]];
+    if (!migrationClass || !migrationClass.prototype.up || !migrationClass.prototype.down) {
+      throw new Error('Presumed migration file is not a TypeORM migration');
+    }
+    this.migrations = {
+      install: migrationClass.prototype.up,
+      remove: migrationClass.prototype.down,
```
aguillenv, line 432: the implementation of `afterInstall` and `beforeRemove` will occur after the RFC is approved, right? is there any special casing that might need to be taken into account here?

src/modules/0.0.16/aws_account/index.ts
```diff
@@ -1,50 +1,49 @@
...
+    // `AwsAccount.provides.context.getAwsClient` would instead use the context *template* that is
+    // global to the codebase.
+    async getAwsClient() {
+      const orm = this.orm;
+      if (this.awsClient) return this.awsClient;
```
aguillenv, line 20: nit:
```suggestion
      if (this.awsClient) return this.awsClient;
      const orm = this.orm;
```

src/modules/0.0.16/aws_ecs_simplified/index.ts
```diff
@@ -48,7 +48,7 @@ const prefix = 'iasql-ecs-';
 
 // TODO: remove this once the aws gateway gets disolved
 async function getAwsClient(ctx: Context): Promise<AWS> {
-  const awsCreds = await ctx?.orm?.findOne(AwsAccount.mappers.awsAccount.entity);
+  const awsCreds = await ctx?.orm?.findOne(awsAccount.awsAccount.entity);
```
aguillenv, line 51: Maybe it should be renamed to `awsAccountModule` to have something like `awsAccountModule.awsAccount.entity`

src/modules/interfaces.ts
dfellis: The nested one in the `provides` object is for `src/services/iasql.ts` and maintains compatibility with the `ModuleInterface`, while the outer one is the one intended to be used by developers of modules, as all of the other `provides` fields are now auto-generated.

It simply copies the outer on to the inner one. Once the old modules are gone, we can drop the inner one without issue, but that will be in about a month.

src/modules/interfaces.ts
dfellis: It should be a backwards-compatible addition to this base class, so I am not super worried?

src/modules/0.0.16/aws_account/index.ts
```diff
@@ -1,50 +1,49 @@
...
-            secretAccessKey: awsCreds.secretAccessKey,
-          },
-        });
-        return this.awsClient;
-      },
-      awsClient: null, // Just reserving this name to guard against collisions between modules.
+class AwsAccount extends ModuleBase {
+  constructor() { super(); super.init(); }
+  dirname = __dirname;
+  dependencies = metadata.dependencies;
```
aguillenv, line 9: with this we could technically remove the metadata.json or are we still using it somewhere else?

src/modules/0.0.16/aws_account/index.ts
```diff
@@ -1,50 +1,49 @@
...
-            accessKeyId: awsCreds.accessKeyId,
-            secretAccessKey: awsCreds.secretAccessKey,
-          },
-        });
-        return this.awsClient;
-      },
-      awsClient: null, // Just reserving this name to guard against collisions between modules.
+class AwsAccount extends ModuleBase {
+  constructor() { super(); super.init(); }
+  dirname = __dirname;
```
aguillenv, line 8: do you foresee a case where `dirname` is not `__dirname`, might be weird to add this in all module's definitions?

aguillenv (approved):

src/modules/0.0.16/aws_account/index.ts
dfellis: So we still use it, along with the `dependencies` field, in the migration generation logic in the `src/scripts` directory, and I can't think of a good way to eliminate it, because we explicitly *don't* want these module `index.ts` files to be loaded since they try to auto-load the migration files that may or may not exist and then blow up in weird ways.

src/modules/0.0.16/aws_ecs_simplified/index.ts
dfellis: For most it's not a huge deal because it's things like `AwsEc2.instance...` and only modules that have a single mapper is this an issue. I can make that change through renaming during import?

src/modules/0.0.16/aws_account/index.ts
dfellis: It is already in all modules, just at the bottom instead of the top (see the end of the diff on this file). Unfortunately, because `__dirname` is a magic variable injected by Node.js itself, I can't hide this from the developers. If I use `__dirname` in the `ModuleBase` base class, it'll always be `src/modules/` because it's inside of the `src/modules/interfaces.ts` file.

PR #1062, opened 2022-08-03
Potential graphql hotfix

dfellis:


src/router/graphql.ts
```diff
@@ -55,6 +63,12 @@ if (config.graphql) {
...
       (await metadata.getDb(dbMan.getUid(req.user), req.params.db)).pgName :
       req.params.db;
-    getPostgraphile(db)(req, res, next);
+    try {
+      getPostgraphile(db)(req, res, next);
+    } catch (e) {
+      // This path is if postgraphile successfully spun up but somehow crapped out anyway
+      logger.error('Postgraphile failure', { e, });
+      res.status(400).end(`${db} does not exist`);
+    }
```
dfellis, line 72: This path is not hit, afaict, *but* it still seems valuable to have in case we ever do get into this situation.

src/router/graphql.ts
```diff
@@ -36,14 +37,21 @@ if (config.graphql) {
             externalUrlBase: `/v1/graphql/${db}`,
             graphiql: withGraphiql,
             watchPg: true,
+            retryOnInitFail: (_error, _attempts) => {
+              postgraphiles[db].graphile.shutdown();
+              return false;
+            },
```
dfellis, line 43: This part is the actual fix. If you don't define `retryOnInitFailure` it simply shuts down the server, which is insane!

src/services/iasql.ts
```diff
@@ -748,6 +748,7 @@ export async function modules(all: boolean, installed: boolean, dbId: string) {
   const allModules = Object.values(Modules)
     .filter((m: any) => m.hasOwnProperty('mappers') && m.hasOwnProperty('name') && !/iasql_.*/.test(m.name))
     .filter((m: any) => process.env.IASQL_ENV !== 'production' || !/aws_ecs_simplified.*/.test(m.name)) // Temporarily disable ecs_simplified in production
+    .filter((m: any) => process.env.IASQL_ENV !== 'production' || !/aws_dynamo.*/.test(m.name)) // Temporarily disable aws_dynamo in production
```
aguillenv, line 751: would this give trouble if someone already has the module installed?

src/router/graphql.ts
aguillenv: what is doing when reaching the problem? just killed the worker? Does it get in a weird state? is it returning some kind of error?

dfellis: The init failure is about the schema generation failing. This function returns `false` meaning "don't continue retrying the initialization process and it calls `shutdown` so the graphile process just always returns query failures. Apparently it returns a `503`. Hopefully that doesn't raise the ire of Fargate.

Screenshot from 2022-08-03 11-06-27

aguillenv: should we test it on staging?

aguillenv (approved):

PR #1064, opened 2022-08-03
Issue #1037: Add module for secrets

yrobla:
It offers the possibility of creating secrets, modifying
secret description and secret value, and removing the
secret

### DIFF
--- a/src/modules/0.0.16/aws_secret/entity/secret.ts
+++ b/src/modules/0.0.16/aws_secret/entity/secret.ts
@@ -27,7 +27,7 @@ export class Secret {
     type: String,
     nullable: true,
   })
-  value!: string | null;
+  value?: string | null;
 
   @AfterLoad()
   @AfterInsert()
--- a/src/modules/0.0.16/aws_secret/entity/secret.ts
+++ b/src/modules/0.0.16/aws_secret/entity/secret.ts
@@ -35,7 +35,7 @@ export class Secret {
   updateNulls() {
     const that: any = this;
     Object.keys(this).forEach((k) => {
-      if (k != "value") {
+      if (k !== "value") {
         if (that[k] === null) that[k] = undefined;
       }
     });
--- a/src/modules/0.0.16/aws_secret/entity/secret.ts
+++ b/src/modules/0.0.16/aws_secret/entity/secret.ts
@@ -1,0 +1,43 @@
+import {
+  AfterInsert,
+  AfterLoad,
+  AfterUpdate,
+  Column,
+  Entity,
+  PrimaryColumn,
+} from "typeorm";
+
+import { cloudId } from "../../../../services/cloud-id";
+
+@Entity()
+export class Secret {
+  @PrimaryColumn({
+    nullable: false,
+    type: "varchar",
+  })
+  @cloudId
+  name: string;
+
+  @Column({
+    nullable: true,
+  })
+  description?: string;
+
+  @Column({
+    type: String,
+    nullable: true,
+  })
+  value!: string | null;
+
+  @AfterLoad()
+  @AfterInsert()
+  @AfterUpdate()
+  updateNulls() {
+    const that: any = this;
+    Object.keys(this).forEach((k) => {
+      if (k != "value") {
+        if (that[k] === null) that[k] = undefined;
+      }
+    });
+  }
+}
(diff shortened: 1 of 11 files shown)


src/modules/0.0.16/aws_secret/index.ts
```diff
@@ -68,7 +69,7 @@ async function getSecret(client: SecretsManager, secretId: string) {
   return undefined;
 }
 
-const getAllSecrets = paginateBuilder<SecretsManager>(
+const getAllSecrets = paginateBuilder<SecretsManager | SecretsManagerClient>(
```
dfellis, line 72: Wait, how would this work? If the `paginateListSecrets` isn't on the `SecretsManager` object, won't it fail when you pass that client to it?

yrobla: it was a leftover sorry. Tests passed because it's not being called with that client

src/modules/0.0.16/aws_secret/entity/secret.ts
```diff
@@ -0,0 +1,43 @@
...
+  @Column({
+    nullable: true,
+  })
+  description?: string;
+
+  @Column({
+    type: String,
+    nullable: true,
+  })
+  value!: string | null;
```
aguillenv, line 30: Shouldn't this be
```suggestion
  @Column({
    type: String,
    nullable: true,
  })
  value?: string | null;
```

src/services/aws_macros.ts
```diff
@@ -1,28 +1,29 @@
...
+import { SSM } from "@aws-sdk/client-ssm";
+import { SecretsManager } from "@aws-sdk/client-secrets-manager";
 
 type AWSCreds = {
-  accessKeyId: string,
-  secretAccessKey: string
-}
+  accessKeyId: string;
+  secretAccessKey: string;
+};
```
aguillenv, line 21: Prettier attacking 😂

src/modules/0.0.16/aws_secret/index.ts
```diff
@@ -0,0 +1,212 @@
...
+async function putSecretValue(
+  client: SecretsManager,
+  input: PutSecretValueCommandInput
+) {
+  const res = await client.putSecretValue(input);
+  if (res) {
+    return res;
+  }
+  return undefined;
+}
```
aguillenv, line 45: I'm not asking for a change here, but aws_macros crudBuilder2 or crurBuilderFormat would have worked here. Maybe they are not that useful?

src/modules/0.0.16/aws_secret/index.ts
```diff
@@ -0,0 +1,212 @@
...
+    ...metadata,
+    utils: {
+      secretsMapper: async (secret: SecretListEntry, ctx: Context) => {
+        const client = (await ctx.getAwsClient()) as AWS;
+        const out = new Secret();
+        if (!secret.Name) return undefined;
+        out.name = secret.Name;
+        if (secret.Description) out.description = secret.Description;
+        return out;
+      },
```
aguillenv, line 93: nit:
```suggestion
      secretsMapper: (secret: SecretListEntry) => {
        const out = new Secret();
        if (!secret.Name) return undefined;
        out.name = secret.Name;
        if (secret.Description) out.description = secret.Description;
        return out;
      },
```

src/modules/0.0.16/aws_secret/index.ts
```diff
@@ -0,0 +1,212 @@
...
+  const res = await client.createSecret(input);
+  if (res) {
+    const s: SecretListEntry = {
+      Name: res.Name,
+      Description: input.Description,
+    };
+    return s;
+  }
+  return undefined;
+}
```
dfellis, line 34: This could just be

```suggestion
const createSecret = crudBuilderFormat<SecretsManager, 'createSecret', { Name: string, Description: String } | undefined>(
  'createSecret',
  (input) => input,
  (res) => !!res ? { Name: res.Name, Description: res.Description, } : undefined,
);
```
async function createSecret(
  client: SecretsManager,
  input: CreateSecretCommandInput
) {
  const res = await client.createSecret(input);
  if (res) {
    const s: SecretListEntry = {
      Name: res.Name,
      Description: input.Description,
    };
    return s;
  }
  return undefined;
}
```

dfellis: The advantage of this form is you can let Typescript's type inference take control and not need to import most of the types from the AWS SDK library.

dfellis: Most of your functions could be rewritten with one of these macro functions, btw.

dfellis (approved):

aguillenv (approved):

PR #1065, opened 2022-08-04
`do` `try` to `catch` `while` we're `await`ing for the temporary postgres user

dfellis:
Apparently Postgres' GRANT and REVOKE are not concurrent-safe. See this discussion: https://www.postgresql.org/message-id/3473.1393693757%40sss.pgh.pa.us

While reducing the number of temporary accounts needed to function is the better long-term plan, this weird `do-try-catch-while` loop lets us quasi-linearize across requests without *too* much contention and prevent request failures from reaching the users.


depombo (approved):
maybe we could wrap into a function for re-use but absolutely not important

PR #1067, opened 2022-08-04
Drop postgraphile integration

dfellis:
Now that our dashboard no longer uses GraphQL but does everything through the `/run` endpoint, we can remove this integration of Postgraphile in the engine. It was unfortunately more brittle than we hoped for and made too many assumptions that it was being run as a standalone process rather than as a library (such that its own built-in error handling logic caused an outage for us where this assumption clashed with using it as a library).

Anyone who wants to use GraphQL with IaSQL is still welcome to, by just using the `postgresql://` connection string and giving it to Postgraphile as a standalone server or to Hasura, or etc.


src/modules/0.0.16/iasql_functions/sql/create_fns.sql
```diff
@@ -147,7 +147,7 @@ begin
 end;
 $$;
 
-create or replace function iasql_module_install(_mods text[]) returns table (
```
dfellis, line 150: The `iasql_module_install` and `iasql_module_uninstall` functions were to work around a postgraphile issue with variadic functions, and were not explicitly documented. Now that we are not supporting postgraphile, we do not need them.

aguillenv (approved):

PR #1068, opened 2022-08-04
Migrate the 'iasql_*' modules to the new class approach

dfellis:
These modules are special versus the rest of them as they are manipulated directly by various `src/services` files, and mostly act as the foundational layer for the useful IaSQL modules, so I converted them in a separate PR to keep this somewhat contained.

Lots of conditional logic in the `src/services` files added that can be removed when this version is the oldest supported version.

### DIFF
--- a/src/modules/0.0.16/iasql_platform/index.ts
+++ b/src/modules/0.0.16/iasql_platform/index.ts
@@ -8,7 +8,7 @@ class IasqlPlatform extends ModuleBase {
   constructor() { super(); super.init(); }
   dirname = __dirname;
   dependencies = metadata.dependencies;
-  IasqlModule = IasqlModule;
-  IasqlTables = IasqlTables;
+  iasqlModule = IasqlModule;
+  iasqlTables = IasqlTables;
 }
 export const iasqlPlatform = new IasqlPlatform();
--- a/src/modules/0.0.16/iasql_platform/index.ts
+++ b/src/modules/0.0.16/iasql_platform/index.ts
@@ -1,14 +1,14 @@
 /* THIS MODULE IS A SPECIAL SNOWFLAKE. DON'T LOOK AT IT FOR HOW TO WRITE A REAL MODULE */
 
-import { Module2, } from '../../interfaces'
+import { ModuleBase, } from '../../interfaces'
 import * as metadata from './module.json'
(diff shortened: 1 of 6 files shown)


src/modules/0.0.16/iasql_platform/index.ts
```diff
@@ -1,14 +1,14 @@
...
-  },
-  mappers: {},
-}, __dirname);
\ No newline at end of file
+class IasqlPlatform extends ModuleBase {
+  constructor() { super(); super.init(); }
+  dirname = __dirname;
+  dependencies = metadata.dependencies;
+  IasqlModule = IasqlModule;
+  IasqlTables = IasqlTables;
```
aguillenv, line 12: any reason to have these 2 capitalized? I think lines like https://github.com/iasql/iasql-engine/pull/1068/files#diff-c8f7acf4155d692e9663c0ae07f0ab80ceb39ab0c6d7e15d9a2e987aa1a04ac9R806 read weird

aguillenv (approved):
It is quite confusing indeed with all the conditionals, but I get it is temporary.

src/modules/0.0.16/iasql_platform/index.ts
dfellis: I was just trying to keep the difference between the versions minimized, but I guess with all of the conditionals in there anyways, I don't have to.

I can adjust them.

yrobla (approved):

aguillenv (approved):

src/services/iasql.ts
```diff
@@ -1111,12 +1117,16 @@ export async function upgrade(dbId: string, dbUser: string) {
         // 4. Uninstall the `iasql_*` modules manually
         const OldModules = (AllModules as any)[versionString];
         const qr = conn.createQueryRunner();
-        await OldModules.IasqlFunctions.migrations.remove(qr);
-        await OldModules.IasqlPlatform.migrations.remove(qr);
+        await OldModules?.IasqlFunctions?.migrations?.remove(qr);
+        await OldModules?.iasqlFunctions?.migrations?.remove(qr);
```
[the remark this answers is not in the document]
dfellis: It doesn't. See the naming difference between the two? In the refactor it's lowercased. This simply runs one version or the other depending on which capitalization is used.

PR #1071, opened 2022-08-04
Convert 'aws_vpc' to the new class approach

dfellis:
Part of #862. This one shows some issues with leaving the Mapper and Crud objects alone. Most things still need to reference the singleton object, so mock-based testing is out of reach (not that that was the main goal here, but eventually getting there seemed reasonable).

Once this PR is passing tests, I am going to think about how to refactor the mappers, too.


yrobla (approved):

dfellis:
This one missed the boat, and I also improved the process, so closing this one out and will open another one later.

PR #1074, opened 2022-08-05
iam custom policy deep comparison

aguillenv:


src/modules/0.0.16/aws_iam/index.ts
```diff
@@ -138,7 +138,23 @@ export const AwsIamModule: Module2 = new Module2({
       // EC2 role instance profile ARN example - arn:aws:iam::257682470237:instance-profile/test-role
       return  arn.split('/').pop();
     },
-    rolePolicyComparison: (a: any, b: any) => isEqual(a, b),
+    // rolePolicyComparison: (a: any, b: any) => isEqual(a, b),
+    rolePolicyComparison: (a: any, b: any) => {
+      if (Object.is(a, b)) return true;
+      if (Object.is(a, null) || Object.is(a, null) || !Object.is(typeof a, 'object') || !Object.is(typeof b, 'object')) return false;
```
dfellis, line 144: Did you mean

```suggestion
      if (Object.is(a, null) || Object.is(b, null) || !Object.is(typeof a, 'object') || !Object.is(typeof b, 'object')) return false;
```

src/modules/0.0.16/aws_iam/index.ts
```diff
@@ -138,7 +138,23 @@ export const AwsIamModule: Module2 = new Module2({
...
-    rolePolicyComparison: (a: any, b: any) => isEqual(a, b),
+    // rolePolicyComparison: (a: any, b: any) => isEqual(a, b),
+    rolePolicyComparison: (a: any, b: any) => {
+      if (Object.is(a, b)) return true;
+      if (Object.is(a, null) || Object.is(b, null) || !Object.is(typeof a, 'object') || !Object.is(typeof b, 'object')) return false;
+      const aKeys = Object.keys(a);
+      const bKeys = Object.keys(b);
+      if (!Object.is(aKeys.length, bKeys.length)) return false;
+      if (Array.isArray(a) && Array.isArray(b)) {
+        return a.every(ai => !!b.find(bj => AwsIamModule.utils.rolePolicyComparison(ai, bj)))
```
dfellis, line 149: Vulnerable to duplicate records, like this:

Screenshot from 2022-08-05 10-22-33

Working with sorted arrays will help, though:

Screenshot from 2022-08-05 10-24-13


```suggestion
        const aSorted = [...a].sort();
        const bSorted = [...b].sort();
        for (let i = 0; i < aSorted.length; i++) {
          if (!AwsIamModule.utils.rolePolicyComparison(a[i], b[i])) return false;
        }
        return true;
```

src/modules/0.0.16/aws_iam/index.ts
```diff
@@ -138,7 +138,27 @@ export const AwsIamModule: Module2 = new Module2({
       // EC2 role instance profile ARN example - arn:aws:iam::257682470237:instance-profile/test-role
       return  arn.split('/').pop();
     },
-    rolePolicyComparison: (a: any, b: any) => isEqual(a, b),
+    rolePolicyComparison: (a: any, b: any) => {
+      if (Object.is(a, b)) return true;
```
aguillenv, line 142:
```suggestion
      if (isEqual(a, b)) return true;
```

dfellis (approved):

src/modules/0.0.16/aws_iam/index.ts
```diff
@@ -138,7 +138,27 @@ export const AwsIamModule: Module2 = new Module2({
...
+          if (!AwsIamModule.utils.rolePolicyComparison(aSorted[i], bSorted[i])) return false;
+        }
+        return true;
+      } else {
+        for (const ak of aKeys) {
+          if (!bKeys.includes(ak)) return false;
+          if (!AwsIamModule.utils.rolePolicyComparison(a[ak], b[ak])) return false;
+        }
+      }
+      return true;
```
depombo, line 160: do we need this in other places where we have policies like s3?

test/modules/aws-iam-integration.ts
```diff
@@ -149,12 +169,33 @@ describe('IAM Integration Testing', () => {
...
+
+  it('check preview', query(`
+    SELECT *
+    FROM iasql_preview_apply();
+  `, (res: any[]) => expect(res.length).toBe(1)));
+
+  it('check preview', query(`
+    SELECT *
+    FROM iasql_preview_apply();
+  `, (res: any[]) => expect(res.length).toBe(1)));
```
depombo, line 197: maybe leave a comment about why we are doing this here?

depombo (approved):
minor comments that can happen in follow-up PRs

src/modules/0.0.16/aws_iam/index.ts
depombo: https://github.com/iasql/iasql-engine/issues/1077

test/modules/aws-iam-integration.ts
```diff
@@ -149,12 +169,33 @@ describe('IAM Integration Testing', () => {
...
+    WHERE role_name = '${principalServArr}';
+  `, (res: any[]) => expect(res.length).toBe(1)));
+
   it('applies the role additions', apply());
 
   it('tries to update a autogenerated field', query(`
     UPDATE role SET arn = 'dummy' WHERE role_name = '${taskRoleName}';
   `));
 
+  it('check preview', query(`
```
aguillenv, line 184:
```suggestion
  // The next 3 checks are to validate that the deep policy comparison is working properly
  it('check preview', query(`
```

PR #1076, opened 2022-08-05
Issue #1035: Create subscriber for managing nulls

yrobla:
Create a generic subscriber, that can be mapped to all entities,
to handle the null/undefined cases.

Closes Issue #1035


src/modules/subscribers.ts
```diff
@@ -0,0 +1,36 @@
...
+  }    
+
+@EventSubscriber()
+export class NullCheckerSubscriber<T> implements EntitySubscriberInterface<T> {
+    constructor(private cls: new (...a:any) => T) {
+
+    }
+    listenTo(): any {
+      return this.cls;
+    }
```
dfellis, line 17: Don't include this method. It says if you want to subscribe to all entities to *not* define this. 

> To listen to any entity you just omit `listenTo` method and use `any`

https://github.com/typeorm/typeorm/blob/master/docs/listeners-and-subscribers.md

dfellis (approved):

PR #1083, opened 2022-08-08
Clean up timeout

aguillenv:
This job is getting stuck sometimes initializing the engine. Let's add a timeout, for now, to avoid endless runnings while it is addressed properly


dfellis (approved):

dfellis:
Can we add this to all of the CI tests, too? Occasionally a test will take longer than 30 minutes, but not a single successful test hits 45 minutes, so we could make that a threshold there.

PR #1086, opened 2022-08-08
Issue #1037: Rename aws_secret module to aws_secrets_manager

yrobla:
In order to be consistent with AWS API rename the module to
match with the naming defined by AWS.


dfellis (approved):

depombo (approved):

src/modules/0.0.16/aws_secrets_manager/index.ts
```diff
@@ -105,15 +106,26 @@ export const AwsSecrectsModule: Module2 = new Module2(
...
+                  // retry until we ensure is created
+                  let rawSecret;
+                  do {
+                    await new Promise(r => setTimeout(r, 2000)); // Sleep for 2s
+
+                    rawSecret = await getSecret(
+                      client.secretsClient,
+                      secretName
+                    );
+                  } while (!rawSecret);
```
dfellis, line 124: There should be a maximum number of checks before it aborts entirely stating that the secret just wasn't created, so the outermost loop can either try again or abort completely back to the user (in case of an outage on AWS's side).

dfellis: I am not sure how long that should be, though.

PR #1091, opened 2022-08-09
Issue #1089: add versioning to secrets manager

### LINKED ISSUE #1089: Add version id to secrets manager
Every time that a new value is set to a secret, the version number is changed. We need to keep track of this number, and only retrieve the latest version, to avoid race conditions.

yrobla:
The secrets manager keeps a versionId number, that is being
updated each time the value changes. In order to be consistent,
map this record in the database and keep it updated.

Closes #1089


src/modules/0.0.17/aws_secrets_manager/index.ts
```diff
@@ -120,11 +124,12 @@ export const AwsSecretsManagerModule: Module2 = new Module2(
 
                     rawSecret = await getSecret(
                       client.secretsClient,
-                      secretName
+                      secretAWS.Name!
                     );
                     i++;
                   } while (!rawSecret && (i<30));
-                  secret.name = secretName;
+                  secret.name = secretAWS.Name!;
```
dfellis, line 131: Why allow the output of `getSecret` to be `undefined` and then assert that it is not to Typescript here? Better to make it return the result without changes and `throw` an error if it's undefined, I think.

aguillenv (commented):
I think it needs a pass from prettier for spacing, but the logic looks good to me.

src/modules/0.0.17/aws_secrets_manager/index.ts
```diff
@@ -20,37 +20,37 @@ class SecretMapper extends MapperBase<Secret> {
...
+      }
+    }
     return out;
   }
 
   createSecret = crudBuilderFormat<
     SecretsManager,
     "createSecret",
-    string | undefined
+    CreateSecretCommandOutput | undefined
```
dfellis, line 47: I think this can be reduced to:

```suggestion
    CreateSecretCommandOutput
```

src/modules/0.0.17/aws_secrets_manager/index.ts
```diff
@@ -20,37 +20,37 @@ class SecretMapper extends MapperBase<Secret> {
...
   >(
     "createSecret",
     (input) => input,
-    (res) => (!!res ? res.Name : undefined)
+    (res) => (!!res ? res : undefined)
```
dfellis, line 51: Assuming you also change this to

```suggestion
    (res) => res
```

dfellis: Though if you do that,  you don't need the `crudBuilderFormat` macro function and could use `crudBuilder2` directly.

src/modules/0.0.17/aws_secrets_manager/index.ts
```diff
@@ -152,24 +162,46 @@ class SecretMapper extends MapperBase<Secret> {
...
+            // retrieve updated secret to avoid race conditions
+            await new Promise(r => setTimeout(r, 2000)); // Sleep for 2s
+
+            const rawSecret = await this.getSecret(
+              client.secretsClient,
+              secret.name
+            );
+            finalSecret = await this.secretsMapper(rawSecret!);
+            i++;
+            if (!finalSecret) break;
```
dfellis, line 189: Do you really want `break` in these three, or `continue`?

yrobla: Actually i can set a continue if the value is undefined, and a break in the other 2 cases, as the version id is invalid

dfellis: So, I get the middle one being a `break`. It created the secret and has a new `versionId` that we want to use, so leaving at that point makes sense, but the last one I thought should be a `continue` as well, because somehow the updated secret doesn't have a value, and I don't understand what an update in that situation even means, and figured it was also some sort of transient error case that needs to be handled.

yrobla: Last one also contains the versionId that we want. In the case of not updating the value, the versionId does not change. So in this case, i just check that the return value contains a versionId and then i break the flow

src/modules/0.0.17/aws_secrets_manager/index.ts
yrobla: I tried to use crudBuilder2, but then it gives me trouble on the input/output types. How shall the input and output values be passed/collected?

src/modules/0.0.17/aws_secrets_manager/index.ts
dfellis: So it's the path taken when the `description` field is changed and nothing else is, I presume?

yrobla: Yes, that's it

dfellis (approved):

aguillenv (approved):

PR #1092, opened 2022-08-09
Class-ify the S3 module

dfellis:
Part of #862. I decided to switch to a simpler module for this first pass, and I made the Mapper also a class, but left the Crud properties mostly alone since there's exactly 1 or 2 of them per mapper, unlike the unknown number of mappers per module.

I attach `this.module` to the Mapper and Crud objects so the methods can get *their instance* of the module by going `this.module.<etc>` and I confirmed it's fully typed by Typescript, so I'm happier with this.

I don't like the `dirname = __dirname` that's required right now, since it's an easy-to-forget line to include, so I added a dependency on `callsite` which let's me get the file information from every line of a stack trace, and use it to find the directory where a module is defined in the new `ModuleBase` base class.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
   },
   "homepage": "https://iasql.com",
   "devDependencies": {
+    "@types/callsite": "^1.0.31",
     "@types/cors": "^2.8.12",
     "@types/jest": "^27.5.0",
     "@types/lodash.isequal": "^4.5.6",
@@ -77,6 +78,7 @@
     "@types/lodash.pick": "^4.4.7",
     "@types/semver": "^7.3.10",
     "axios": "^0.27.2",
+    "callsite": "^1.0.0",
     "cors": "^2.8.5",
     "express": "^4.18.1",
     "express-jwt": "^6.1.0",
--- a/src/modules/0.0.17/aws_account/index.ts
+++ b/src/modules/0.0.17/aws_account/index.ts
@@ -1,10 +1,31 @@
 import { AWS, } from '../../../services/aws_macros'
 import { AwsAccountEntity, } from './entity'
-import { Context, Crud2, Mapper2, ModuleBase, } from '../../interfaces'
+import { Context, Crud2, MapperBase, ModuleBase, } from '../../interfaces'
 import * as metadata from './module.json' // TODO: Eliminate this?
 
+class AccountMapper extends MapperBase<AwsAccountEntity> {
+  module: AwsAccount;
+  entity = AwsAccountEntity;
+  equals = (_a: AwsAccountEntity, _b: AwsAccountEntity) => true;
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -77,6 +78,7 @@
     "@types/lodash.pick": "^4.4.7",
     "@types/semver": "^7.3.10",
     "axios": "^0.27.2",
+    "callsite": "^1.0.0",
```
depombo, line 81: nice, glad it worked out in the end

src/modules/0.0.17/aws_account/index.ts
```diff
@@ -1,11 +1,31 @@
 import { AWS, } from '../../../services/aws_macros'
 import { AwsAccountEntity, } from './entity'
-import { Context, Crud2, Mapper2, ModuleBase, } from '../../interfaces'
+import { Context, Crud2, MapperBase, ModuleBase, } from '../../interfaces'
 import * as metadata from './module.json' // TODO: Eliminate this?
 
+class AccountMapper extends MapperBase<AwsAccountEntity> {
+  module: AwsAccount;
+  entity = AwsAccountEntity;
+  equals = (_a: AwsAccountEntity, _b: AwsAccountEntity) => true;
```
depombo, line 9: this is not related to this change, but why doesn't this have a comparison function?

src/modules/0.0.17/aws_account/index.ts
```diff
@@ -30,20 +50,12 @@ class AwsAccount extends ModuleBase {
...
-      read: (ctx: Context, id?: string) => ctx.orm.find(AwsAccountEntity, id ? {
-        where: {
-          id,
-        },
-      } : undefined),
-      update: async (_e: AwsAccountEntity[], _ctx: Context) => { /* Do nothing */ },
-      delete: async (_e: AwsAccountEntity[], _ctx: Context) => { /* Do nothing */ },
-    }),
-  });
+  awsAccount: AccountMapper;
```
depombo, line 53: why is this not `mapper: AccountMapper`? wouldn't that be more standardized?

src/modules/0.0.17/aws_account/index.ts
dfellis: There's no equivalent record in the cloud, and the cloud mapper is a no-op that does nothing but return whatever is in the database on read and executes nothing on any of the others.

This module is not the best one to see what a "normal" module would look like in the new form, that's what the S3 module rewrite is for. ;)

src/modules/0.0.17/aws_account/index.ts
dfellis: A module can have multiple mappers (and usually does). This module is an exception in that way, too, but I wanted to minimize the number of differences this one has versus other modules, if possible.

depombo (approved):

PR #1094, opened 2022-08-10
Class-ify 'aws_acm_list'

dfellis:
Part of #862

### DIFF
--- a/src/modules/0.0.17/aws_acm_import/index.ts
+++ b/src/modules/0.0.17/aws_acm_import/index.ts
@@ -1,7 +1,7 @@
 import { ACM, ImportCertificateCommandInput, paginateListCertificates, } from '@aws-sdk/client-acm'
 import { AWS, paginateBuilder, } from '../../../services/aws_macros'
 import { Context, Crud2, MapperBase, ModuleBase, } from '../../interfaces'
-import { AwsAcmListModule } from '../aws_acm_list'
+import { awsAcmListModule } from '../aws_acm_list'
 import { CertificateImport } from './entity'
 import * as metadata from './module.json'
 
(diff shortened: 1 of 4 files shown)


src/modules/0.0.17/aws_acm_import/index.ts
```diff
@@ -29,6 +29,19 @@ class CertificateImportMapper extends MapperBase<CertificateImport> {
...
+    delete: (es: CertificateImport[], ctx: Context) => ctx.orm.remove(CertificateImport, es),
+    read: async (ctx: Context, id?: string) => {
+      const opts = id ? {
+        where: {
+          id,
+        }
+      } : {};
+      return await ctx.orm.find(CertificateImport, opts);
+    },
+  });
```
dfellis, line 44: Due to a bug in how we were detecting `@cloudId` vs `entityId` usage, these were being autogenerated *incorrectly* before, and only worked for our usage by chance.

Fixed that bug and now added these in.

src/modules/interfaces.ts
```diff
@@ -299,7 +299,7 @@ export class MapperBase<E> {
       this.db.entityId = this.entityId;
       this.db.dest = 'db';
       this.db.module = this.module;
-    } else if (!!cloudColumn) {
+    } else if (!!cloudColumn && !((cloudColumn as any) instanceof Error)) {
```
dfellis, line 302: This is the bugfix. `getCloudId` doesn't return `undefined` when not specified, it returns an `Error` object instead, which is truthy.

depombo (approved):

src/modules/interfaces.ts
aguillenv: how did you catch it??

dfellis: When converting this module I forgot to include the `module` property on the mapper and was getting errors that didn't make sense, so I read through the initialization logic *and* the functions it calls, noticing this error and fixing it but it didn't fix the bug I was seeing, and then I finally realized what I had done wrong during the conversion process.

PR #1100, opened 2022-08-12
Class-ify 'aws_lambda'

dfellis:
Also eliminate the need to import the `module.json` file manually just to attach the dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Infrastructure as SQL",
   "main": "dist/index.js",
   "scripts": {
-    "build": "yarn clean && tsc && copyfiles -u 1 src/**/*.sql dist/",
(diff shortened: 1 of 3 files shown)


depombo (approved):

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Infrastructure as SQL",
   "main": "dist/index.js",
   "scripts": {
-    "build": "yarn clean && tsc && copyfiles -u 1 src/**/*.sql dist/",
+    "build": "yarn clean && tsc && copyfiles -u 1 src/**/*.sql dist/ && copyfiles -u 1 src/**/*.json dist/",
```
aguillenv, line 7: why do we need this?

PR #1101, opened 2022-08-12
Class-ify 'aws_ecr'

dfellis:


dfellis:
https://github.com/iasql/iasql-engine/runs/7811880412?check_suite_focus=true#step:5:1314

AWS is toggling back and forth between the `Action` field being an array with one value vs being a single string :/

depombo (approved):

PR #1107, opened 2022-08-13
Drop explicit dependency declarations

dfellis:
The new class-based modules don't need to explicitly declare their dependencies anymore, so dropping that from the converted modules.


depombo (approved):
nice!

PR #1124, opened 2022-08-15
move docs into engine

depombo:


docs/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "web",
+  "version": "0.0.0",
```
dfellis, line 3:
```suggestion
  "version": "0.0.16",
```

And we probably should hook the version bump script up to this file, too, in the future (along with running the docusaurus command to create a new version to release to the docs).

docs/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "web",
```
dfellis, line 2:
```suggestion
  "name": "docs",
```

dfellis (approved):
We'll follow up on things later, let's just get docs working again. :)

PR #1128, opened 2022-08-16
Issue #1084: Add api gateway coverage

### LINKED ISSUE #1084: Implement `aws_api_gateway`

yrobla:
Adding the basic coverage for API gateway, with the
    possibility of creating REST API entries.
    
    Closes #1084


dfellis:
@yrobla we can't make changes to `v0.0.18` anymore (or we break uninstall logic for anyone who has already upgraded) so can you move these changes to `v0.0.19` instead? (probably just copy-paste the `src/modules/0.0.18/aws_api_gateway` directory on top of `src/modules0.0.19/aws_api_gateway` and then `git reset --hard origin/main -- src/modules/0.0.18`)

src/services/aws_macros.ts
```diff
@@ -39,8 +41,14 @@ type ArgumentTypes<F> = F extends (...args: infer A) => infer B ? [...A, B] : ne
 // break.
 type PromiseReturnType<F> = ArgumentTypes<ArgumentTypes<F>[2]>[1];
 
+// defaults for slow retries
+const SLOW_STRATEGY_RETRIES = 10; // retry a max of 10 times, it will allow to bypass quotas
+const SLOW_STRATEGY_BASE_DELAY = 1000; // wait 1 second - as only delete api has a 5 second quota
+const SLOW_STRATEGY_MAXIMUM_RETRY_DELAY = 20 * 5000;
```
dfellis, line 47: Nit: Why this weird way to write 100 seconds? `100 * 1000` would take a bit less time to parse, I think?

src/services/aws_macros.ts
```diff
@@ -59,9 +67,26 @@ export class AWS {
...
+  slowRetryStrategy: StandardRetryStrategy;
 
   constructor(config: AWSConfig) {
+    // declare an specific slow retry strategy, to reuse in slow apis
+    this.slowRetryStrategy = new StandardRetryStrategy(async () => SLOW_STRATEGY_RETRIES, {
+      delayDecider: (_, attempts) =>
+        Math.floor(
+          Math.min(
+            SLOW_STRATEGY_MAXIMUM_RETRY_DELAY,
+            Math.random() * 2 ** attempts * SLOW_STRATEGY_BASE_DELAY,
```
dfellis, line 79: Slapping in the `Math.random()` in this way causes a highly variable retry delay from 0ms to `(2^attempts) * SLOW_STRATEGY_BASE_DELAY`ms. Perhaps this would be a bit better?

```suggestion
            (Math.random() + 2) ** attempts * SLOW_STRATEGY_BASE_DELAY,
```

Then it will always take at least `(2^attempts) * SLOW_STRATEGY_BASE_DELAY` in time, but could take up to `(3^attempts) * SLOW_STRATEGY_BASE_DELAY)`

yrobla: i just picked the strategy used by aws, but increasing the delay time. We can optimize that for the future, but if we modify we could create a default strategy for all clients.

dfellis (approved):

src/services/aws_macros.ts
```diff
@@ -59,9 +67,26 @@ export class AWS {
   secretsClient: SecretsManager;
   cloudfrontClient: CloudFront;
   memoryDBClient: MemoryDB;
+  slowRetryStrategy: StandardRetryStrategy;
```
aguillenv, line 70: nit: maybe be more specific here like `apiGatewayRetryStrategy`

src/services/aws_macros.ts
```diff
@@ -39,8 +41,14 @@ type ArgumentTypes<F> = F extends (...args: infer A) => infer B ? [...A, B] : ne
 // break.
 type PromiseReturnType<F> = ArgumentTypes<ArgumentTypes<F>[2]>[1];
 
+// defaults for slow retries
+const SLOW_STRATEGY_RETRIES = 10; // retry a max of 10 times, it will allow to bypass quotas
+const SLOW_STRATEGY_BASE_DELAY = 1000; // wait 1 second - as only delete api has a 5 second quota
+const SLOW_STRATEGY_MAXIMUM_RETRY_DELAY = 20 * 5000;
+
```
aguillenv, line 48: Same here if they are specific to API Gateway, something like `API_GATEWAY_SLOW_STRATEGY_RETRIES`

src/modules/0.0.19/aws_api_gateway/index.ts
```diff
@@ -0,0 +1,165 @@
...
+  getApis = crudBuilderFormat<ApiGatewayV2, 'getApis', ApiAWS[] | undefined>(
+    'getApis',
+    () => ({}),
+    res => res?.Items,
+  );
+
+  async createApi(client: ApiGatewayV2, input: CreateApiCommandInput) {
+    const newRestApi = await client.createApi(input);
+    return newRestApi;
+  }
```
aguillenv, line 37: nit: why not `crudBuilder2` here?

src/modules/0.0.19/aws_api_gateway/index.ts
```diff
@@ -0,0 +1,165 @@
...
+    },
+  });
+
+  constructor(module: AwsApiGatewayModule) {
+    super();
+    this.module = module;
+    super.init();
+  }
+
+  apiMapper(instance: any) {
```
aguillenv, line 139: this needs to be `any` or it could be `ApiAWS`

src/modules/0.0.19/aws_api_gateway/index.ts
```diff
@@ -0,0 +1,165 @@
...
+    const r: Api = new Api();
+    if (!instance.ApiId) throw new Error('Received an API without a id');
+    if (!instance.Name) throw new Error('Received an API without a name');
```
aguillenv, line 142: here we are not `throw`ing but returning `undefined`, this way we do not fail and just ignore misconfigured resources.

src/modules/0.0.19/aws_api_gateway/index.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  cloud = new Crud2<Api>({
+    create: async (rs: Api[], ctx: Context) => {
+      const client = (await ctx.getAwsClient()) as AWS;
+      const out = [];
+      for (const r of rs) {
+        // add a default protocol
+        if (!r.protocolType) r.protocolType = Protocol.HTTP;
+
+        // if we have an id already, check if exists
```
aguillenv, line 51: nit: do not follow this comment? maybe is a legacy?

aguillenv (approved):

src/services/aws_macros.ts
yrobla: Actually i wanted to make it generic... we can use it for apiGateway but maybe we can reuse in the future for other apis that are slow?

src/modules/0.0.19/aws_api_gateway/index.ts
yrobla: I actually needed to add any, because the mappers for Create/Update/Get are not consistent with Name property, some of them are seeing Name as mandatory, some others don't

src/services/aws_macros.ts
yrobla: i want to make it generic

PR #1129, opened 2022-08-17
Block most operations during an upgrade

### LINKED ISSUE #1090: Upgrading can fail if there is a concurrent query on the database
Screenshot from 2022-08-09 08-33-25

Particularly if there's a query involving one of the `iasql_functions`, which could be triggered by the 30-second polling the dashboard performs on all databases while it is running. If a database is in the process of being upgraded, we should simply block requests to `/run`.

One possible solution is to have an `isUpgrading` flag on the database table in the metadata database. This would require two DB queries per desired query, but since the workers are stateless, this seems unavoidable?

dfellis:
Resolves #1090

### DIFF
--- a/src/entity/index.ts
+++ b/src/entity/index.ts
@@ -33,6 +33,12 @@ export class IasqlDatabase {
   })
   operationCount: number;
 
+  @Column({
+    type: 'boolean',
+    default: false,
+  })
+  upgrading: boolean;
+
   @CreateDateColumn()
   createdAt: Date;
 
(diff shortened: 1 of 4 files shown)


aguillenv (approved):

depombo:
I believe there is a unit test that tests the upgrade functionality. it would be nice to incorporate testing for this change there?

dfellis:
> I believe there is a unit test that tests the upgrade functionality. it would be nice to incorporate testing for this change there?

So [this is the only upgrade test I am aware of](https://github.com/iasql/iasql-engine/blob/main/test/common/total-failure.ts#L104) and it checks that it early returns because there's nothing to do and the new code would never trigger the `upgrading` flag change.

depombo (approved):

PR #1145, opened 2022-08-22
pure SQL ecs simplified module

depombo:
To avoid custom mappers at SQL level and for the module to be "stateless" such that install/uninstall-ing works, we completely recreate all entries in `ecs_simplified` when there is any change to the underlying tables by simply looping over the existing rows in `service` and inserting the ones that conform to the defaults of the ECS simplified module


aguillenv (commented):
looking good! It does look simpler to maintain. My question here is, if someone creates a service using `ecs_simplified`, then uninstall the module and installs it again, are they going to see the resource they just created?

yrobla (approved):

src/modules/0.0.18/aws_ecs_simplified/entity/ecs_simplified.ts
```diff
@@ -47,4 +47,15 @@ export class EcsSimplified {
 
   @Column({ nullable: true, })
   loadBalancerDns?: string;
+
+  @Column({
+    default: false,
+  })
+  forceNewDeployment: boolean;
```
depombo, line 54: I am actually thinking to keep things simpler we shouldn't expose this, but set it to `true` in `service` always.

src/modules/0.0.18/aws_ecs_simplified/entity/ecs_simplified.ts
```diff
@@ -47,4 +47,15 @@ export class EcsSimplified {
 
   @Column({ nullable: true, })
   loadBalancerDns?: string;
+
+  @Column({
+    default: false,
+  })
```
depombo, line 53: I'm actually thinking we shouldn't expose this and just always set it to `true` in `service`

depombo:
> looking good! It does look simpler to maintain. My question here is, if someone creates a service using `ecs_simplified`, then uninstall the module and installs it again, are they going to see the resource they just created?

Updated the PR description explaining how I get around this.

src/modules/0.0.19/aws_ecs_simplified/sql/create_triggers.sql
```diff
@@ -0,0 +1,320 @@
...
+$$;
+
+CREATE OR REPLACE FUNCTION get_cpu_mem_enum_from_parts(cpu INTEGER, mem INTEGER)
+RETURNS TEXT LANGUAGE plpgsql AS $$
+DECLARE
+  _cpu DECIMAL;
+  _mem DECIMAL;
+BEGIN
+  _cpu = ROUND(GREATEST(COALESCE(cpu, 256), 256) / 1024.0, 2);
+  _mem = ROUND(GREATEST(COALESCE(mem, 512), 256) / 1024.0, 2);
```
aguillenv, line 22: nit: should this be
```suggestion
  _mem = ROUND(GREATEST(COALESCE(mem, 512), 512) / 1024.0, 2);
```
?

src/modules/0.0.19/aws_ecs_simplified/sql/create_triggers.sql
```diff
@@ -0,0 +1,320 @@
...
+  -- setup ecs: cloudwatch + iam + ecr
+  INSERT INTO log_group (log_group_name) VALUES (NEW.app_name || '-log-group');
+
+  INSERT INTO cluster (cluster_name) VALUES(NEW.app_name || '-cluster');
+
+  INSERT INTO role (role_name, assume_role_policy_document, attached_policies_arns)
+  VALUES (NEW.app_name || '-ecs-task-exec-role', '{"Version":"2012-10-17","Statement":[{"Sid":"","Effect":"Allow","Principal":{"Service":"ecs-tasks.amazonaws.com"},"Action":"sts:AssumeRole"}]}', array['arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy']);
+
+  INSERT INTO task_definition ("family", task_role_name, execution_role_name, cpu_memory)
+  VALUES (NEW.app_name || '-td', NEW.app_name || '-ecs-task-exec-role', NEW.app_name || '-ecs-task-exec-role', NEW.cpu_mem::TEXT::task_definition_cpu_memory_enum);
```
aguillenv, line 73: could not be cast the enum directly?

src/modules/0.0.19/aws_ecs_simplified/sql/create_triggers.sql
```diff
@@ -0,0 +1,320 @@
...
+    is_valid = _app_name IS NOT NULL AND _app_port IS NOT NULL AND _security_group_id IS NOT NULL;
+
+    is_valid = is_valid AND 1 = (SELECT COUNT(*) FROM security_group_rule WHERE security_group_id = _security_group_id AND is_egress = false AND ip_protocol = 'tcp' AND from_port = _app_port AND to_port = _app_port AND cidr_ipv4 = '0.0.0.0/0');
+
+    is_valid = is_valid AND 1 = (SELECT COUNT(*) FROM security_group_rule WHERE security_group_id = _security_group_id AND is_egress = true AND ip_protocol = '-1' AND from_port = -1 AND to_port = -1 AND cidr_ipv4 = '0.0.0.0/0');
+
+    SELECT cpu_memory::TEXT
+    INTO _cpu_mem
+    FROM task_definition
+    WHERE id = _task_definition_id AND execution_role_name = task_role_name;
```
aguillenv, line 272: is the `AND execution_role_name = task_role_name` comparison needed?

src/modules/0.0.19/aws_ecs_simplified/sql/create_triggers.sql
```diff
@@ -0,0 +1,320 @@
...
+    WHERE task_definition_id = _task_definition_id AND host_port = _app_port AND container_port = _app_port AND essential = true LIMIT 1;
+
+    IF _cpu_mem IS NULL THEN
+      _cpu_mem = get_cpu_mem_enum_from_parts(_cpu, _mem);
+    END IF;
+
+    IF _repository_name IS NOT NULL THEN
+      SELECT repository_uri INTO _repository_uri FROM repository WHERE repository_name = _repository_name;
+    END IF;
+
```
aguillenv, line 287: should not be here an `is_valid` check for the props you just have checked (cpu_mem and/or repository_name)?

src/modules/0.0.19/aws_ecs_simplified/sql/create_triggers.sql
depombo: that is a typo, yes

src/modules/0.0.19/aws_ecs_simplified/sql/create_triggers.sql
depombo: unfortunately, no

src/modules/0.0.19/aws_ecs_simplified/sql/create_triggers.sql
depombo: the simplified module has them be the same, so yes?

src/modules/0.0.19/aws_ecs_simplified/sql/create_triggers.sql
depombo: well, both parameters can be null/optional until they hit their respective conditionals. or do you mean adding a new check after the conditionals? I will do that

depombo: actually, `_repository_name` *can* be NULL so that broke the test. reverted back to what I had

src/modules/0.0.19/aws_ecs_simplified/sql/create_triggers.sql
```diff
@@ -0,0 +1,320 @@
...
+      VALUES (_app_name, _desired_count, _app_port, _cpu_mem::ecs_simplified_cpu_mem_enum, _image_tag, _public_ip, _load_balancer_dns, _repository_uri, _env_variables, _force_new_deployment);
+    END IF;
+  END LOOP;
+
+  RETURN NULL;
+END
+$$;
+
+CREATE TRIGGER ecs_simplified_service_trigger
+AFTER INSERT OR DELETE OR UPDATE ON service
```
aguillenv, line 313: my last question would be, what happens when any other table is edited that might cause a misconfiguration?

depombo: good point :) I realized that today and I was going to address that in a follow up PR since it requires adding more tests to check for these changes and I was afraid it would be too hard to follow, plus it had already dragged on for a while

aguillenv: I'm good with that, just something to be aware

aguillenv (approved):

src/modules/0.0.19/aws_ecs_simplified/sql/create_triggers.sql
depombo: Actually, I just added it in this PR

aguillenv (approved):

PR #1146, opened 2022-08-22
Multi-Region Support RFC

dfellis:
While writing this out, realized there was a 6th option combining stuff from the others I thought of. Let me know if you can think of any other options, and what you think of the ones I've laid out so far! (And feel free to post a lot of question, too, on anything that I didn't explain clearly enough!)

### DIFF
--- a/rfcs/003 - Multi-Region Support RFC.md
+++ b/rfcs/003 - Multi-Region Support RFC.md
@@ -1,0 +1,137 @@
+# 003 - Multi-Region Suppor RFC
+
+## Current Status
+
+### Proposed
+
+2022-08-22
+
+### Accepted
+
+YYYY-MM-DD
+
+#### Approvers
+
+- Luis Fernando de Pombo <l@iasql.com>
+- Alejandro Guillen <alejandro@iasql.com>
+- Yolanda Robla <yolanda@iasql.com>
+
+### Implementation
+
+- [ ] Implemented: [One or more PRs](https://github.com/iasql/iasql-engine/some-pr-link-here) YYYY-MM-DD
+- [ ] Revoked/Superceded by: [RFC ###](./000 - RFC Template.md) YYYY-MM-DD
+
+## Author(s)
+
+- David Ellis <david@iasql.com>
+
+## Summary
+
+Due to the structure of the AWS API, as well as some hesitance to accidentally over-complicate our tables, we implemented the root `aws_account` module to require a specific AWS Region to be chosen for the database. Thus to manage your entire AWS account you would need a separate database for every region you're using, and be careful not to install modules for services that are "global" in scope like Route53 on more than one of these databases. It is also impossible to safely represent cross-region entities like DynamoDB clusters. This is an untenable situation for IaSQL for serious usage, and so enabling management across all regions within a single database must be tackled.
+
+## Proposal
+
+No explicit proposal yet, just a collection of options to consider from implementation and usability perspectives. The resulting schema needs to be intuitive for users as the highest priority, but as this migration will likely take time, the ability to get from here to there without a "break the world" parallel branch is also desired, if possible.
+
+### Alternatives Considered
+
+#### `region` columns on all relevant tables
+
+The simplest solution is to just slap a `region` column onto every table that is region-specific and cloud reads cycle through all regions every time.
+
+##### Pros
+
+1. We can implement this piecemeal, module-by-module ignoring the `region` column in the `aws_account` and doing their own thing.
+2. Very clear where resources are located.
+3. Could have a `regions` table in the `aws_account` that lists only regions your account has access to, and these can all be FKs for safety.
+
+##### Cons
+
+1. Creating records becomes more cumbersome with explicit regions all over the place.
+2. Cloud Reads likely become 20x slower than they currently are with sequential reads across all regions.
+3. Auto-populated tables from AWS become filled with records where the vast majority are "irrelevant" for most users (which default security group is *my* default security group?)
+
+#### Regions as schema names
+
+Instead of using the `public` schema, we could have each region be a [schema](https://www.postgresql.org/docs/current/ddl-schemas.html) (plus a `global` schema for non-region-based tables) and the user can `SET search_path TO us_west_2;` and continue using the exact same queries they have so far.
+
+##### Pros
+
+1. No changes to the relationships between the mappers in the modules, just a mechanism in the client to figure out which schema is involved to populate the region in the AWS client.
+2. Users can specify a preference for one region or another by just making that region's schema their default search path.
+
+##### Cons
+
+1. The same cloud read performance issue as the region-column approach.
+2. Cross-schema joins may be impossible, or very awkward, for things like a DynamoDB cluster.
+3. The set of regions can't be dynamic in this approach without some very tricky changes in the engine to re-apply the "install" operations on the newly-discovered region.
+4. More mangling of TypeORM to toggle the schema the entities are in (though this is likely not too complicated).
+
+#### Regions as table prefixes
+
+Similar to the prior one, but instead of `us_west_2.security_groups` it would be `public.us_west_2_security_groups`.
+
+##### Pros
+
+1. Every table is immediately visible unlike the regions-as-schema-names approach.
+
+##### Cons
+
+1. TypeORM entity logic needs trickery to take a single logical entity and map it to several similarly-named entities, and then somehow inform the mapper of what trickery was used for which instance.
+2. You can't hide regions you don't care about, and every query has to be altered.
+
+#### `Region` mega-join table
+
+A specialized join table that contains `table`, `column`, `value`, and `region` columns to uniquely identify *any* row of any table and specify the region that applies to it. For rows that don't have a record in there, a `default_region` column in the `aws_account` would be assumed to be the answer if it is relevant. There could also be a list of `active_regions` that can trim the AWS API access to only regions the user is interested in.
+
+##### Pros
+
+1. Also requires zero schema changes to other modules, can be done in a single shot.
+2. Preserves the "default region" concept to keep inserts effectively identical.
+3. Can actually be faster by cutting out irrelevant regions.
+
+##### Cons
+
+1. Joins would be very slow and meta, with no good way to guarantee correctness as the joins will not be enforced by Postgres, but only us.
+2. Inserts for non-default regions are all two-step, and *must* be explicitly done before `iasql_apply` or it will accidentally create in the wrong region.
+3. Querying what's in what region will be much tougher than other options.
+
+#### Region-prefixed-to-@cloudId-column
+
+The IaSQL `@cloudId` concept (an ID that isn't the autogenerated database ID, but the ID the cloud provider generates once the resource is created) can double-up as the region indicator, when necessary. On insert you insert the region in question, and once it is created in the cloud, it it set to `region:cloudId` in the column.
+
+##### Pros
+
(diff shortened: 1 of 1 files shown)


rfcs/003 - Multi-Region Support RFC.md
```diff
@@ -0,0 +1,137 @@
...
+Coming full circle on this, we can improve on the first option in this list by taking the `default_region` and `active_regions` idea from the prior two options and applying that to it.
+
+##### Pros
+
+1. Inserts *can* stay the same and not specify a region, with the mappers automatically assuming the default region in question.
+2. It can also be migrated to over time.
+
+##### Cons
+
+1. Nulling-out the region column after it has been populated can have unpredictable results, and is easier to accidentally do than the prior option.
```
depombo, line 124: This is the option I had thought about most recently after we talked on Friday. In my head the `region` column in `aws_account` would be renamed to `default_region` and still be mandatory, but fall back to `us-east-1` as the default. The `region` column in every region-specific entity could then be mandatory which would prevent this con? Unless I am missing something

dfellis: So if you require the `region` column to be defined, then you don't need the `default_region` property as it wouldn't do anything?

dfellis: Also, requiring that column then alters all `insert`s to require specification of the `region` column, which restores the same issue as in the first option listed.

depombo: I guess my thought was that every `region` column would need to be defined and would fall back to a default which is the foreign key relation to whatever the `default_region` is. However, I think your point might be what happens if the `default_region` changes after it is set?  I think we could add a trigger that updates every single `region` field?

depombo: Talked offline. Part of my confusion was that the default to the `region` column on every table is a schema change. However, we came up with yet another option that lets the default of the `region` column be defined by a function that simply looks at the `default_region` field in `aws_account`

rfcs/003 - Multi-Region Support RFC.md
```diff
@@ -0,0 +1,149 @@
...
+
+##### Pros
+
+1. This also requires no schema changes and can be updated over time.
+2. Could similarly have a `default_region` and `active_regions` concept like the mega-join table to gain its benefits.
+
+##### Cons
+
+1. Querying the records by region will be complicated with string prefix checking, so can never be performant, and also varies entity-by-entity.
+2. The cloud ID itself similarly would need to be parsed out to find the relevant part of the string, and so joins between entities *on* the cloudId become weirder.
```
aguillenv, line 111: I think not all our entities have an explicit `cloudId`

rfcs/003 - Multi-Region Support RFC.md
```diff
@@ -0,0 +1,149 @@
...
+
+Minor tweak to the above, instead of the region being nullable, they are not-nullable but have a default value equal to the `default_region` column in the `aws_account`. This can be accomplished with a simple Postgres function that returns the `default_region` value.
+
+##### Pros
+
+1. Same insert benefits as the above, but also there's never weirdness with the region not being defined.
+
+##### Cons
+
+1. There may be an expectation that changing the default region changes the values of records with the same region?
```
aguillenv, line 137: after reading all I'm leaning toward this option. We could also make the default region in aws_account as read-only, letting users just set it on db creation. They might want to change the default region but not all the resources to that region or as you said in the cons, a massive movement of resources from one region to another, but any of those would be an opinionated decision.

rfcs/003 - Multi-Region Support RFC.md
```diff
@@ -0,0 +1,149 @@
...
+
+1. No changes to the relationships between the mappers in the modules, just a mechanism in the client to figure out which schema is involved to populate the region in the AWS client.
+2. Users can specify a preference for one region or another by just making that region's schema their default search path.
+
+##### Cons
+
+1. The same cloud read performance issue as the region-column approach.
+2. Cross-schema joins may be impossible, or very awkward, for things like a DynamoDB cluster.
+3. The set of regions can't be dynamic in this approach without some very tricky changes in the engine to re-apply the "install" operations on the newly-discovered region.
+4. More mangling of TypeORM to toggle the schema the entities are in (though this is likely not too complicated).
```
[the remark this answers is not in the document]
dfellis: Well, thankfully they all refer to their regions differently so there'd be no overlap, but the approach of cloud regions being encoded in the schema does make the module system a bit weirder and should be listed as another issue with this idea.

rfcs/003 - Multi-Region Support RFC.md
dfellis: This is also true, EC2 instances don't, for instance, and we'd still have to do something weird for them. I'll add that to the cons list here.

rfcs/003 - Multi-Region Support RFC.md
dfellis: It does feel like the best of them.

I think the default region should be mutable but not-NULLable and ideally dependent on a Regions table that we pull with their credentials, but all of those requirements produce a chicken-and-egg problem on how to store the creds before writing the default region.

If you're okay with this, though, I'll elevate this approach to the suggested one and out of the alternatives considered and flesh it out a bit more?

aguillenv: definitely, something to think about. it might need to happen in a 3 step process behind scenes: 
- save cloud creds
- complete regions table
- save the default region


Yes I think we should go a bit deeper with this solution

rfcs/003 - Multi-Region Support RFC.md
```diff
@@ -0,0 +1,161 @@
...
+
+I bring this up because it suggests a solution for both problems.
+
+We split the `aws_account` into two tables, `aws_credentials` and `aws_regions`. The credentials table *only* has the credentials for the AWS account, while the `aws_regions` table has three columns, the `region`, `is_default`, and `is_enabled`, with the last two being boolean columns and the `is_default` column having a constraint that only one row has a default value.
+
+When first populating the `aws_regions` table, all regions returned that the user has access to should be enabled by default, and the first one is set as the default. We can then overload the `aws_default_region` function we intend to use for reading the default region to also perform the necessary operations to alter the default region (this requires temporarily disabling the constraint on the table only allowing one `is_default == true` row, making the current default `false`, then setting the new default to `true` and re-enabling the constraint within a transaction).
```
aguillenv, line 52: I think if in the same `UPDATE` statement you put one to false, and another to true there's no need to disable the constraint temporarily.

aguillenv (commented):
sounds like a plan to me! 🚀

rfcs/003 - Multi-Region Support RFC.md
```diff
@@ -0,0 +1,149 @@
...
+##### Pros
+
+1. We can implement this piecemeal, module-by-module ignoring the `region` column in the `aws_account` and doing their own thing.
+2. Very clear where resources are located.
+3. Could have a `regions` table in the `aws_account` that lists only regions your account has access to, and these can all be FKs for safety.
+
+##### Cons
+
+1. Creating records becomes more cumbersome with explicit regions all over the place.
+2. Cloud Reads likely become 20x slower than they currently are with sequential reads across all regions.
```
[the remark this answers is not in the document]
depombo: We should check to see if the rate-limiting is per account or per region. The aws client object is instantiated per region so I think we *might* be in the clear here and Cloud reads can be parallelized across regions.

rfcs/003 - Multi-Region Support RFC.md
```diff
@@ -0,0 +1,161 @@
...
+
+The first thing that we desire is that the default region be guaranteed to be an actual region that the user is able to access. That depends on their credentials, so we should have an `aws_regions` table contains the list of all regions that we acquire via the API, and then joined in on the `default_region` column.
+
+The second thing that is necessary with this approach is that the default region function output needs to always return a region.
+
+These two are not compatible with each other, and here's why:
+
+If we put a foreign key join on the `aws_account` table to the `aws_regions` table, then when we first attempt to insert the credentials to make it available to the Mapper in charge of the `aws_regions` table, we would have to have the `default_region` be set to `NULL` to not fall afoul of the foreign key constraints. But if it is `NULL` then the default value function would need to have some other value it always spits out to users which we don't *know* for sure could be something like `us-east-1` because it depends on the response for the regions available to the credentials in question.
```
depombo, line 44: I take it from this that AWS regions available vary per account? Perhaps we can bootstrap the initial list with default list of readily accessible regions from AWS and once we get credentials we can update it to get over this hump?

dfellis: Read a bit further on, I have a solution already :)

rfcs/003 - Multi-Region Support RFC.md
depombo, line 52: This means that the `aws_account` mapper would be able to make schema changes, right? That would be the first mapper of its kind?

dfellis: No? The schema would stay the same, calling the `aws_default_region()` function without arguments to return the default region as the default value if not provided on `insert` in other schemas.

The schema these other modules would use would have a dynamic default value instead of a static one, but the schema itself is static.

depombo: Got it. I guess it within the function and not the mapper. LGTM.

rfcs/003 - Multi-Region Support RFC.md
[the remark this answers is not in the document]
aguillenv: I think it would not be possible, but it's worth to try it out, would be a good thing if we can

aguillenv (approved):

depombo (approved):

rfcs/003 - Multi-Region Support RFC.md
mtp1376: [This](https://stackoverflow.com/a/50105479) suggests that constraints might be either `IMMEDIATE` or `DEFERRED`. And states that
> DEFERRED constraints are not checked until transaction commit
So maybe we can use a `DEFERRED` constraint and do these operations in a single transaction.

rfcs/003 - Multi-Region Support RFC.md
```diff
@@ -0,0 +1,163 @@
...
+### Alternatives Considered
+
+#### `region` columns on all relevant tables
+
+The simplest solution is to just slap a `region` column onto every table that is region-specific and cloud reads cycle through all regions every time.
+
+##### Pros
+
+1. We can implement this piecemeal, module-by-module ignoring the `region` column in the `aws_account` and doing their own thing.
+2. Very clear where resources are located.
```
mtp1376, line 71: @dfellis in your main solution, how to determine which region a resource is located in?

rfcs/003 - Multi-Region Support RFC.md
dfellis: So in the past we have decided to ignore this distinction because we can't control if the user writes their SQL inside of a transaction or not. If they don't, then deferred behaves identically to immediate and the problem would still exist.

rfcs/003 - Multi-Region Support RFC.md
dfellis: Oh, if something is region-specific it will have a region column. The columns will all get their default value from a function that returns their chosen default, so there is not change necessary for any `insert` so only users that go multiregion will have the added complexity.

PR #1153, opened 2022-08-23
Issue #1120: initial version of aws_cloudfront

yrobla:
First version of aws_cloudfront module, that only maps
    the distribution configuration. It needs proper testing,
    and adding support for collecting metadata.
    
    Closes: #1120


src/modules/0.0.18/aws_cloudfront/entity/distribution.ts
```diff
@@ -0,0 +1,90 @@
+import {
```
depombo, line 1: nit: I think we use 2 spaces instead of 4 everywhere else. We don't enforce in the lint and it is not that important, but can we keep the convention? VSCode let's you change the configuration for it

src/modules/0.0.18/aws_cloudfront/entity/distribution.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export enum originProtocolPolicyEnum {
+    HTTP_ONLY = "http-only",
+    MATCH_VIEWER = "match-viewer",
+    HTTPS_ONLY = "https-only"
+}
+
+@Entity()
+export class Distribution {
```
depombo, line 22: I had never used Cloudfront, but this data model is more complex than I anticipated. Not related to this PR, but it might also make sense to have a high level form of this module in the future

src/modules/0.0.18/aws_cloudfront/entity/distribution.ts
```diff
@@ -0,0 +1,90 @@
...
+    @Column({
+        type: 'json',
+        nullable: false,
+    })
+
+    @Column({
+        type: 'json',
+        nullable: false,
+    })
+    origins: { DomainName: string | undefined, Id: string | undefined,
```
depombo, line 72: oh I didn't realize we could type json? nice

src/modules/0.0.18/aws_cloudfront/module.json
```diff
@@ -0,0 +1,4 @@
+{
+    "name": "aws_cloudfront",
+    "dependencies": ["iasql_platform", "aws_account"]
```
depombo, line 3: not related to this PR, but we might eventually have this depend on s3 so users can expose files from it?

depombo (approved):
I am not super familiar with cloudfront, but LGTM

src/modules/0.0.18/aws_cloudfront/module.json
yrobla: I even didn't add support for s3 yet... only custom distro. But we can add in next steps

src/modules/0.0.18/aws_cloudfront/entity/distribution.ts
yrobla: we were discussing about dividing between distribution and distribution_metadata, but as a follow-up

src/modules/0.0.18/aws_cloudfront/entity/distribution.ts
```diff
@@ -0,0 +1,90 @@
...
+    type: 'json',
+    nullable: false,
+  })
+  defaultCacheBehavior: { TargetOriginId: string | undefined, ViewerProtocolPolicy: viewerProtocolPolicyEnum, CachePolicyId: string | undefined };
+
+  @Column({
+    type: 'json',
+    nullable: false,
+  })
+
```
aguillenv, line 67: I think this is duplicated
```suggestion
```

src/modules/0.0.18/aws_cloudfront/entity/distribution.ts
```diff
@@ -0,0 +1,90 @@
...
+  @Column({
+    type: 'json',
+    nullable: false,
+  })
+  origins: { DomainName: string | undefined, Id: string | undefined,
+    CustomOriginConfig: { HTTPPort: number|undefined, HTTPSPort: number|undefined, OriginProtocolPolicy: originProtocolPolicyEnum }
+  }[];
```
aguillenv, line 74: ohh be aware it is not duplicated but need to be here the decorator!

src/modules/0.0.18/aws_cloudfront/index.ts
```diff
@@ -0,0 +1,288 @@
...
+import { Distribution, viewerProtocolPolicyEnum, } from './entity'
+import { Context, Crud2, MapperBase, ModuleBase, } from '../../interfaces'
+import { WaiterOptions } from '@aws-sdk/util-waiter'
+
+class DistributionMapper extends MapperBase<Distribution> {
+    module: AwsCloudfrontModule;
+    entity = Distribution;
+    equals = (a: Distribution, b: Distribution) =>
+        Object.is(a.callerReference, b.callerReference) && Object.is(a.comment, b.comment) && Object.is(a.enabled, b.enabled) &&
+            Object.is(a.isIPV6Enabled, b.isIPV6Enabled) && Object.is(a.webACLId, b.webACLId);
```
aguillenv, line 20: hmm, I do not see here as many comparisons as entity properties. That means that if someone change one of the properties we are not comparing the row could get in a "corrupt" state because it will not have accurate information. If they are not updatable we should still compare them and restore them. This could happen in a follow-up PR, but I think we should leave a `TODO` here.

src/modules/0.0.18/aws_cloudfront/index.ts
```diff
@@ -0,0 +1,288 @@
...
+          const cloudRecord = ctx?.memo?.cloud?.Distribution?.[e.distributionId ?? ""];
+          const isUpdate = Object.is(
+            this.module.distribution.cloud.updateOrReplace(cloudRecord, e),
+            'update'
+          );
+          if (isUpdate) {
+            if (!e.eTag) throw new Error("Cannot update a distribution without an etag");  // cannot update without etag
+            const req = await this.getDistributionConfigForUpdate(client.cloudfrontClient, e);
+
+            if (e.eTag) {
```
aguillenv, line 228: nit: I think this is not necessary since we are already `throw`ing if `!e.eTag` in line 225

src/modules/0.0.18/aws_cloudfront/entity/distribution.ts
```diff
@@ -0,0 +1,80 @@
...
+  })
+  defaultCacheBehavior: { TargetOriginId: string | undefined, ViewerProtocolPolicy: viewerProtocolPolicyEnum, CachePolicyId: string | undefined };
+
+  @Column({
+    type: 'json',
+    nullable: false,
+  })
+  origins: { DomainName: string | undefined, Id: string | undefined, OriginShield: any,
+    CustomOriginConfig: { HTTPPort: number|undefined, HTTPSPort: number|undefined, OriginProtocolPolicy: originProtocolPolicyEnum }
+  }[];
```
dfellis, line 69: 1. A general note that the `type: 'json',` records have no constraints inside of the database itself, so even with the strong typing here in TS-land, there's nothing to prevent the user from setting things like `TargetOriginId: true`, so we should *definitely* think about how to eliminate these in the future.
2. It would be more legible to format these like this:

```suggestion
  @Column({
    type: 'json',
    nullable: false,
  })
  defaultCacheBehavior: {
    TargetOriginId: string | undefined,
    ViewerProtocolPolicy: viewerProtocolPolicyEnum,
    CachePolicyId: string | undefined
  };

  @Column({
    type: 'json',
    nullable: false,
  })
  origins: {
    DomainName: string | undefined,
    Id: string | undefined,
    OriginShield: any,
    CustomOriginConfig: {
      HTTPPort: number | undefined,
      HTTPSPort: number | undefined,
      OriginProtocolPolicy: originProtocolPolicyEnum
    }
  }[];
```

src/modules/0.0.18/aws_cloudfront/index.ts
```diff
@@ -0,0 +1,323 @@
...
+import isEqual from 'lodash.isequal'
+import _, { forEach } from 'lodash'
+
+class DistributionMapper extends MapperBase<Distribution> {
+    module: AwsCloudfrontModule;
+    entity = Distribution;
+    equals = (a: Distribution, b: Distribution) => {
+      // specific origins comparison
+      const originsA = a.origins.map(({Id, DomainName}) => ({Id, DomainName}));
+      const originsB = b.origins.map(({Id, DomainName}) => ({Id, DomainName}));
```
dfellis, line 23: Won't the later `isEqual` call depend on origin ordering in these output arrays?

Perhaps it'd be better to do something like:

```suggestion
      const originsA = Object.fromEntries(a.origins.map(({Id, DomainName}) => ([Id, DomainName])));
      const originsB = Object.fromEntries(b.origins.map(({Id, DomainName}) => ([Id, DomainName])));
```

This turns it into objects where the keys are the `Id` values` and the values are the `DomainName` values, and object comparison doesn't care about key ordering, I am pretty sure.

src/modules/0.0.18/aws_cloudfront/index.ts
```diff
@@ -0,0 +1,323 @@
...
+}
+
+class AwsCloudfrontModule extends ModuleBase {
+    distribution: DistributionMapper;
+
+    constructor() {
+    super();
+    this.distribution = new DistributionMapper(this);
+    super.init();
+    }
```
dfellis, line 321: Nit

```suggestion
    constructor() {
      super();
      this.distribution = new DistributionMapper(this);
      super.init();
    }
```

src/modules/0.0.18/aws_cloudfront/index.ts
```diff
@@ -0,0 +1,323 @@
...
+          }
+        }
+      },
+    });
+
+    constructor(module: AwsCloudfrontModule) {
+        super();
+        this.module = module;
+        super.init();
+      }
```
dfellis, line 311: Nit

```suggestion
    constructor(module: AwsCloudfrontModule) {
      super();
      this.module = module;
      super.init();
    }
```

src/modules/0.0.18/aws_cloudfront/index.ts
yrobla: i wasn't aware that lodash was checking the order to compare, good catch

dfellis (approved):

aguillenv (approved):

PR #1157, opened 2022-08-26
Prettier

### LINKED ISSUE #1171: Add code style rules

aguillenv:
Resolves #1171 
Massive PR... but it will give us code style consistency.

- Rules defined in `.prettierrc`
- Add `style` job to be part of the CI flow
- Add `format` script to autoformat the code before pushing


.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
+  "semi": true,
+  "trailingComma": "es5",
```
dfellis, line 3: Why not `true` here?

dfellis, line 3: Why not `true` here?

.prettierrc
```diff
@@ -0,0 +1,11 @@
...
+  "singleQuote": true,
+  "printWidth": 120,
+  "useTabs": false,
+  "tabWidth": 2,
+  "bracketSpacing": true,
+  "arrowParens": "avoid",
+  "endOfLine": "lf"
+}
```
dfellis, line 11: Is there a prettier rule on import order?

The style I took from Python developers is:

```
Built-in module imports

Third-party module imports

Project relative path imports
```

with each block sorted alphabetically.

.prettierrc
aguillenv: These are the possible values https://prettier.io/docs/en/options.html#trailing-commas. I set it to `es5` since it seems to be the default. I could change it to `all`

.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
+  "semi": true,
+  "trailingComma": "es5",
+  "singleQuote": true,
+  "printWidth": 120,
```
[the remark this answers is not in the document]
aguillenv: Sounds good!

.prettierrc
aguillenv: Oh this would be nice, I did not see it in the docs, but maybe there's a third-party plugin for it

.prettierrc
dfellis: So I personally prefer `all` because re-ordering lists won't accidentally produce a situation where two array-like values are next to each other without a separating comma between them.

dfellis: A more practical example:

If there's a lot of imports from a module listed and I want to sort them but it's still a single line, I navigate to the import line in question, type `:s/\([{,]\)/\1\r/g` which instantly, correctly multi-lines the import statement, then I use `V` mode to highlight the imports and then type `:sort` to alphabetically sort them. Then in `V` mode again I highlight the whole import multi-line statement and hit `Shift+J` to immediately snap it back into a single line import again, but with re-ordered imports.

This doesn't work correctly when there isn't a trailing comma in the import, though, which is why you see me always re-inserting it in my PRs. (It's also why you see the re-ordering happening in the first place, it's a few seconds of typing per import that's not properly sorted.)

yrobla (approved):

dfellis (approved):

PR #1159, opened 2022-08-26
Allocate a test account per integration test

dfellis:


.github/workflows/ci.yml
```diff
@@ -133,11 +130,32 @@ jobs:
       - name: Install dependencies
         run: yarn
 
+      - name: Determine Test Account To Use
+        id: determine-test-account
+        env:
+          SECRET_HEADER: ${{ secrets.SECRET_HEADER }}
+        run: |
+          ACCOUNT_INDEX=$(curl -X POST -H "Content-Type: text/plain" -H "${SECRET_HEADER}" -d '1' https://fkk3xlyad5tobhmezx4onpvele0zyehu.lambda-url.us-east-2.on.aws/ 2>/dev/null | jq '.[0]')
```
aguillenv, line 138: What's the `-d '1'`?

aguillenv (approved):
good to go for me after removing the `mwahahahahahaaaa`

depombo (approved):

dfellis:
> good to go for me after removing the `mwahahahahahaaaa`

Will do it once the tests pass, since it should prevent it from running the full test suite at that time.

depombo:
> > good to go for me after removing the `mwahahahahahaaaa`
> 
> Will do it once the tests pass, since it should prevent it from running the full test suite at that time.

Tests are passing!

.github/workflows/ci.yml
yrobla, line 138: maybe is worth to check result code and do some retries? in case that we get some timeout, temporary failure, etc?

dfellis: > maybe is worth to check result code and do some retries? in case that we get some timeout, temporary failure, etc?

I assumed if `curl` gave a non-zero result it would crap out here and basically not run the rest of the test because there's nothing to run. @aguillenv @depombo you guys are more familiar with how Github Actions works, am I right that a non-zero program return will error out this run or no?

aguillenv: Fair point, maybe adding the retry option to curl, I think it has it

yrobla (approved):

.github/workflows/ci.yml
dfellis: So, I wanted it to fail hard here, but it wasn't exactly, so following a suggestion from @yrobla to add `|| exit 1` to be absolutely sure the test exits if we can't allocate an account for it.

PR #1163, opened 2022-08-28
Use Express.js API instead of child process RPC

mtp1376:


src/services/scheduler-api.ts
```diff
@@ -1,46 +1,19 @@
...
-  return p;
+import fetch from 'node-fetch';
+import logger from '../services/logger';
+
+const schedulerAddress = 'http://localhost:14527';
+
+async function fetchOrRaise(url: string) {
+  const response = await fetch(url);
+  if (response.ok) {
+    return
```
dfellis, line 9: Should this be `return response;` instead?

mtp1376: On success, there's no response, just a `OK` text with 200 status code. Do we need the response object anyways?

dfellis: At the moment, no, so we can refactor later. It may just be surprising to write a new scheduler function that returns values back to the caller and they're inaccessible.

mtp1376: Good point. Changing it as you proposed.

mtp1376:
I reverted the changes on `shouldRpc` to keep using the local function invocation when testing. @dfellis Are we good to merge?

src/services/scheduler.ts
```diff
@@ -203,30 +204,52 @@ export async function init() {
...
-        .then(() => process.send?.(['stopAllComplete', ...args, undefined]))
-        .catch(e => process.send?.(['stopAllComplete', ...args, e.message]));
-      break;
+if (require.main === module) {
+  const app = express();
+  const port = 14527;
+
+  function respondErrorAndDie(res: any, err: any) {
+    res.status(500).send(err);
+    logger.error(`Scheduler exited with error: ${err}`);
```
depombo, line 213: we might want to port over the sentry logic so this, and other exceptions, get logged to sentry. that can happen in a follow up PR though

src/services/scheduler.ts
```diff
@@ -203,30 +204,52 @@ export async function init() {
...
+      .catch(e => respondErrorAndDie(res, e.message));
+  });
+
+  app.get('/stopAll/', async (req: any, res: any) => {
+    stopAll()
+      .then(() => res.sendStatus(200))
+      .catch(e => respondErrorAndDie(res, e.message));
+  });
+
+  app.head('/health/', async (req: any, res: any) => {
```
depombo, line 248: we can safely drop the `async` on these functions

depombo:
I really how simple this solution is over what was in place whilst fixing the underlying problem

src/services/scheduler.ts
mtp1376: Do we have a ticket for it? I can take it next if we do. If not, can you brief me on it so that I can go for it?

src/services/scheduler.ts
mtp1376: You're right. Removed them in the latest commit.

mtp1376:
Some considerations regarding local run of iasql-engine after the merge:
- When running iasql-engine on your local computer, you'll need to start an instance of `dist/services/scheduler.js` along with the main process. If you don't you'll get an error regarding the engine is unable to connect to the scheduler.
  - Same logic happening in the `docker-entrypoint.sh`

src/services/scheduler.ts
depombo: Yes, basically we want exceptions within the scheduler to also go to sentry. Currently, only the routes in the `db.ts` router get this treatment from the main Express server: https://github.com/iasql/iasql-engine/blob/main/src/index.ts. Basically we want to port the sentry setup logic from that file to this one. You can test that it works locally but setting the sentry logic in the config here:

> Some considerations regarding local run of iasql-engine after the merge:
> 
> * When running iasql-engine on your local computer, you'll need to start an instance of `dist/services/scheduler.js` along with the main process. If you don't you'll get an error regarding the engine is unable to connect to the scheduler.
>   
>   * Same logic happening in the `docker-entrypoint.sh`

Starting the docker container locally would spin it up for us, right?

depombo:
> Some considerations regarding local run of iasql-engine after the merge:
> 
> * When running iasql-engine on your local computer, you'll need to start an instance of `dist/services/scheduler.js` along with the main process. If you don't you'll get an error regarding the engine is unable to connect to the scheduler.
>   
>   * Same logic happening in the `docker-entrypoint.sh`

Starting the docker container locally would spin it up for us, right?

mtp1376:
> > Some considerations regarding local run of iasql-engine after the merge:
> > 
> > * When running iasql-engine on your local computer, you'll need to start an instance of `dist/services/scheduler.js` along with the main process. If you don't you'll get an error regarding the engine is unable to connect to the scheduler.
> >   
> >   * Same logic happening in the `docker-entrypoint.sh`
> 
> Starting the docker container locally would spin it up for us, right?

Yes if you use the docker container, it'll be OK. I added that comment just in case someone is running the server on their host machine (for debugging purposes, etc).

depombo:
Btw can we also update this section of the contributing documentation https://github.com/iasql/iasql-engine/blob/main/CONTRIBUTING.md#how-iasql-postgresql-functions-work with the new architecture in this PR? :)

mtp1376:
> Btw can we also update this section of the contributing documentation https://github.com/iasql/iasql-engine/blob/main/CONTRIBUTING.md#how-iasql-postgresql-functions-work with the new architecture in this PR? :)
> 
> It should be just removing the mention of RPC and instead that we spin up an ENTRYPOINT in the docker container which is just another express server where the schedulers live

I've added the descriptions you mentioned in the latest commit.

PR #1164, opened 2022-08-28
Optimize Dockerfile

mtp1376:
Old Dockerfile issues:
- We can make it faster by separating the copy for `package.json` and the codebase
  - https://www.ginkgobioworks.com/2020/05/18/optimizing-your-dockerfile/
  - Should not do `yarn install` again when the `package.json`/`yarn.lock` has not been changed
- Mounting the current directory makes it problematic if the docker container and the host machine’s OS are different(darwin vs linux)
   - The inner container installs the packages and overrides the `node_modules` folder on the host machine, making it problematic when trying to work on the host machine
    - Adding `node_modules` directory to the `.dockerignore` file fixes this

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,19 @@
 FROM node:16-bullseye
 WORKDIR /engine/
+
+COPY package.json /engine/package.json
+COPY yarn.lock /engine/yarn.lock
+RUN ["yarn", "install"]
+
+RUN ["apt", "update"]
+RUN ["apt", "upgrade", "-y"]
+RUN ["apt", "install", "postgresql-client-13", "-y"]
+
 COPY . /engine/
 ARG SENTRY_RELEASE
 ENV SENTRY_RELEASE=$SENTRY_RELEASE
 ARG IASQL_ENV
 ENV IASQL_ENV=$IASQL_ENV
-RUN apt update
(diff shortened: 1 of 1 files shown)


dfellis (approved):
It looks like the `.dockerignore` file also needs rebasing on latest `main`?

Dockerfile
```diff
@@ -1,14 +1,19 @@
 FROM node:16-bullseye
 WORKDIR /engine/
+
+COPY package.json /engine/package.json
+COPY yarn.lock /engine/yarn.lock
+RUN ["yarn", "install"]
+
+RUN ["apt", "update"]
+RUN ["apt", "upgrade", "-y"]
+RUN ["apt", "install", "postgresql-client-13", "-y"]
```
dfellis, line 10: I am curious why you went to the array-based syntax instead of the CLI-like syntax.

aguillenv (approved):

Dockerfile
depombo: +1

depombo (approved):

Dockerfile
mtp1376: I changed the entrypoint in another PR to use the array style (and I remember having problems with the normal style). So I also changed these to match that :D

PR #1166, opened 2022-08-28
Improved logging in CI

ajejoseph22:
I Improved logging in the CI. As @dfellis confirmed on Friday (26/08/2022), the CI would be aborted given a failure response from the Lambda (because of the `jq` query to which the response is piped). However, there is no way to view the error response from the Lambda (for debugging purposes) without logging into CloudWatch and digging - which could easily take minutes of development time.

I have simply added a way to log the error so we can view it in GH actions directly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,9 @@ jobs:
         env:
           SECRET_HEADER: ${{ secrets.SECRET_HEADER }}
         run: |
-          ACCOUNT_INDEX=$(curl -X POST -H "Content-Type: text/plain" -H "${SECRET_HEADER}" -d '1' https://fkk3xlyad5tobhmezx4onpvele0zyehu.lambda-url.us-east-2.on.aws/ 2>/dev/null | jq '.[0]' || exit 1)
+          RESPONSE=$(curl -X POST -H "Content-Type: text/plain" -H "${SECRET_HEADER}" -d '1' https://fkk3xlyad5tobhmezx4onpvele0zyehu.lambda-url.us-east-2.on.aws/ 2>/dev/null)
+          echo $RESPONSE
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -135,7 +135,9 @@ jobs:
         env:
           SECRET_HEADER: ${{ secrets.SECRET_HEADER }}
         run: |
-          ACCOUNT_INDEX=$(curl -X POST -H "Content-Type: text/plain" -H "${SECRET_HEADER}" -d '1' https://fkk3xlyad5tobhmezx4onpvele0zyehu.lambda-url.us-east-2.on.aws/ 2>/dev/null | jq '.[0]' || exit 1)
+          RESPONSE=$(curl -X POST -H "Content-Type: text/plain" -H "${SECRET_HEADER}" -d '1' https://fkk3xlyad5tobhmezx4onpvele0zyehu.lambda-url.us-east-2.on.aws/ 2>/dev/null)
+          echo $RESPONSE
```
aguillenv, line 139: if the `curl` failed, would still be saved in `RESPONSE`? I don't remember if it will fail and not continue if we do not add some silent fail property

ajejoseph22: Well, `2>/dev/null` is a silent fail mechanism. This silences errors that `curl` considers errors e.g timeouts, wrong URL etc.

In the event of a lambda-specific error, like when we pass a wrong secret, the lambda returns "Unauthorized", with a `401` code. As far as `curl` is concerned, this is not a failure. This is not silenced by `2>/dev/null` and it also doesn't break the CI (although CI will break when `jq` tries to read `.[0]` property from "Unauthorized" and exits with `1` code)

Screen Shot 2022-08-29 at 16 52 00

aguillenv (approved):

PR #1185, opened 2022-08-30
AWS policy and object diff

ajejoseph22:


src/util/common.ts
```diff
@@ -0,0 +1,11 @@
...
+  return Array.isArray(obj) && obj.every(isString);
+}
+
+export function isString(obj: unknown): obj is string {
+  return typeof obj === 'string';
+}
+
+export function isObject(obj: unknown): obj is object {
+  return typeof obj === 'object';
+}
```
dfellis, line 11: Nit: Does it make sense to pull this functions into a `common` file when there is a singular usage of them? (eg, if you have a second usage you think they should go to soon, then I'm okay with this)

src/util/policy-diff.ts
```diff
@@ -0,0 +1,89 @@
...
+      case IAM_KEY.Statement:
+        same = policiesAreSame(obj1[key], obj2[key]);
+        break;
+      // Can be JSON or string
+      case IAM_KEY.Principal:
+      // Can be string or array of strings
+      case IAM_KEY.Action:
+      case IAM_KEY.Resource:
+      // Can only be JSON
+      case IAM_KEY.Condition:
```
dfellis, line 73: This logic is IAM-specific. Are we sure there will be no keyname collisions with other usages in the future?

src/modules/0.0.19/aws_ecr/index.ts
```diff
@@ -9,6 +9,7 @@ import { AWS, crudBuilder2, crudBuilderFormat, paginateBuilder } from '../../../
 import logger from '../../../services/logger';
 import { Context, Crud2, MapperBase, ModuleBase } from '../../interfaces';
 import { PublicRepository, Repository, RepositoryPolicy, ImageTagMutability } from './entity';
+import policiesAreSame from '../../../util/policy-diff';
```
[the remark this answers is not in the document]
ajejoseph22: I guess they are generally the same. I will refactor these into the `services` folder for uniformity.

yrobla (commented):
I can explain one difference i saw, not for policies but for caching documents:

Entry one: {Quantity: 0, Items: undefined}
Entry two: {Quantity: 0}

AWS seem to be dropping the mapped objects when there aren't any. Maybe this can be done in a follow-up, but it's a case that i'm seeing on objects returned by AWS.

src/util/common.ts
ajejoseph22: My vision is for this to be used across all the files where we currently have this repetition

image

ajejoseph22: For `isObject`

image

I guess the only one with singular use case is `isStringArray`. I will extract this one to the `services/policy-diff.ts` file

dfellis (approved):

ajejoseph22:
> I can explain one difference i saw, not for policies but for caching documents:
> 
> Entry one: {Quantity: 0, Items: `undefined`} Entry two: {Quantity: 0}
> 
> AWS seem to be dropping the mapped objects when there aren't any. Maybe this can be done in a follow-up, but it's a case that i'm seeing on objects returned by AWS.

Can you please share a reference to the code in `iasql-engine`? @yrobla

aguillenv (approved):

src/services/object-diff.ts
```diff
@@ -0,0 +1,99 @@
+import { isString } from './common';
```
depombo, line 1: I would probably call this `aws-diff.ts` since we will eventually add more cloud providers and this is only a thing for AWS

test/common/object-diff.ts
```diff
@@ -0,0 +1,297 @@
+import {objectsAreSame, policiesAreSame} from "../../src/services/object-diff";
+
+jest.setTimeout(30000);
+
+describe("policiesAreSame", () => {
+    describe("when cloud and DB policies are exactly the same", () => {
```
depombo, line 6: super nit - I think all other files have a tab space of 2 and this one has 4 spaces. it would be cool to enforce that in the style, but all super minor

depombo (approved):

dfellis:
https://github.com/iasql/iasql-engine/runs/8137146333?check_suite_focus=true#step:6:65773

The RDS test suite has been very stable in the past, and these failures are all related to parameter groups. I think there might be something not quite right here?

ajejoseph22:
> https://github.com/iasql/iasql-engine/runs/8137146333?check_suite_focus=true#step:6:65773
> 
> The RDS test suite has been very stable in the past, and these failures are all related to parameter groups. I think there might be something not quite right here?

Checking

ajejoseph22:
> > https://github.com/iasql/iasql-engine/runs/8137146333?check_suite_focus=true#step:6:65773
> > The RDS test suite has been very stable in the past, and these failures are all related to parameter groups. I think there might be something not quite right here?
> 
> Checking

@dfellis Found the issue while inspecting the test logs. 
It seems I forgot to cover the case where one object has `<some_key>: undefined` while the other object doesn't have `<some_key>` at all. AWS was returning `undefined` values for some keys in the DB parameter objects. I fixed it now and it works locally.
image

ajejoseph22:
> 

Tests passed remotely too. Merging.

PR #1193, opened 2022-08-31
Rework the cleanup workflow

dfellis:


yrobla (approved):

.github/workflows/scheduled-cleanup.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - setup-module-cleanup
     uses: iasql/iasql-engine/.github/workflows/module-cleanup.yml@main
     with:
-      modules: ${{ needs.setup-module-cleanup.outputs['set-test-module-names'] }}
+      account_ids: ${{ needs.setup-module-cleanup.outputs['set-test-account-ids'] }}
       regions: "['ap-northeast-1', 'ap-northeast-2', 'ap-northeast-3', 'ap-south-1', 'ap-southeast-1', 'ap-southeast-2']"
```
aguillenv, line 37: ok, if I'm not wrong, this approach will not work, because we will be generating `nRegions x 100` jobs and in all cases will be higher than the 256 jobs limit. we might need to reduce the matrices, maybe we will need to create n cleanup steps with 2 regions each 🤔

dfellis: I was worried about that. And 100 test accounts is only the limit *for now*. We may increase that in the future and that may fail unexpectedly.

How about instead of a `regions` matrix at that level, it's just a `region` property, and we have an intermediate matrix splitting all regions into a per-region job?

aguillenv: that might work, creating here the matrix per region and modifying the `module-cleanup.yaml` to receive only a region.

dfellis: I decided to just devolve it into explicit tasks per region, just in case `matrix of matrix` counts towards the 256 limit, since there aren't that many regions.

Issue #1194, opened 2022-08-31
Elasticache module fails due to invalid node type


yrobla:
In the tests for elasticache we are choosing a node type that doesn't seem to be present in all regions, causing the test to fail.
We need to query for available node types on the chosen region and use on the the available ones.

PR #1195, opened 2022-08-31
Issue #1194: Select cache node type from an available list

yrobla:
Instead of hardcoding the node type, query the list of
available nodes for this type, and pick the first one on
the list.

Closes: #1194

### DIFF
--- a/test/modules/aws-elasticache-integration.ts
+++ b/test/modules/aws-elasticache-integration.ts
@@ -1,5 +1,7 @@
-import config from "../../src/config";
-import * as iasql from "../../src/services/iasql";
+import { ElastiCache } from '@aws-sdk/client-elasticache';
+
+import config from '../../src/config';
+import * as iasql from '../../src/services/iasql';
 import {
   runQuery,
   runInstall,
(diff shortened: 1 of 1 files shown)


aguillenv (approved):
eventually, it might make sense to have a node type table by region and/or availability zone

yrobla:
> eventually, it might make sense to have a node type table by region and/or availability zone

I thought about it, but in this case this availability is dynamic. It means that you only can select the node types that are available for that region in that period of time, and it can change dynamically. I think that it will make sense to add this support from the module itself, allowing to retrieve a list of available nodes, but it sounds like a future improvement.

dfellis (approved):

depombo (approved):

depombo:
> > eventually, it might make sense to have a node type table by region and/or availability zone
> 
> I thought about it, but in this case this availability is dynamic. It means that you only can select the node types that are available for that region in that period of time, and it can change dynamically. I think that it will make sense to add this support from the module itself, allowing to retrieve a list of available nodes, but it sounds like a future improvement.

Absolutely. Created an issue to track this https://github.com/iasql/iasql-engine/issues/1198

PR #1199, opened 2022-09-01
add new bootstrap configuration

depombo:


aguillenv (approved):

dfellis (approved):

docker-entrypoint.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-yarn forever start dist/services/scheduler.js
+yarn forever dist/services/scheduler.js
```
depombo, line 2: this makes it so the process is not started in a daemon such that logs still get displayed

src/config/bootstrap.ts
```diff
@@ -0,0 +1,32 @@
...
+  },
+  logger: {
+    debug: true,
+    test: false,
+  },
+  cors: {
+    origin: 'http://localhost:3000',
+  },
+  telemetry: {
+    amplitudeKey: '8fd6aa1d61c115e59e35b3adfd5dd41a',
```
yrobla, line 28: is that safe to set the key here? or better add it as a secret?

dfellis: Amplitude keys end up in website source code, so it's not really a secret.

yrobla (approved):

PR #1205, opened 2022-09-02
multi region account support

dfellis:
- Initial refactoring of the aws_account module
- Change hardwired aws_account table name in here (todo: don't do this at all in the future)
- Rework two iasql functions with hardwired aws_account table setups to handle the new form
- Refactor the test suite
- Regen migration file
- the `default_aws_region` Postgres function that multi-region-aware modules can use for their region column's default value.
- new tests


src/modules/0.0.19/aws_account/index.ts
```diff
@@ -36,35 +38,122 @@ class AccountMapper extends MapperBase<AwsAccountEntity> {
...
+      }
+      return (await this.getRegions(client.ec2client)).map(r => {
+        const awsRegion = new AwsRegions();
+        awsRegion.isDefault = false;
+        awsRegion.isEnabled = true;
+        awsRegion.region = r;
+        return awsRegion;
+      });
+    },
+    update: async (_e: AwsRegions[], _ctx: Context) => {
```
yrobla, line 83: in the create/update, could we manage/restrict somehow just having a single default region? and how do we manage the flow in case that default region is not enabled?

src/modules/0.0.19/aws_account/index.ts
```diff
@@ -36,35 +38,122 @@ class AccountMapper extends MapperBase<AwsAccountEntity> {
...
+      // The only controllable fields here have nothing to do with AWS itself, but for how IaSQL works
+      // with multiple regions, so we can literally no-op this one.
+    },
+    delete: async (e: AwsRegions[], ctx: Context) => {
```
yrobla, line 87: could we delete the default region? or shall we error in that case?

yrobla (commented):
comments on review

src/modules/0.0.19/aws_account/entity/index.ts
```diff
@@ -17,7 +13,28 @@ export class AwsAccountEntity {
...
+}
+
+@Entity()
+export class AwsRegions {
+  @PrimaryGeneratedColumn()
+  id: number;
 
   @Column()
+  @cloudId
   region: string;
```
aguillenv, line 25: I think the `id` might need to be just autogenerated number but not `primary` and let the region column be the PK if we want to join by the region name

test/modules/aws-acm-list-integration.ts
```diff
@@ -24,10 +24,16 @@ describe('AwsAcmList Integration Testing', () => {
...
-    INSERT INTO aws_account (region, access_key_id, secret_access_key)
-    VALUES ('${process.env.AWS_REGION}', '${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}')
+    INSERT INTO aws_credentials (access_key_id, secret_access_key)
+    VALUES ('${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}')
   `, undefined, false));
 
+  it('syncs the regions', sync());
+
+  it('sets the default region', query(`
+    UPDATE aws_regions SET is_default = TRUE WHERE region = '${process.env.AWS_REGION}';
```
aguillenv, line 34: does this work if thers already a `is_defualt = TRUE`? I think at this point we will have 2 regions as default in the regions table maybe this query will need to be something like:
```sql
update aws_regions
set is_default = case is_default 
    when FALSE then TRUE
    else FALSE
    end
where region = '${process.env.AWS_REGION}' OR is_default = TRUE;
```

Have not tested it but something like that should work.

src/modules/0.0.19/aws_account/index.ts
dfellis: So right now the code has a final fallback to `us-east-1` if the default isn't set.

src/modules/0.0.19/aws_account/index.ts
dfellis: We have several tables that work this way, especially for the `_metadata` modules. It's not super clear how else to handle tables that are not really mutable by the end users?

src/modules/0.0.19/aws_account/entity/index.ts
dfellis: We can still add a secondary index on this to allow a join, can't we? I'm pretty sure you're allowed to have a foreign key relation onto a secondary index.

test/modules/aws-acm-list-integration.ts
aguillenv: I see your point. I was thinking further by the time you add the constraint to always having only one `is_default`, by the time you `sync` you will need to fix that is_default to a fallback value, and then I guess you will have to do something like what I've suggested maybe.

I know this is WIP, I'm just thinking out loud.

dfellis: Oh, on that point I plan to have a `default_aws_region` postgres function in this PR. If called with no arguments it returns the current default region (or it automatically sets `us-east-1` as the default region for you and returns that), while if you call it with a string argument, it changes the default region to the one you provided and then returns that region back to you (or errors if the string provided is not a region).

test/modules/aws-account-integration.ts
```diff
@@ -121,13 +129,38 @@ describe('AwsAccount Integration Testing', () => {
...
+  it('tries to set a second default region', query(`
+    UPDATE aws_regions SET is_default = TRUE WHERE region = 'us_east_1';
+  `));
+
+  it('confirms that the default region was not changed', query(`
+    SELECT * FROM aws_regions WHERE is_default = TRUE;
+  `, (res: any[]) => {
+    expect(res.length).toBe(1);
+    expect(res[0].region).toBe(process.env.AWS_REGION);
+  }));
```
aguillenv, line 155: so we cannot have 2 default regions but we could have the case where none of the regions is the default one, right? In that case the postgres fn will return `us-east-1`

test/modules/aws-account-integration.ts
```diff
@@ -121,13 +129,38 @@ describe('AwsAccount Integration Testing', () => {
...
+
+  it('updates the default region with the handy `default_aws_region` function', query(`
+    SELECT * FROM default_aws_region('us_east_1');
```
aguillenv, line 158: nit: maybe it makes sense to use another region rather than the fallback value to ensure it is working?

test/readonly/test.ts
```diff
@@ -29,10 +29,16 @@ describe('Aws read only Integration Testing', () => {
...
-    VALUES ('us-east-1', '${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}')
+    INSERT INTO aws_credentials (access_key_id, secret_access_key)
+    VALUES ('${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}')
   `, undefined, false));
 
+  it('syncs the regions', sync());
+
+  it('sets the default region', query(`
+    UPDATE aws_regions SET is_default = TRUE WHERE region = 'us-east-1';
+  `));
```
aguillenv, line 40: technically all these tests could be changed to the following, right?
```suggestion
  it('sets the default region', query(`
    SELECT * FROM default_aws_region('us-east-1');
  `));
```

test/modules/aws-account-integration.ts
aguillenv: also, the regions are not supposed to be `us-east-1` instead of `us_east_1`?

test/readonly/test.ts
dfellis: Yes, they should be equivalent. Do you think it'd be better to make that change?

src/modules/0.0.19/aws_account/entity/index.ts
dfellis: I'll switch this, anyways.

yrobla (approved):

src/modules/0.0.19/aws_account/entity/index.ts
```diff
@@ -1,13 +1,9 @@
...
 import { cloudId } from '../../../../services/cloud-id';
 
-// This is ridiculous. Can we fix this?
-
-@Entity({
-  name: 'aws_account',
-})
-export class AwsAccountEntity {
+@Entity()
+export class AwsCredentials {
```
depombo, line 6: TODO on constraint that this table can only have one row?

aguillenv (approved):

src/services/iasql.ts
```diff
@@ -285,7 +285,10 @@ export async function dump(dbId: string, dataOnly: boolean) {
   const dbMeta = await MetadataRepo.getDbById(dbId);
   if (dbMeta?.upgrading) throw new Error('Currently upgrading, cannot dump this database');
   const pgUrl = dbMan.ourPgUrl(dbId);
-  const excludedDataTables = "--exclude-table-data 'aws_account' --exclude-table-data 'iasql_*'";
```
depombo, line 288: if we support snapshotting it might make sense get rid of export since we are not encouraging migrations systems anymore

test/modules/aws-account-integration.ts
```diff
@@ -121,13 +129,47 @@ describe('AwsAccount Integration Testing', () => {
...
+
+  it('confirms that the default region was not changed', query(`
+    SELECT * FROM aws_regions WHERE is_default = TRUE;
+  `, (res: any[]) => {
+    expect(res.length).toBe(1);
+    expect(res[0].region).toBe(process.env.AWS_REGION);
+  }));
+
+  it('updates the default region with the handy `default_aws_region` function', query(`
+    SELECT * FROM default_aws_region('us-east-1');
```
depombo, line 167: I think all the docs changes are coming on later on, but https://docs.iasql.com/aws/ should include the new UI screen and this function?

test/modules/aws-acm-list-integration.ts
depombo: once we generalize cloud APIs (https://github.com/iasql/iasql-engine/issues/1152) it might make sense to provide credentials on the very first function call to `install_provider('aws', '*****', '******')` such that regions are filled in from the start?

src/modules/0.0.19/aws_account/entity/index.ts
```diff
@@ -17,7 +13,25 @@ export class AwsAccountEntity {
 
   @Column()
   secretAccessKey: string;
+}
 
-  @Column()
+@Entity()
+export class AwsRegions {
```
depombo, line 19: I thought this would be an enum, but this makes it easier to maintain

depombo (approved):

test/modules/aws-account-integration.ts
dfellis: No UI screen changes in the other PR, so I think we're still good?

depombo: Right, but there are paired PRs? We should update the docs with the new screen that let users select the region is what I mean. I guess it can be another PR other than this one, that is fine.

PR #1206, opened 2022-09-04
Allow multiple hosted zones on the same domain name

mtp1376:


mtp1376:
Trying to debug the failing tests on route53. The strange part is not why they're failing now, but why they were passing before.

yrobla:
looks good to me although i miss a more descriptive commit message, it was difficult for me to understand the reasoning without looking carefully

mtp1376:
Regarding failed tests:
- `tries to update a hosted zone domain name field (replace)`: tries to run this query:
```UPDATE hosted_zone SET domain_name = '${replaceDomainName}' WHERE domain_name = '${domainName}';```
I believe this should not be possible and should return error (as it is). The AWS console does not also provide the user with option to change the domain name of their hosted zone. The only question I have is why it's not throwing errors on other CI runs. Currently investigating the previous behavior.
image

**Comment by aguillenv** [comment]
> Regarding failed tests:
> 
> * `tries to update a hosted zone domain name field (replace)`: tries to run this query:
> 
> ```sql
> UPDATE hosted_zone SET domain_name = '${replaceDomainName}' WHERE domain_name = '${domainName}';
> ```
> 
> I believe this should not be possible and should return error (as it is). The AWS console does not also provide the user with option to change the domain name of their hosted zone. The only question I have is why it's not throwing errors on other CI runs, because the connected resource record sets all have the domain name in them. Currently investigating the previous behavior. image

With IaSQL we would like to be able to perform any change a user does to the database without fail. So for those cases where AWS does not let you modify/update, we fully replace the resource, meaning, when possible, create a new resource with the same db information in AWS and then delete the old one. In some cases, when AWS requires a property that needs to be unique, we use other strategies like first deleting de old one and then creating the new one (but this can produce downtimes).

For the case you are trying to debug, take a look at the replacement logic here https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.19/aws_route53_hosted_zones/index.ts#L135L160

File: src/services/scheduler.ts [hunk 1]
```diff
@@ -350,7 +350,11 @@ if (require.main === module) {
...
-  app.listen(port, () => {
-    logger.info(`Scheduler running on port ${port}`);
-  });
+  init()
+    .then(() => {
+      app.listen(port, () => {
+        logger.info(`Scheduler running on port ${port}`);
+      });
+    })
+    .catch(e => respondErrorAndDie(undefined, e));
```
dfellis, line 359: We had to revert the change related to this before because it caused unexpected issues in staging/production. Can we move the scheduler changes to a separate PR in case this happens again?

src/modules/0.0.21/aws_route53_hosted_zones/index.ts
```diff
@@ -346,17 +368,6 @@ class ResourceRecordSetMapper extends MapperBase<ResourceRecordSet> {
       for (const e of es) {
         const cloudRecord = ctx?.memo?.cloud?.ResourceRecordSet[this.module.resourceRecordSet.entityId(e)];
         // First check if theres a new hosted zone, the name need to change since it is based on the domain name
```
aguillenv, line 370: maybe this comment needs to be removed too?

aguillenv (commented):
Looks good to me. Just a minor comment.

dfellis (approved):

depombo (approved):

aguillenv (approved):

PR #1215, opened 2022-09-05
chore: added support for undefined policy comparisons

ajejoseph22:


yrobla (approved):

aguillenv (approved):

src/services/aws-diff.ts
```diff
@@ -55,6 +55,9 @@ export function objectsAreSame(obj1: any = {}, obj2: any = {}): boolean {
 
 // Returns true if the policies mean the same thing (regardless of structure), and false if they aren't
 export function policiesAreSame(obj1: any, obj2: any): boolean {
+  // comparison for when one or both values are undefined
+  if (!obj1 || !obj2) return obj1 === obj2;
```
dfellis, line 59: Screenshot from 2022-09-05 10-22-25

This may revive the `null` vs `undefined` comparison issues we've had before.

PR #1221, opened 2022-09-05
fix: take cert from stdout if it exist

### LINKED ISSUE #1220: privkey generation fails on local tests
Modules like ELB rely on private key generation functions (on helpers.ts file).
There is an option for local environments, that just generates a key on the fly, and for other envs it just tries to read from a file, crashing if the file does not exist.
We can modify the test, to try reading the key from stdout. And if it does not exist, fallback to existing privkey file

yrobla:
In order to generate a private key, first try to take
the content from the generated openssl command. If it
does not exist, fallback to privkey content

Closes #1220

### DIFF
--- a/test/helpers.ts
+++ b/test/helpers.ts
@@ -1,10 +1,9 @@
-import { execSync, } from 'child_process'
-import { v4 as uuidv4, } from 'uuid'
+import { execSync } from 'child_process';
+import { createConnection } from 'typeorm';
+import { v4 as uuidv4 } from 'uuid';
 
-import { createConnection, } from 'typeorm'
-
-import MetadataRepo from '../src/services/repositories/metadata'
-import logger from '../src/services/logger'
+import logger from '../src/services/logger';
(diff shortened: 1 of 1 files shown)


yrobla:
there is a big difference, actually caused by `prettier` formatting.

dfellis (approved):

aguillenv (approved):

dfellis:
Unfortunately it seems this PR doesn't actually work on Linux :( https://github.com/iasql/iasql-engine/runs/8194647676?check_suite_focus=true#step:6:9874

dfellis:
BTW, why not invert the check? If there's a `.pem` file generated, use it, otherwise use the stdout value?

yrobla:
> BTW, why not invert the check? If there's a `.pem` file generated, use it, otherwise use the stdout value?

makes sense, it will be simpler

aguillenv (approved):

PR #1222, opened 2022-09-05
fix: use an available node type for elasticache update test

yrobla:
Not all node types are available to be used on elasticache,
and they need to be taken from an api. While the test was using
this workflow for creating the cache, it was not using it
for updating it.
Modify the tests to pick a valid node as well for that test.

### DIFF
--- a/test/modules/aws-elasticache-integration.ts
+++ b/test/modules/aws-elasticache-integration.ts
@@ -136,11 +136,10 @@ describe('Elasticache Integration Testing', () => {
   SELECT * FROM cache_cluster WHERE cluster_id='${clusterId}' AND node_type='${updatedNodeType}';
 `,
       (res: any) => expect(res.length).toBe(1),
-    ),
-      (e?: any) => {
-        if (!!e) return done(e);
-        done();
-      };
+    )((e?: any) => {
+      if (!!e) return done(e);
+      done();
+    });
   });
 
   it(
--- a/test/modules/aws-elasticache-integration.ts
+++ b/test/modules/aws-elasticache-integration.ts
@@ -39,14 +39,12 @@ const elasticacheclient = new ElastiCache({
 });
 
 const getAvailableNodeTypes = async () => {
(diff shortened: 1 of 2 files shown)


test/modules/aws-elasticache-integration.ts
```diff
@@ -38,26 +38,31 @@ const elasticacheclient = new ElastiCache({
...
 
 jest.setTimeout(620000);
 beforeAll(async () => {
   // just sleep
-  nodeType = await getAvailableNodeType();
+  const nodes = await getAvailableNodeTypes();
+  nodeType = nodes.pop() ?? '';
+  updatedNodeType = nodes.pop() ?? '';
   await execComposeUp();
 }, 60000);
```
aguillenv, line 67: this timeout seems off, I do not see it in other tests

yrobla: this came from experimentation with async. I guess it could be removed. The main issue i had is that those values were not recognized, just were taken as `undefined` . The `done` trick made them work, and values properly picked

test/modules/aws-elasticache-integration.ts
```diff
@@ -114,24 +119,28 @@ describe('Elasticache Integration Testing', () => {
...
 
-  it(
-    'checks that cache_cluster have been modified',
+  it('checks that cache_cluster have been modified', done => {
     query(
       `
-  SELECT * FROM cache_cluster WHERE node_type='cache.t2.small';
+  SELECT * FROM cache_cluster WHERE cluster_id='${clusterId}' AND node_type='${updatedNodeType}';
 `,
       (res: any) => expect(res.length).toBe(1),
```
dfellis, line 138:
```suggestion
      (res: any) => expect(res.length).toBe(1)
```

Can you delete this trailing comma here? I suspect JS shenanigans.

dfellis: Nevermind on this, I am wrong.

dfellis (approved):

aguillenv (approved):

PR #1224, opened 2022-09-06
Upgrade Postgres to 14.4

dfellis:
This PR is to test whether or not things break moving to Postgres 14. It is expected to be a drop-in replacement, but we'll soon see.

### DIFF
--- a/test/modules/aws-rds-integration.ts
+++ b/test/modules/aws-rds-integration.ts
@@ -36,7 +36,7 @@ describe('RDS Integration Testing', () => {
   it('creates an RDS instance', query(`
     BEGIN;
       INSERT INTO rds (db_instance_identifier, allocated_storage, db_instance_class, master_username, master_user_password, availability_zone, engine, backup_retention_period)
-        VALUES ('${prefix}test', 20, 'db.t3.micro', 'test', 'testpass', '${availabilityZone}', 'postgres:13.4', 0);
+        VALUES ('${prefix}test', 20, 'db.t3.micro', 'test', 'testpass', '${availabilityZone}', 'postgres:14.4', 0);
       INSERT INTO rds_security_groups (rds_id, security_group_id) SELECT
         (SELECT id FROM rds WHERE db_instance_identifier='${prefix}test'),
         (SELECT id FROM security_group WHERE group_name='default');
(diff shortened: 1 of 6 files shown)


dfellis:
https://github.com/iasql/iasql-engine/runs/8211468751?check_suite_focus=true#step:6:24418 Oops. Fixing that.

depombo (approved):

test/modules/aws-rds-integration.ts
```diff
@@ -61,7 +61,7 @@ describe('RDS Integration Testing', () => {
   it('creates an RDS instance', query(`
     BEGIN;
       INSERT INTO rds (db_instance_identifier, allocated_storage, db_instance_class, master_username, master_user_password, availability_zone, engine, backup_retention_period)
-        VALUES ('${prefix}test', 20, 'db.t3.micro', 'test', 'testpass', '${availabilityZone}', 'postgres:13.4', 0);
+        VALUES ('${prefix}test', 20, 'db.t3.micro', 'test', 'testpass', '${availabilityZone}', 'postgres:14.1', 0);
```
yrobla, line 64: why 14.1 here? aren't we upgrading to 14.4 ?

yrobla (changes requested):

test/modules/aws-rds-integration.ts
aguillenv: This is the test for RDS integration and RDS does not let you upgrade from 13.4 to 14.4, I think they only allow smaller gaps for engine updates, so he is creating the first RDS instance with 14.1 and updating to 14.4 in line 85.

aguillenv (approved):

yrobla (approved):
thanks for the explanation, looking good then

test/modules/aws-rds-integration.ts
dfellis: That's right, but it also brings up a wrinkle in us moving staging and production over to Postgres 14, which is: how to convince RDS to do it. We may have to make a snapshot and restore that snapshot into a new RDS instance, then swap the `db.iasql.com` and similar domains over, but that means we need explicit downtime to make the jump, which isn't great.

dfellis:
Will soon be completely irrelevant, but would need to be mostly redone, anyways.

Issue #1192, opened 2022-08-31
Implement custom functions RFC


depombo:

aguillenv:
Creation phase:
- [ ] Add the new `iasql_rpc` entity and scheduler task
- [ ] Add the `rpc` type and behaviours to the Module interfaces
- [ ] Add the `sql` type and behaviours to the Module interfaces
- [ ] Copy the `iasql.ts` functions into the `iasql_functions` module, drop the `iasql_operation`-based handwritten functions, and expose the new ones via RPC
- [ ] Port all hand-edited migration files to use the new `sql` sub-object instead
- [ ] Add `iasql_rpc` auto-vacuum

Cleanup phase (~1 month after creation phase is done):
- [ ] Drop duplicated functions out of `iasql.ts`
- [ ] Drop the old scheduler `operation` logic

PR #1225, opened 2022-09-06
Add `iasql_rpc` + scheduler job

aguillenv:
Part of #1192 

image


src/services/scheduler.ts
```diff
@@ -147,6 +147,71 @@ export async function start(db: IasqlDatabase) {
...
+            //       recordCount,
+            //       operationCount,
+            //     },
+            //     {
+            //       params,
+            //       output,
+            //       error,
+            //     },
+            //     uid,
+            //   );
```
dfellis, line 209: Copy-pasted from the current handler?

src/services/scheduler.ts
```diff
@@ -147,6 +147,71 @@ export async function start(db: IasqlDatabase) {
           }
         }
       },
+      rpc: async (payload: any) => {
+        const { params, opid, moduleName, methodName } = payload;
+        let promise;
```
dfellis, line 152: I presume this is where the RPC call is gonna go based on `moduleName`, `methodName`, and `params` above. It won't need to be a `let` variable in here :)

aguillenv: yes, I still have to think exactly how it will look but yes, technically it will not be necessary, just looking for the method in the module and calling it.

src/services/scheduler.ts
aguillenv: yes, I added a todo to refactor with specific logic to this case, we will still need to handle the LIST case

src/services/scheduler.ts
```diff
@@ -146,6 +146,76 @@ export async function start(dbId: string, dbUser: string) {
...
+          }
+          output = await Modules[modulename][methodname](params);
+          // once the rpc completes updating the `end_date`
+          // will complete the polling
+          const query = `
+            update iasql_rpc
+            set end_date = now(), output = '${output}'
+            where opid = uuid('${opid}');
+          `;
+          logger.debug(query);
```
yrobla, line 172: is it a leftover or you want to keep it for testing purposes?

yrobla (commented):
small nit

src/modules/0.0.19/iasql_functions/index.ts
```diff
@@ -8,5 +8,9 @@ class IasqlFunctions extends ModuleBase {
     super.init();
   }
   iasqlOperationType = IasqlOperationType;
+  // TODO: here for testing purpose. To be delete it
+  customCall = async () => {
```
dfellis, line 12: How are parameters passed in here?

src/services/scheduler.ts
```diff
@@ -147,17 +147,21 @@ export async function start(dbId: string, dbUser: string) {
...
+        const dbAlias = user?.iasqlDatabases?.[0]?.alias;
         try {
-          output = await promise;
+          const versionString = await TypeormWrapper.getVersionString(dbId);
+          const Modules = (modules as any)[versionString];
+          if (!Modules[modulename]) throw new Error(`Module ${modulename} not found`);
+          if (!Modules[modulename][methodname]) {
+            throw new Error(`Method ${methodname} in module ${modulename} not found`);
+          }
+          output = await Modules[modulename][methodname](params);
```
dfellis, line 164: Maybe it would make sense to do:

```suggestion
          output = await Modules[modulename][methodname](...params);
```

So arguments can be separately named?

src/modules/0.0.19/iasql_functions/index.ts
dfellis: I found out below, as a single `string[]` argument. It might be better to `...` it (as suggested below) for more natural looking functions.

dfellis: Also, can you confirm that this form also works?

```suggestion
  async customCall() {
```

It *should*, but I'd prefer confirmation since that's the only way to get access to `this` in these custom functions.

src/services/scheduler.ts
dfellis: Oh, another thing. It may make sense to get a `Context` object here and pass it through, or a custom RPC function that needs to work with the CRUD functions will have to build one itself.

This is what `apply` and `sync` do, but I think the pattern may be common enough that passing in a context object as the first argument could make sense in general?

src/services/scheduler.ts
aguillenv: So the structure is similar to the current `operation` and that debug log was there. I think it would be easier to debug in the case is needed having that log there.

src/modules/0.0.19/iasql_functions/index.ts
aguillenv: I can test it, I also think it should work, but let you know. Also, I expect this to change once the `rpc` subtype is in place so was not so thoughtful on the method definition.

src/services/scheduler.ts
aguillenv: Ahh true!! let me see how I can get it and pass it.

src/modules/0.0.19/iasql_functions/index.ts
dfellis: Right. We need to think about that. The reasoning behind it was to prevent users from manually calling an RPC on a function not meant to be an RPC, but it will probably need to be instantiated like the `extends BaseMapper` classes in order to have access to the module's `this` property and be able to do non-trivial things.

src/services/scheduler.ts
aguillenv: Thinking about it, I've removed it. that query could be seen in the postgres container logs.

src/modules/0.0.19/iasql_functions/index.ts
aguillenv: It worked

src/services/scheduler.ts
aguillenv: Already passing params and context. Not sure if the way I'm passing the context is the right one, though. Please, let me know what you think!

yrobla (approved):

src/services/scheduler.ts
```diff
@@ -161,15 +161,15 @@ export async function start(dbId: string, dbUser: string) {
           if (!Modules[modulename][methodname]) {
             throw new Error(`Method ${methodname} in module ${modulename} not found`);
           }
-          output = await Modules[modulename][methodname](params);
+          const moduleCtx = (Modules[modulename] as ModuleInterface).provides?.context;
```
dfellis, line 164: This is unfortunately not everything needed. The `provides.context` of all installed modules need to be mixed together. I am pretty sure that logic lives in the apply and sync functions.

aguillenv: ok, I think my latest update should do the work!

depombo (approved):

yrobla (approved):

PR #1226, opened 2022-09-06
pick region randomly per module

aguillenv:


dfellis (approved):
This looks GTM for now. We can eventually figure out per-module region restrictions, but this will allow rerunning individual tests that get stuck on an unsupported region.

yrobla (approved):

PR #1228, opened 2022-09-07
Revert "Refactor Scheduler functions and add Sentry error logging (#1182)"

depombo:
This reverts commit c3262f011938bf8628cf64aad6ce9af1c0e9ae7a.


depombo:
so sentry actually already works without https://github.com/iasql/iasql-engine/pull/1182, I think I mistakenly thought that wasn't the case
we properly init sentry https://github.com/iasql/iasql-engine/blob/main/src/services/scheduler.ts#L187
and log to sentry https://github.com/iasql/iasql-engine/blob/main/src/services/scheduler.ts#L109. my bad!

dfellis (approved):

PR #1235, opened 2022-09-07
Custom sql outside of migrations

aguillenv:
Part of #1192


src/modules/0.0.19/iasql_functions/index.ts
```diff
@@ -1,10 +1,15 @@
...
+import { readFileSync } from 'fs';
 import { ModuleBase } from '../../interfaces';
 import { IasqlOperationType } from './entity';
 
 class IasqlFunctions extends ModuleBase {
   constructor() {
     super();
+    this.sql = {
+      afterInstall: readFileSync(`${__dirname}/sql/create_fns.sql`, 'utf8'),
+      beforeUninstall: readFileSync(`${__dirname}/sql/drop_fns.sql`, 'utf8')
```
aguillenv, line 11: not a fan of this but `this.dirname` is available after `init()` and we need this first

dfellis: There's a better solution here.

The same module that's used to figure out the `dirname` property could be used here. The `sql` object could be defined statically on the module just with a relative path string for the `sql` files, and then the `init` logic could replace them with the contents of the files.

aguillenv: oh, that definitely will work

src/modules/interfaces.ts
```diff
@@ -539,6 +539,16 @@ export class ModuleBase {
...
       }
+    } else {
+      // If no path specified, try to get the default
+      try {
+        const sql = fs.readFileSync(`${this.dirname}/sql/after_install.sql`, 'utf8');
+        this.migrations.afterInstall = async (q: QueryRunner) => {
+          await q.query(sql);
+        };
+      } catch (e) {
+        logger.warn(`Unable to read default file ${this.dirname}/sql/after_install.sql`);
```
dfellis, line 550: I don't know if it should log anything in this case, since there may not even be a file in question for the module?

src/modules/0.0.20/iasql_functions/index.ts
```diff
@@ -5,6 +5,10 @@ import { IasqlOperationType } from './entity';
 class IasqlFunctions extends ModuleBase {
   constructor() {
     super();
+    this.sql = {
+      afterInstallSqlPath: 'sql/create_fns.sql',
+      beforeUninstallSqlPath: 'sql/drop_fns.sql',
+    };
```
dfellis, line 11: Oh, I thought these were gonna be inferred?

dfellis (approved):
LGTM. Only one minor comment.

src/modules/0.0.20/iasql_functions/index.ts
aguillenv: well I have left both options, if it is here, takes priority, if not is inferred in `sql/after_install.sql` or `sql/before_uninstall.sql`

dfellis: Ah, right, and you're exercising both paths, here and in `aws_memory_db`

PR #1236, opened 2022-09-07
RFC for better policy UX

ajejoseph22:
In the second draft, the newly proposed solution is a function that takes in policy inputs, validates and returns JSON. Details in PR.


rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+                        ) WHERE id = '<bucket_id>';
+```
+
+These functions can be used to create policies for any AWS service (SQS, VPC, SNS etc.). We can also add more functions to create policies for
+services in other clouds like GCP, Azure (out of scope).
+
+Now, once the JSON blob is generated from the `Policy` and `Statement` tables, the data in these tables are no longer needed.
+We can choose to keep them but the tables will keep growing and will span billions of records in no time. I recommend we
+have a retention policy for the tables (perhaps we can retain them for 12-24 hours and then delete).
+We can also choose to delete them immediately after the JSON blob is generated, though a periodic bulk-delete is cheaper.
```
dfellis, line 138: This part doesn't make sense to me? Wouldn't the `bucket` table in your example still have a relation to the `policy` table by the `policy_id`? How can we delete it without removing the policy from the bucket?

Also, how would it grow to billions of records in no time?

rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+![Proposed model for s3](./assets/004/s3_example.png)
+
+And for Updating, we can equally have something like this:
+
+```sql
+UPDATE bucket set policy_id = iasql_create_policy( -- returns policy ID after creation
+                              'some_policy_name',
+                              '2012-10-17', -- nullable, defaults to 2012-10-17
+                              iasql_create_policy_statement -- returns policy statement ID after creation
+                                '<policy_id>',
```
dfellis, line 120: How would this inner policy statement function get the policy ID if it is created before the policy record?

rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+Instead of the customer having to look up and put together a JSON blob like this and then verifying and then inserting, we can also give them
+the option to do this using functions that would generate this blob.
+
+```sql
+-- User creates policy
+select * from iasql_create_policy(
+    'some_policy_name',
+    '2012-10-17' -- nullable, defaults to 2012-10-17
```
dfellis, line 63: If the policy document version is changed to another version, how will the tables handle the changes to the JSON properties allowed?

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: It's an error there. <policy_id> would not be part of that insert query.

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: I should have added that the policy_id field will be set to `NULL`(if we decode to let go of the records) .

Well, each policy statement represents a record in the `Statement` table. Policies can contain many statements. In the event that the Statements table is relatively large, it becomes easier to create new statements instead of querying and to see what can be reused. I imagine that the combined `Statement` and `Policy` records for all users across all accounts could grow quite fast and considering the records have no value anymore, it seems like a waste of resources to me.

rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+                            '<policy_id>',
+                            'allow',
+                            'aws:arn:123456789012:bucket-resource',
+                            's3:GetObjects', -- comma separated list of actions
+                            '123456789012' -- comma separated list of principals
+                            ), -- Statement 1
+                          iasql_create_policy_statement(...), -- Statement 2
+                          ...
+                          iasql_create_policy_statement(...) -- Statement N
+                    );
```
dfellis, line 107: If I replace this insert with a JSON string, even if human-legibly formatted, it seems to have no impact on complexity for the user?

```suggestion
insert into bucket (
                name,
                policy_document, -- nullable if policy_id is provided
                policy_id -- nullable if policy_document is provided
            ) values (
                      'some_bucket',
                      $${
                        "Version": "2012-10-17",
                        "Statement": [{
                          "Effect": "Allow",
                          "Action": [ "s3:GetObjects" ],
                          "Resource": "aws:arn:123456789012:bucket-resource",
                          "Prinicpal": {
                            "wheredidthiskeygo": "123456789012"
                          }
                        }]
                      }$$
            );
```

So I don't think it is providing what could be considered a better experience for the users?

rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+    '2012-10-17' -- nullable, defaults to 2012-10-17
+    );
+
+-- User creates statement and attaches to the policy
+select * from iasql_create_policy_statement(
+        '<policy_id>',
+        'allow',
+        'aws:arn:123456789012:bucket-resource',
+        's3:GetObjects', -- comma separated list of actions
+        '123456789012' -- comma separated list of principals
```
dfellis, line 72: The example above shows the Principals as a key-value structure. This doesn't seem possible?

rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+    'some_policy_name',
+    '2012-10-17' -- nullable, defaults to 2012-10-17
+    );
+
+-- User creates statement and attaches to the policy
+select * from iasql_create_policy_statement(
+        '<policy_id>',
+        'allow',
+        'aws:arn:123456789012:bucket-resource',
+        's3:GetObjects', -- comma separated list of actions
```
dfellis, line 71: The actions still being strings means that if I typo `s3:GetObjects` as `s3;GetObjects` it will fail on me and IaSQL will provide no warning ahead of time that this is the case.

rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+select * from iasql_create_policy(
+    'some_policy_name',
+    '2012-10-17' -- nullable, defaults to 2012-10-17
+    );
+
+-- User creates statement and attaches to the policy
+select * from iasql_create_policy_statement(
+        '<policy_id>',
+        'allow',
+        'aws:arn:123456789012:bucket-resource',
```
dfellis, line 70: Similar concern on the resources, however I am unsure if it is even possible to get a listing of all ARNs from the AWS SDK to use for validation purposes.

rfcs/004 - Better Policy UX RFC.md
dfellis: But these tables will be isolated from each other. Statements and Policies created by other users are not visible in your own database, so the growth potential would be limited only by what the users themselves decide to insert?

rfcs/004 - Better Policy UX RFC.md
dfellis: Got it.

rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+These functions can be used to create policies for any AWS service (SQS, VPC, SNS etc.). We can also add more functions to create policies for
+services in other clouds like GCP, Azure (out of scope).
+
+Now, once the JSON blob is generated from the `Policy` and `Statement` tables, the data in these tables are no longer needed.
+We can choose to keep them but the tables will keep growing and will span billions of records in no time. I recommend we
+have a retention policy for the tables (perhaps we can retain them for 12-24 hours and then delete).
+We can also choose to delete them immediately after the JSON blob is generated, though a periodic bulk-delete is cheaper.
+
+
+### Alternatives Considered
```
aguillenv, line 141: in the alternatives, I see at least one more option which is just having tables instead of the mixed proposal. Maybe would be good to expand on that one a bit.

rfcs/004 - Better Policy UX RFC.md
dfellis: Perhaps there is a way to check if an ARN string is a valid one, though?

rfcs/004 - Better Policy UX RFC.md
aguillenv, line 138: hmm but if we delete the policy, how is the relation between tabled kept? Also, if removed, how are we going to be able to know if there was a change in the policy for a resource while doing `sync` or `apply`?

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: There are only 2 versions right now (2008 and 2012) and AWS strongly warns against using the 2008 one for new/existing policies.
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_version.html

AFAIK, both versions support the same JSON properties, but the old version is missing some features like policy variables.
Case: If the user sets a version as `2008-10-17` and sets the `Principal` as `${aws:username}`, the policy is generated as-is still. Though, on `apply`, they will get the regular AWS error saying "Principal does not exist", because version `2008-10-17` will read the variable as a string literal.

What specific case do you envision?

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: Yes, that's correct.

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: My thinking here was to have the postgres function generate something like this
```
{
  "AWS": [
          "22232323232",
          "56677889765"
        ]
}
```
when multiple strings are provided. The way it is on https://awspolicygen.s3.amazonaws.com/policygen.html

dfellis: So digging through [the documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies.html) it says

> Principal (Required in only some circumstances) – If you create a resource-based policy, you must indicate the account, user, role, or federated user to which you would like to allow or deny access. If you are creating an IAM permissions policy to attach to a user or role, you cannot include this element. The principal is implied as that user or role.

It's not a property that is always necessary (in fact, most of the examples do not include this).

There are also other properties not being modeled here, particularly `sid` and `condition`, where `condition` [can have some very strange JSON embedded in it](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_aws_deny-requested-region.html) that this proposal can't model.

rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+
+* Creating helper postgres functions to help the user generate policies from scratch (or from existing policies).
```
yrobla, line 143: i believe that this could be a good approach for the user . Using https://awspolicygen.s3.amazonaws.com/policygen.html as a base, have some stored procedure that accepts some parameters, and returns the generated json ? that can be called easily in inserts: INSERT INTO x (policy_value) SELECT generate_policy('type', vals...)

rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+Now, once the JSON blob is generated from the `Policy` and `Statement` tables, the data in these tables are no longer needed.
+We can choose to keep them but the tables will keep growing and will span billions of records in no time. I recommend we
+have a retention policy for the tables (perhaps we can retain them for 12-24 hours and then delete).
+We can also choose to delete them immediately after the JSON blob is generated, though a periodic bulk-delete is cheaper.
+
+
+### Alternatives Considered
+
+* Creating helper postgres functions to help the user generate policies from scratch (or from existing policies). 
+These functions would return JSON e.g. `create_bucket_write_object_policy(create_bucket_read_object_policy(bucket_name))`.
```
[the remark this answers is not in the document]
ajejoseph22: Yes, will do this.

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: Thanks for the suggestion. I considered this one. I will add details as to why I don't think it may be better that the proposed solution.

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: Good point. Will document.

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: The intention is to only use the tables for input, just to generate the JSON blob. Once the JSON blob is generated and stored on the entity, we can discard the table records and do all the validations and checks for changes in the same way we currently do now (using the blobs).

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: Do we really need to do this? From my observations, TF, CDK, CFN and others don't yet have this kind of validation, except at `apply` time.

If we absolutely do need to, then perhaps we use a custom Fn/graphile to make a call using the AWS SDK to validate the entire policy JSON (not just the ARN).
ref: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/AccessAnalyzer.html#validatePolicy-property

We can also just do a quick regex check on properties like `Resource` ARNs, for simplicity.
A regex check will work for `Actions` too and we can also check them against an enum/set of enums.

ajejoseph22: Pushing more details on this in a minute

dfellis: There are a couple of different concerns I want to address here:

1. On the comparison vs Terraform, CDK, etc. You are right that other deployment tools are not solving this issue, but that's part of why IaSQL can actually make sense. Better, earlier enforcement of what you're constructing in the database, enforced by foreign keys, column types/constraints, etc, should make `apply` less of a *gamble* than it is today. My concern/question here is: is this an unavoidable issue with AWS's design, or can we do better than the competition here?
2. This is not to say that I would *require* a solution here. "Perfect is the enemy of good" and all. If the time/resources to fix this are deemed too high by the RFC process itself, then we can note it as a "nice to have" in the future. If the RFC process says its impossible, then we can not worry about it at all. But if the RFC process says, "yeah, we can do that, and it'd add an extra hour or day of development work," then it seems reasonable to me to modify the RFC to include that as a requirement for the eventual implementation.

rfcs/004 - Better Policy UX RFC.md
```diff
@@ -0,0 +1,156 @@
...
+                            '123456789012' -- comma separated list of principals
+                            ), -- Statement 1
+                          iasql_create_policy_statement(...), -- Statement 2
+                          ...
+                          iasql_create_policy_statement(...) -- Statement N
+                    );
+```
+
+### ER diagram for s3 example
+![Proposed model for s3](./assets/004/s3_example.png)
```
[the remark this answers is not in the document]
ajejoseph22: It would. I have modified this in second draft

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: I have addressed this in the second draft by suggesting regex checks and possibly a call to the AWS SDK to validate the policy

rfcs/004 - Better Policy UX RFC.md
ajejoseph22: I have addressed these in the second draft.
Condition always takes the form of 
```
{
  "<Condition>": {
     "<Key>": "<Value>"
  }
}
```
I represented this as an array of triples

PR #1240, opened 2022-09-08
rpc subtype

aguillenv:
Part of #1192


src/modules/interfaces.ts
```diff
@@ -343,6 +349,16 @@ export class MapperBase<E> {
   }
 }
 
+export class RpcBase {
+  module: ModuleInterface;
+  name: string;
+  call: (ctx: Context, ...args: string[]) => Promise<string>;
```
dfellis, line 355: This type doesn't look very good to me, tbh. Instead of returning the JSON-stringified version, wouldn't it be better to enforce that it's an array of objects that all have the same keys? (Representing that last bit in Typescript is hard, but not impossible. It could just be `Promise<any[]>` for now, and still be a bit of an improvement, though.)

Then the logic in the scheduler can have the `JSON.stringify` call, instead (so it is still done).

However, it may also make sense to change the return type to something a bit different. Right now all of the columns are parsed as `text` on the other side, but that means all of the users of these custom functions have to cast columns manually.

What if we require there to be a `types` property on this object that is an object of `columnName: postgresTypeString` that the Postgres function generation code can use to automatically generate both the casting of the columns, but an explicit return column list to make usage and documentation of the RPC on the Postgres side clearer?

aguillenv: Hmm, so that the return table for this custom function would be dynamic, returning something like:

```
module_name | method_name | return_column_1 | return_column_2 | return_column_n
text        | text        | type_1          | type_2          | type_n
```

But the `iasql_rpc` table would still be saving an `output::text`?

dfellis: It may be better to explain it in a voice chat, but, no, the actual return type across the RPC boundary between the engine and postgres would remain the *same*, but the return type from the RPC function definition in the engine to the code in the scheduler is where the stronger typing would exist for the actual return columns.

*Relatedly*, we could attach extra metadata to this class in the form of `columnName: postgresTypeToCastTo` that the logic in here to *generate* the actual postgres function on the other side can use to both describe the output records better *and* automatically cast the data for the user.

src/modules/0.0.21/iasql_functions/iasql.ts
```diff
@@ -0,0 +1,301 @@
+// TODO: It seems like a lot of this logic could be migrated into the iasql_platform module and make
+// sense there. Need to think a bit more on that, but module manipulation that way could allow for
+// meta operations within the module code itself, if desirable.
```
dfellis, line 3: This TODO is being accomplished in this PR (when I wrote it, we didn't have a separate `iasql_functions` module).

src/modules/0.0.21/iasql_functions/rpcs/iasql_apply.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export class IasqlApply extends RpcBase {
+  module: IasqlFunctions;
+  name = 'iasql_apply';
+  output = {
+    action: 'varchar',
+    table_name: 'varchar',
+    id: 'varchar',
+    description: 'varchar',
+  } as const;
```
dfellis, line 15: What does `as const` do?

src/modules/0.0.21/iasql_functions/rpcs/iasql_apply.ts
```diff
@@ -0,0 +1,36 @@
+import { snakeCase } from 'typeorm/util/StringUtils';
+
+import { IasqlFunctions } from '..';
+import { Context, RpcBase, RpcResponseObject } from '../../../interfaces';
+import * as iasql from '../iasql';
+
+export class IasqlApply extends RpcBase {
+  module: IasqlFunctions;
+  name = 'iasql_apply';
```
dfellis, line 9: Should we try to auto-derive this `name` variable from the class name using the `snakeCase` function?

src/modules/0.0.21/iasql_functions/rpcs/iasql_apply.ts
```diff
@@ -0,0 +1,36 @@
...
+    dbId: string,
+    _dbUser: string,
+    _ctx: Context,
+  ): Promise<RpcResponseObject<typeof this.output>[]> => {
+    const applyRes = (await iasql.apply(dbId, false)).rows;
+    const formattedRes = applyRes.map(rec =>
+      Object.keys(rec).reduce((acc, key) => {
+        acc[snakeCase(key)] = rec[key];
+        return acc;
+      }, {} as any),
```
dfellis, line 26: Not a big nit, but in case you weren't aware, you can also write this as:

```suggestion
      Object.fromEntries(Object.entries(rec).map((k, v) => [snakeCase(k), v]))
```

src/modules/0.0.21/iasql_functions/rpcs/iasql_preview_apply.ts
```diff
@@ -0,0 +1,36 @@
+import { snakeCase } from 'typeorm/util/StringUtils';
```
dfellis, line 1: Same general comments here, and if you find that you're repeating a lot of cruft in here, it may make sense to either bake it into `RpcBase` or provide some `super.method(...)` functions (like that `formattedRes` could become `return super.snakeCaseRecords(applyRes);`)

src/modules/interfaces.ts
```diff
@@ -530,45 +635,17 @@ export class ModuleBase {
...
-      try {
-        const sql = fs.readFileSync(`${this.dirname}/sql/after_install.sql`, 'utf8');
-        this.migrations.afterInstall = async (q: QueryRunner) => {
-          await q.query(sql);
-        };
-      } catch (_) {
-        /** Don't do anything if the default file is not there */
-      }
+    const afterInstallMigration = customAfterInstallSql + rpcAfterInstallSql;
+    const beforeUninstallMigration = rpcBeforeUninstallSql + customBeforeUninstallSql;
```
dfellis, line 639: Should there be a `"\n"` added in between these two to make sure concatenation doesn't produce unparseable SQL?

src/services/iasql.ts
```diff
@@ -383,6 +383,8 @@ function colToRow(cols: { [key: string]: any[] }): { [key: string]: any }[] {
   return out;
 }
 
+// ! DEPRECATED
+// `apply` HAVE BEEN MOVED TO `iasql_functions` MODULE. DO NOT UPDATE HERE!
```
dfellis, line 387: # :partying_face:

test/modules/aws-rds-integration.ts
```diff
@@ -45,7 +45,7 @@ describe('RDS Integration Testing', () => {
         VALUES ('${prefix}test', 20, 'db.t3.micro', 'test', 'testpass', '${availabilityZone}', 'postgres:13.4', 0);
       INSERT INTO rds_security_groups (rds_id, security_group_id) SELECT
         (SELECT id FROM rds WHERE db_instance_identifier='${prefix}test'),
-        (SELECT id FROM security_group WHERE group_name='default');
+        (SELECT id FROM security_group WHERE group_name='default' LIMIT 1);
```
dfellis, line 48: This isn't really the correct logic here, but it works. What is going on is a few test accounts accidentally have two or more VPCs defined that weren't cleaned up, and this code isn't also joining the security group on the default VPC.

Choosing a random VPC by selecting a random default security group is not something anyone would actually want to do outside of our own test suite, but copy-pasting happens, so I'd prefer to set a better example. :)

src/modules/0.0.21/iasql_functions/rpcs/iasql_apply.ts
aguillenv: oh it was the first thing that came to my mind, yours seems cleaner, I can test it

src/modules/0.0.21/iasql_functions/rpcs/iasql_preview_apply.ts
aguillenv: 💯

src/modules/interfaces.ts
aguillenv: I make sure of that by adding break lines in the strings with the template strings syntax ``

test/modules/aws-rds-integration.ts
aguillenv: oh yes, I will remove this, it also does not belong to this PR. It was only to avoid a test failure.

src/modules/interfaces.ts
dfellis: got it

src/modules/interfaces.ts
```diff
@@ -479,13 +516,74 @@ export class ModuleBase {
...
+  private getRpcSql(): [string, string] {
+    let afterInstallSql = '';
+    let beforeUninstallSql = '';
+    for (const [key, rpc] of Object.entries(this.rpc ?? {})) {
+      const rpcOutputEntries = Object.entries(rpc.output ?? {});
+      const rpcOutputTable = rpcOutputEntries
+        .map(([columnName, columnType]) => `${columnName} ${columnType}`)
+        .join(', ');
+      afterInstallSql += `
+        create or replace function ${rpc.name}(variadic _args text[] default array[]::text[]) returns table (
```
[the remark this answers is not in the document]
aguillenv: oh, this would be nice, could be a follow-up improvement, though.

> Does Postgres complain if you did something like `SELECT * FROM my_rpc_fn(3);` or will it auto-cast to text for you?
Yes, it complains it does not exist any function with that name waiting for an `integer` parameter.

src/modules/0.0.21/iasql_functions/rpcs/iasql_modules_list.ts
```diff
@@ -0,0 +1,33 @@
...
+  call = async (
+    dbId: string,
+    _dbUser: string,
+    _ctx: Context,
+  ): Promise<RpcResponseObject<typeof this.outputTable>[]> => {
+    const res = await iasql.modules(true, false, dbId);
+    return res.map(rec => {
+      const updatedRec = {
+        moduleName: rec.moduleName,
+        moduleVersion: `${rec.moduleVersion}`,
```
dfellis, line 21: How is this any different from:

```suggestion
        moduleVersion: rec.moduleVersion,
```

src/services/db-manager.ts
```diff
@@ -22,9 +20,13 @@ export async function migrate(conn: Connection) {
     await iasqlPlatform.migrations.afterInstall(qr);
   }
   await qr.query(`INSERT INTO iasql_module VALUES ('iasql_platform@${version}')`);
-  await iasqlFunctions.migrations.install(qr);
-  if (iasqlFunctions.migrations.afterInstall) {
-    await iasqlFunctions.migrations.afterInstall(qr);
+  await ModuleSet?.IasqlFunctions?.migrations?.install(qr);
```
dfellis, line 23: Looking forward to dropping a lot of weirdness in this file once `v0.0.16` is gone.

dfellis (approved):
Only a couple of minor comments.

src/services/db-manager.ts
aguillenv: Agree!

src/services/scheduler.ts
```diff
@@ -194,7 +201,7 @@ export async function start(dbId: string, dbUser: string) {
           try {
             const recordCount = await iasql.getDbRecCount(conn);
             const rpcCount = await iasql.getRpcCount(conn);
-            await MetadataRepo.updateDbCounts(dbId, recordCount, rpcCount);
+            await MetadataRepo.updateDbCounts(dbId, recordCount, undefined, rpcCount);
```
depombo, line 204: TODO to remove this as well

depombo (approved):

src/services/scheduler.ts
aguillenv: I did not add it exactly here but in the function that needs to be updated.

PR #1241, opened 2022-09-08
Several fixes to issues encountered during 0.0.19 deployment attempt

dfellis:
The issues fixed:

- `docker-entrypoint.sh` now shows the stdout of the scheduler process for better debugging needs
- `iasql_upgrade` no longer barfs on the `sync` call which left the DB half-upgraded
- `default_aws_region` no longer returns NULL if there are no `aws_regions` records with `is_default = true`

Confirmed with manual testing that was then turned into new unit/integration tests.

### DIFF
--- a/src/config/bootstrap.ts
+++ b/src/config/bootstrap.ts
@@ -6,7 +6,7 @@ const config: ConfigInterface = {
   },
   modules: {
     latestVersion: '0.0.19',
-    oldestVersion: '0.0.16'
+    oldestVersion: '0.0.16',
   },
   db: {
     host: 'db-bootstrap.iasql.com',
--- a/test/upgrade/run-upgrade.sh
+++ b/test/upgrade/run-upgrade.sh
@@ -14,6 +14,9 @@ LATESTVERSION=`./node_modules/.bin/ts-node src/scripts/latestVersion.ts`
 OLDESTVERSION=`./node_modules/.bin/ts-node src/scripts/oldestVersion.ts`
 CURRENTGITSHA=`git rev-parse HEAD`
 
+# Github Actions apparently doesn't pull down the tags by default?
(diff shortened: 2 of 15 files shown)


test/upgrade/run-upgrade.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
```
depombo, line 1: Amazing!

test/upgrade/run-upgrade.sh
```diff
@@ -0,0 +1,74 @@
...
+  UPGRADECHECKCOUNT=$((${UPGRADECHECKCOUNT}-1))
+done
+
+# The actual check!
+if [ "${ISUPGRADED}" == "false" ]; then
+  echo "Did not successfully upgrade!";
+  exit 2;
+else
+  echo "Successfully upgraded!";
+fi
```
depombo, line 74: I think I would have written this script in Node.js with a ton of `execSync` calls to avoid this bit, but I am just not that familiar with bash

src/config/bootstrap.ts
```diff
@@ -6,7 +6,7 @@ const config: ConfigInterface = {
   },
   modules: {
     latestVersion: '0.0.19',
-    oldestVersion: '0.0.16'
+    oldestVersion: '0.0.16',
```
depombo, line 9: we should change the `.version` to add the trailing comma or call `yarn format` before committing

depombo (approved):

src/config/bootstrap.ts
dfellis: 👍

depombo (approved):

yrobla (approved):

aguillenv (approved):

Issue #1230, opened 2022-09-07
Add a ECR image list table to `aws_ecr` module


depombo:
Each listed image in the table (could be called `image_list`?) should have a foreign key to `repository` or `public_repository` in `aws_ecr`. This will let us remove the usage of the AWS CLI in the ECS tutorial which is currently necessary to delete images before deleting the repository. We should try to follow the same convention as the ACM modules. Perhaps simple be able to list and delete, but not upload for the first pass?

depombo:
@yrobla this can be a good candidate for a task to take once you are done with ACM request and cloud front + s3. I am not sure exactly how it should work, but you have thought about this for a while with ACM so you probably have a better idea. And whatever design we go with should also be what we follow in #1231

PR #1242, opened 2022-09-09
 feat: add image support on ecr repos

yrobla:
Closes: #1230


src/modules/0.0.20/aws_ecr/index.ts
```diff
@@ -9,7 +18,182 @@ import { policiesAreSame } from '../../../services/aws-diff';
...
+      const opts = repositoryName
+        ? {
+            where: {
+              repositoryName,
+            },
+          }
+        : {};
+      return await ctx.orm.find(RepositoryImage, opts);
+    },
+  });
```
aguillenv, line 113: why do we need this special db read method?

src/modules/0.0.20/aws_ecr/index.ts
```diff
@@ -9,7 +18,182 @@ import { policiesAreSame } from '../../../services/aws-diff';
...
+    },
+    updateOrReplace: () => 'update',
+    update: async (es: RepositoryImage[], ctx: Context) => {
+      return [];
+    },
+    delete: async (es: RepositoryImage[], ctx: Context) => {
+      const client = (await ctx.getAwsClient()) as AWS;
+      for (const e of es) {
+        if (e.privateRepository) await this.deleteRepositoryImage(client.ecrClient, e.privateRepository);
+        else if (e.publicRepository) await this.deleteRepositoryImage(client.ecrClient, e.publicRepository);
```
aguillenv, line 186: I think we are using the wrong client here for delete the public images

dfellis (approved):

src/modules/0.0.20/aws_ecr/index.ts
```diff
@@ -105,6 +346,8 @@ class PublicRepositoryMapper extends MapperBase<PublicRepository> {
     delete: async (es: PublicRepository[], ctx: Context) => {
       const client = (await ctx.getAwsClient()) as AWS;
       for (const e of es) {
+        // need to delete images associated with this repository
+        await this.module.repositoryImages.deletePublicRepositoryImages(client.ecrPubClient, e);
```
depombo, line 350: hmm I am not 100% sure about this one. we would be overriding the database functionality. The user should have deleted entries in the `image` table or done a delete query on `public_repository` with `CASCADE` and then the lazyLoader should converge on deleting the images and then the repository?

src/modules/0.0.20/aws_ecr/index.ts
```diff
@@ -256,6 +499,9 @@ class RepositoryMapper extends MapperBase<Repository> {
     delete: async (es: Repository[], ctx: Context) => {
       const client = (await ctx.getAwsClient()) as AWS;
       for (const e of es) {
+        // need to delete images associated with this repository
+        await this.module.repositoryImages.deleteRepositoryImages(client.ecrClient, e);
+
```
depombo, line 504: same comment as above

test/modules/aws-ecr-integration.ts
```diff
@@ -1,251 +1,475 @@
...
+    query(
+      `
     INSERT INTO aws_credentials (access_key_id, secret_access_key)
     VALUES ('${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}')
-  `, undefined, false));
+  `,
+      undefined,
+      false,
+    ),
+  );
```
depombo, line 60: the format seems to be changing drastically which is totally fine, but there are so many re-formats that it makes the *actual* changes harder to track. can we avoid this in the future by doing it in separate PRs / commits?

depombo: (this is a nit)

yrobla:
@depombo i tried to use the db approach you suggested. But i found a problem with on_cascade property. If i setup in manytoone side as i did there, it is not deleting the related images.
If i add the onDelete:cascade on the onetomany side, it throws an error: update or delete on table "repository" violates foreign key constraint "FK_3c7183f97647067d207d9276df6" on table "repository_image"

What we need here is to actually delete the associated images but before deleting the parent row. What should be the right settings to setup in the models to achieve it?

depombo:
> @depombo i tried to use the db approach you suggested. But i found a problem with on_cascade property. If i setup in manytoone side as i did there, it is not deleting the related images. If i add the onDelete:cascade on the onetomany side, it throws an error: update or delete on table "repository" violates foreign key constraint "FK_3c7183f97647067d207d9276df6" on table "repository_image"
> 
> What we need here is to actually delete the associated images but before deleting the parent row. What should be the right settings to setup in the models to achieve it?

Mmm I would take a look at what we do in ECS, maybe we need this https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.20/aws_ecs_fargate/entity/task_definition.ts#L113. I am not sure. We could update the remaining fk constraints that get violated to null as described here
https://stackoverflow.com/questions/29305653/postgresql-deleting-rows-referenced-from-another-table, but I am actually not sure how to express that in typeorm. Tbh I would leave the models alone, delete the lines I initially commented on and explicitly delete images under a given repository in the tests.

yrobla:
I tried your cascade approach but it actually tries to first delete the repo and then the images. And it cannot be done because aws complains that cannot delete a repo with images.
So you prefer that we do not remove images automatically when we delete the repo but instead user has to delete that manually ?

test/modules/aws-ecr-integration.ts
```diff
@@ -1,251 +1,494 @@
...
-  
-    it('deletes the public repository', query(`
+
+    it(
+      'deletes the repository images',
+      query(`
+    DELETE FROM repository_image WHERE public_repository= '${pubRepositoryName}';
+  `),
+    );
+    it('applies deletes the public repository image', apply());
```
depombo, line 427: this should work without this `apply` or this here on purpose? the lazyLoader should figure out the correct order to delete the resources in even if there is a transient error

depombo (approved):

PR #1271, opened 2022-09-14
AWS Codebuild support

depombo:


src/modules/0.0.21/aws_codebuild/entity/build.ts
```diff
@@ -0,0 +1,75 @@
...
+  @Column({
+    type: 'timestamp without time zone',
+    nullable: true,
+  })
+  startTime?: Date;
+}
+
+// TODO allow overrides
+@Entity()
+export class CodebuildBuildImport {
```
aguillenv, line 64: why do we need this entity?

src/modules/0.0.21/aws_codebuild/mappers/build.ts
```diff
@@ -0,0 +1,220 @@
...
+      const opts = id
+        ? {
+            where: {
+              id,
+            },
+          }
+        : {};
+      return await ctx.orm.find(CodebuildBuildImport, opts);
+    },
+  });
```
aguillenv, line 188: isn't this the default implementation for the `db.read`?

src/modules/0.0.21/aws_codebuild/mappers/project.ts
```diff
@@ -0,0 +1,173 @@
...
+    },
+    updateOrReplace: (_a: CodebuildProject, _b: CodebuildProject) => 'replace',
+    update: async (pjs: CodebuildProject[], ctx: Context) => {
+      const out = [];
+      for (const pj of pjs) {
+        const cloudRecord = ctx?.memo?.cloud?.CodebuildProject?.[pj.projectName ?? ''];
+        if (pj.arn !== cloudRecord.arn) {
+          pj.arn = cloudRecord.arn;
+          if (this.module.project.equals(pj, cloudRecord)) {
+            await this.module.project.db.update(cloudRecord, ctx);
```
aguillenv, line 143: nit: if we are restoring the record here technically it is not a `replace` but an `update` and the `updateOrReplace` should return the correct string?

src/modules/0.0.21/aws_codebuild/entity/build.ts
depombo: to start/create a build for a given project

src/modules/0.0.21/aws_codebuild/mappers/build.ts
depombo: hmm I think so, I copied it over from ACM modules since that was the only reference point I had for it. I can try removing it here, but we should probably remove it there?

src/modules/0.0.21/aws_codebuild/mappers/project.ts
depombo: I also copied this over from the ACM module

depombo: I thought the `updateOrReplace` was only called explicitly in the mapper and not in the interface?

src/modules/0.0.21/aws_codebuild/entity/build.ts
aguillenv: if it is only to trigger something this could be a good use case for a custom `rpc` function. We can implement it later.

src/modules/0.0.21/aws_codebuild/mappers/project.ts
aguillenv: hmm not sure what do you mean? What I wanted to say is that the `updateOrReplace` method for this mapper is returning always `replace` and in the `update` method we are doing a restore, so that could lead to confusion in the code and to the user when we show the changes made or to be applied.

src/modules/0.0.21/aws_codebuild/mappers/build.ts
aguillenv: I've just seen the default implementation, it is 
```
where: {
                  [cloudColumn]: id,
                },
```
so it would depend if in his case we are referring to the db `id` or `id` is the `cloudColumn`?

src/modules/0.0.21/aws_codebuild/entity/build.ts
depombo: oh, yeah it is just a trigger. that is true. we can replace this table and the acm one with it!

src/modules/0.0.21/aws_codebuild/mappers/build.ts
depombo: I was able to remove it!

docs/docs/fargate-tutorials/prisma.md
```diff
@@ -134,126 +140,214 @@ node index.js
...
+  });
+  await prisma.codebuild_build_import.create({
+    data: {
+      project_name: appName,
+    }
+  });
 
-```bash
-docker tag <project-name>-repository:latest ${QUICKSTART_ECR_URI}:latest
+  console.dir(await prisma.$queryRaw`SELECT * from iasql_apply();`)
```
aguillenv, line 249: do we want this around?

docs/docs/fargate-tutorials/prisma.md
```diff
@@ -134,126 +140,214 @@ node index.js
...
-    create: data,
-    update: data,
+    where: { app_name: appName},
+    create: ecsData,
+    update: ecsData,
   });
 
-  const apply = await prisma.$queryRaw`SELECT * from iasql_apply();`
-  console.dir(apply)
+  console.dir(await prisma.$queryRaw`SELECT * from iasql_apply();`)
```
aguillenv, line 186: do we want this around?

examples/ecs-fargate/prisma/infra/index.js
```diff
@@ -1,60 +1,109 @@
...
-    create: data,
-    update: data,
+    where: { app_name: appName},
+    create: ecsData,
+    update: ecsData,
   });
 
-  const apply = await prisma.$queryRaw`SELECT * from iasql_apply();`
-  console.dir(apply)
+  console.dir(await prisma.$queryRaw`SELECT * from iasql_apply();`)
```
aguillenv, line 43: same

examples/ecs-fargate/prisma/infra/index.js
```diff
@@ -1,60 +1,109 @@
...
+      project_name: appName,
+    }
+  });
 
-  console.log('Pushing image...')
-  console.log(execSync(
-    `docker push ${repoUri}:latest`,
-    { stdio: 'pipe', encoding: 'utf8', },
-  ));
+  console.dir(await prisma.$queryRaw`SELECT * from iasql_apply();`)
```
aguillenv, line 106: same

docs/docs/fargate-tutorials/prisma.md
yrobla, line 249: is this debug message intended to be here?

depombo: yes, this is to display the output of the `apply`

src/services/aws_macros.ts
```diff
@@ -82,6 +85,18 @@ export class AWS {
           ),
         ),
     });
+    // Get around some strange internal caching inside of codebuild of the IAM roles
```
yrobla, line 88: does it make sense to create a shared retry strategy for that, or needs to be a custom one?

aguillenv (approved):

yrobla (commented):
just some questions

src/services/aws_macros.ts
depombo: the `retryDecider` is specific to this strategy unfortunately

docs/docs/fargate-tutorials/prisma.md
depombo: yes! without it the user / we can't see the output of `apply`

yrobla (approved):

docs/docs/fargate-tutorials/prisma.md
```diff
@@ -134,126 +140,214 @@ node index.js
...
+      build:
+        commands:
+          - echo Building the Docker image...
+          - docker build -t ${appName}-repository examples/ecs-fargate/prisma/app
+          - docker tag ${appName}-repository:latest ${repoUri}:latest
+      post_build:
+        commands:
+          - echo Pushing the Docker image...
+          - docker push ${repoUri}:latest
+  `;
```
dfellis, line 229: Isn't there a Postgres function that does something like this now? Perhaps it would be better to use that here instead of exposing codedeploy's yaml syntax?

docs/docs/fargate-tutorials/prisma.md
```diff
@@ -134,126 +140,214 @@ node index.js
...
-docker tag <project-name>-repository:latest ${QUICKSTART_ECR_URI}:latest
+  console.dir(await prisma.$queryRaw`SELECT * from iasql_apply();`)
+}
 ```
 
-5. Push your image
-
-```bash
-docker push ${QUICKSTART_ECR_URI}:latest
+The two `SELECT * from iasql_apply();` queries will [`apply`](../concepts/apply-and-sync.md) the changes described in the hosted db to your cloud account which can take a few minutes waiting for AWS. It will then print arrays of javascript objects with the cloud resources that have been created, deleted, or updated.
```
dfellis, line 253: Is it possible to reduce this to just one `apply` call?

docs/docs/fargate-tutorials/prisma.md
```diff
@@ -134,126 +140,214 @@ node index.js
...
-```
 curl ${QUICKSTART_LB_DNS}:8088/health
 ```
 
 ## Delete managed cloud resources
 
 :::warning
 
-If you did not create a new account this section will delete **all** records managed by IaSQL, including the ones that previously existed in the account under any of the used modules. Run `SELECT * FROM iasql_plan_apply()` after `SELECT delete_all_records();` and before `SELECT iasql_apply();` to get a preview of what would get deleted. To undo `SELECT delete_all_records();`, simply run `SELECT iasql_sync();` which will synchronize the database with the cloud's state.
+If you did not create a new AWS account this section will delete **all** records managed by IaSQL, including the ones that previously existed in the account under any of the used modules. Run `SELECT * FROM iasql_plan_apply()` after `SELECT delete_all_records();` and before `SELECT iasql_apply();` to get a preview of what would get deleted. To undo `SELECT delete_all_records();`, simply run `SELECT iasql_sync();` which will synchronize the database with the cloud's state.
```
dfellis, line 346: Not for this PR, but with it down to just `codebuild` and `ecs_simplified`. Is it possible to provide a hand-written set of `DELETE` statements here? I am pretty sure `ecs_simplified` will do the right thing if you delete its record and drop all of the related records at the same time in the other modules, right?

src/modules/0.0.21/aws_codebuild/entity/project.ts
```diff
@@ -0,0 +1,111 @@
...
+    type: 'enum',
+    enum: SourceType,
+  })
+  sourceType: SourceType;
+
+  @Column({
+    type: 'simple-json',
+    nullable: true,
+  })
+  environmentVariables?: [{ name: string; value: string; type: EnvironmentVariableType }];
```
dfellis, line 103: We don't actually have real type-safety here. The user could insert whatever arbitrary JSON they want here but Typescript will tell us it matches the exact type you have here.

This is a general problem we need to figure out, but not excited to see more tech debt getting piled up here.

Also, since there's only one type of environment variable supported here, why not just make it `{ [key: string]: string, }` as the type, which is much more natural to work with?

src/modules/0.0.21/aws_codebuild/mappers/build.ts
```diff
@@ -0,0 +1,205 @@
...
+      builds = await this.getBuilds(client.cbClient, input);
+      if (!builds) return;
+      const out = [];
+      for (const build of builds) {
+        const outB = await this.buildListMapper(build, ctx);
+        if (outB) out.push(outB);
+      }
+      return out;
+    },
+    updateOrReplace: () => 'update',
```
dfellis, line 119: Nit. This isn't necessary since this is the exact function auto-provided for you if you don't define the `updateOrReplace` method.

src/modules/0.0.21/aws_codebuild/entity/project.ts
depombo: well, eventually I was hoping it would support secrets manager so I would rather leave the format as-is

src/modules/0.0.21/aws_codebuild/mappers/build.ts
depombo: got it, copied it from another module

docs/docs/fargate-tutorials/prisma.md
```diff
@@ -134,126 +140,214 @@ node index.js
...
-const PORT = 8088;
+// TODO replace with your desired project name
+const appName = pkg.name;
+const cbRole = `${appName}codebuild`;
+const ghToken = process.env.GH_PAT;
+const region = process.env.AWS_REGION;
+const port = 8088;
+const codebuildPolicyArn = 'arn:aws:iam::aws:policy/AWSCodeBuildAdminAccess';
+const cloudwatchLogsArn = 'arn:aws:iam::aws:policy/CloudWatchLogsFullAccess';
+// TODO provide ECR permissions once inline policies are supported in roles
```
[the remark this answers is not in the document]
depombo: this is not in the actual file and I forgot to remove it here

docs/docs/fargate-tutorials/prisma.md
depombo: yeah

docs/docs/fargate-tutorials/prisma.md
depombo: https://github.com/iasql/iasql-engine/issues/1335 and yes

src/modules/0.0.21/aws_codebuild/sql/after_install.sql
```diff
@@ -0,0 +1,21 @@
+CREATE
+OR REPLACE FUNCTION generate_put_ecr_image_build_spec (region TEXT, image_tag TEXT, repo_name TEXT, repo_uri TEXT, build_path TEXT) RETURNS TEXT LANGUAGE plpgsql AS $$
+BEGIN
+  RETURN 'version: 0.2
+
+  phases:
+    pre_build:
+      commands:
+        - echo Logging in to Amazon ECR...
+        - aws ecr get-login-password --region ' || region || ' | docker login --username AWS --password-stdin ' || repo_uri || '
```
dfellis, line 10: I see now that codebuild has no explicit support for inserting into ECR, but it's a pattern you can configure it to follow like in other CI systems, and that's why you can't (easily) turn it into a single `apply` call. I kinda *do* want a `codebuild_simplified` module, now, since github -> codebuild -> ecr and then ecr -> ecs feels like a very common, desired pipeline.

docs/docs/fargate-tutorials/prisma.md
```diff
@@ -134,126 +140,197 @@ node index.js
...
+  const buildSpecRes = await prisma.$queryRaw`SELECT generate_put_ecr_image_build_spec('${region}', 'latest', '${appName}-repository', '${repoUri}', 'examples/ecs-fargate/prisma/app')`;
+  const buildSpec = buildSpecRes[0]['generate_put_ecr_image_build_spec'];
 
-4. Tag your image
+  const pjData = {
+    project_name: appName,
+    source_type: 'GITHUB',
+    service_role_name: cbRole,
+    source_location: ghUrl,
+    build_spec: buildSpec,
```
dfellis, line 219: Does prisma not allow an `upsert` column value to come from a Postgres function? That's a weird limitation. Let me look into it.

dfellis (approved):

PR #1273, opened 2022-09-14
Erase v0.0.19 from the codebase

dfellis:


depombo (approved):

mtp1376:
What is the purpose for this PR?

dfellis:
> What is the purpose for this PR?

`v0.0.19` was never deployed because it was broken. I have spent a little over a week figuring out how multi-region support broke it in subtle ways that our test suite never caught and have been adding tests and fixing bugs involving it.

Because this was figured out *after* development of `v0.0.20` began, this is simply excising it from the set of versions for today's deployment so it will be impossible for anyone to get onto the broken release.

PR #1279, opened 2022-09-16
Skip 'bad' image records from ECR

dfellis:


src/modules/0.0.20/aws_ecr/index.ts
```diff
@@ -239,7 +246,7 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
         if (e.privateRepository) {
           await this.deleteRepositoryImage(client.ecrClient, [imageId], e.privateRepository.repositoryName);
         } else if (e.publicRepository) {
-          await this.deleteRepositoryImage(client.ecrClient, [imageId], e.publicRepository.repositoryName);
+          await this.deleteRepositoryImage(client.ecrPubClient, [imageId], e.publicRepository.repositoryName);
```
depombo, line 249: shouldn't the type system complain here?

dfellis:
This was a bad change based on erroneous assumptions. Fixed the real issue.

Issue #1284, opened 2022-09-18
Add support for appSync module


yrobla:
AppSync is used by Lambda, Amplify, and other sources. Add support for it, starting with basic GraphQL support

PR #1285, opened 2022-09-18
feat: add appsync support, starting with graphql

yrobla:
Create the initial module and add support for the first basic entity, that will be graphql

Closes: #1284


src/modules/0.0.21/aws_appsync/index.ts
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    return out;
+  }
+
+  async createGraphqlApi(client: AppSync, input: CreateGraphqlApiCommandInput) {
+    const res = await client.createGraphqlApi(input);
+    if (res && res.graphqlApi) return res.graphqlApi;
+    else return undefined;
+  }
```
aguillenv, line 54: nit: why not `crudBuilderFormat`or `crudBuilder2`?

src/modules/0.0.21/aws_appsync/index.ts
```diff
@@ -0,0 +1,180 @@
...
+        const cloudRecord = ctx?.memo?.cloud?.GraphqlApi?.[api.name ?? ''];
+        const isUpdate = Object.is(this.module.graphqlApi.cloud.updateOrReplace(cloudRecord, api), 'update');
+        if (isUpdate) {
+          // in case of key fields being modified, restore them
+          if (api.apiId !== cloudRecord.apiId || api.arn !== cloudRecord.arn) {
+            api.apiId = cloudRecord.apiId;
+            api.arn = cloudRecord.arn;
+            await this.module.graphqlApi.db.update(api, ctx);
+            out.push(api);
+          } else {
```
aguillenv, line 127: hmm, I think this might need to be refactored a little bit. If the user updates the `api_id` and an updateable field like `authenticationType` it will only restore the record but ignore the cloud update. I think what we do in other cases is keep track if an update occurred, if not, check if we need to restore.

src/modules/0.0.21/aws_appsync/entity/graphql_api.ts
```diff
@@ -0,0 +1,74 @@
...
+  @Column({
+    type: 'json',
+    nullable: true,
+  })
+  openIDConnectConfig?: {
+    authTTL: number | undefined;
+    clientId: string | undefined;
+    iaTTL: number | undefined;
+    issuer: string | undefined;
+  };
```
dfellis, line 62: Nit. These two objects use `Ttl` and `TTL`. Can we standardize on one of the two?

dfellis (commented):
Looks good to me. :) Only a couple of nits that I could see.

aguillenv (approved):

dfellis (approved):

Issue #1283, opened 2022-09-18
Add a check in CI to warn when changes are brought to older versions


yrobla:
New pull requests are only allowed for the current development version. If code modifications are brought to older versions it may be because of a mistake. We need to add a rule that detects it and fails

PR #1288, opened 2022-09-19
feat: add ci check to detect changes in not current module

yrobla:
When working in a module sometimes there is the need to do a rebase and start working on newever version. This can bring rebase problems, such as leaving files in the older module. Add a check to prevent it.

Closes: #1283

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,9 @@ jobs:
         with:
           node-version: 16.x
 
+      - name: Install dependencies
+        run: yarn
+
       - name: checks modified files
         run: yarn changes-in-version
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,8 +60,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
 
       - name: Use Node.js 16.x
         uses: actions/setup-node@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,14 @@ jobs:
       - name: Run code style check
         run: yarn style
 
+  changes-in-latest-version:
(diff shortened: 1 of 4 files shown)


scripts/check_changes_in_right_version.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
depombo, line 1: we have a lot of bash scripts in the top-level folder that we should probably move here too

scripts/check_changes_in_right_version.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# get the current development version
+CURRENT_VERSION=$(grep 'latestVersion' src/config/ci.ts | cut -d ':' -f 2 | sed "s/'//g" | sed "s/,//g")
```
depombo, line 4: in case we ever change our config format and this breaks from under us, maybe we can have a small unit in the common folder that appends a string to an old version and checks that this fails correctly?

scripts/check_changes_in_right_version.sh
depombo: (doesn't have to happen in this PR, just thinking out loud)

.github/workflows/ci.yml
```diff
@@ -56,6 +56,14 @@ jobs:
       - name: Run code style check
         run: yarn style
 
+  changes-in-latest-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: checks modified files
+        run: yarn changes-in-version
```
dfellis, line 65: You need to add another step in here to install the dependencies like on line 53 above.

scripts/check_changes_in_right_version.sh
yrobla: yep, just thought the same when i saw the scripts so i started on a new folder

dfellis (approved):

aguillenv (approved):

Issue #1257, opened 2022-09-10
Support IAM users in `aws_iam`


depombo:
The main building block of IAM, but we didn't have a use case for it so we never created it

https://docs.aws.amazon.com/IAM/latest/UserGuide/id.html

PR #1300, opened 2022-09-20
feat: initial version for adding users into iam

yrobla:
Closes: #1257


src/modules/0.0.21/aws_iam/entity/user.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// TODO complete schema
+@Entity()
+export class IamUser {
+  @Column({
+    nullable: true,
+  })
+  arn?: string;
+
+  // Guaranteed unique in AWS
```
dfellis, line 13: Ugh. Validation of this is next-to-impossible :(

dfellis (approved):

src/modules/0.0.21/aws_iam/entity/user.ts
yrobla: this is just readonly so we do not need to validate

aguillenv (approved):

PR #1301, opened 2022-09-20
fix: check that we are comparing with version number

yrobla:
The script was checking for src/modules/ but was not considering that there are files at the root of this folder, not just version numbers. Add a check to actually compare the difference if we find a version number pattern.

### DIFF
--- a/scripts/check_changes_in_right_version.sh
+++ b/scripts/check_changes_in_right_version.sh
@@ -15,7 +15,7 @@ for FILE in "${MODIFIED_FILES[@]}"; do
   if  [[ "$FILE" == src/modules/* ]]; then
     # need to check the version
     VERSION=$(echo $FILE|cut -d '/' -f 3)
-    if [[ "$VERSION" =~ ^[0-9]+(\.[0-9]+){2,3}$ ]]; then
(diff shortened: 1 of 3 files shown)


scripts/check_changes_in_right_version.sh
```diff
@@ -12,15 +12,15 @@ IFS=$'\n' MODIFIED_FILES=($(git diff --no-commit-id --name-only -r  origin/main.
...
-    echo "check version"
+  if  [[ "$FILE" == src/modules/* ]]; then
     # need to check the version
     VERSION=$(echo $FILE|cut -d '/' -f 3)
-    echo "version is $VERSION"
-    if [ "$VERSION" != "$CURRENT_VERSION" ]; then
-      # incorrect version
-      echo "Error, out of version modifications found in file $FILE"
-      exit 1
+    if [[ "$VERSION" =~ ^[0-9]+(\.[0-9]+){2,3}$ ]]; then
```
aguillenv, line 18: why is the group expected to happen `{2,3}`? should not be only {2} to be a valid semver format?

aguillenv: we could also use one of the official regex but they are awful https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

aguillenv (approved):

PR #1303, opened 2022-09-20
Switch from batch cleaning to pre-cleaning on each CI run

dfellis:
Now that every test run gets its own account to execute on, there is no problem with a cleanup job accidentally blowing away a concurrently running test job's work, and we don't need the batch cleanup job.

This PR lazily cleans the account (on a per-region basis, for now) just before the test run itself. So all tests are gonna take a few more minutes (I'm guessing 3-5) but hopefully they will be more reliable (and/or we see state that we can't clean up and need to investigate why).


.github/workflows/ci.yml
```diff
@@ -189,27 +189,93 @@ jobs:
...
+
+      - name: Pre-clean Test Account
+        id: pre-clean-test-account
+        env:
+          IASQL_ENV: ci
+          ACCOUNT_INDEX: ${{ steps['determine-test-account'].outputs['account_index'] }}
+          AWS_REGION: ${{ steps['determine-test-account'].outputs['region'] }}
+          ACCESS_KEY_IDS: ${{ secrets.ACCESS_KEY_IDS }}
+          SECRET_ACCESS_KEYS: ${{ secrets.SECRET_ACCESS_KEYS }}
+        run: |
```
depombo, line 207: it might make sense to break this up or put it in `scripts/` so we can reuse it in the other repo for prisma tutorial runs?

depombo (approved):

aguillenv (approved):

.github/workflows/ci.yml
dfellis: I based this off of what we do in the prisma tutorial runs ;)

aguillenv (approved):

PR #1305, opened 2022-09-20
Convert 'aws_dynamo' to be multi-region aware

dfellis:


src/modules/0.0.21/aws_dynamo/index.ts
```diff
@@ -102,27 +103,32 @@ class DynamoTableMapper extends MapperBase<DynamoTable> {
...
+      const enabledRegions = (await ctx.getEnabledAwsRegions()) as string[];
       if (id) {
-        const rawTable = await this.getTable(client.dynamoClient, id);
-        if (!rawTable?.Table) return;
-        return this.dynamoMapper(rawTable.Table);
+        for (const region of enabledRegions) {
+          const client = (await ctx.getAwsClient(region)) as AWS;
+          const rawTable = await this.getTable(client.dynamoClient, id);
+          if (!rawTable?.Table) continue;
+          return this.dynamoMapper(rawTable.Table, region);
```
depombo, line 112: shouldn't we be collecting these and then returning all of them outside the for-loop?

dfellis: This is on the "request by ID" path, where the ID is the dynamoDB table name, but we don't have any information on what region that particular table is located in, so we check them in order.

But the only way for this query to make sense is if there is only one table with that name across all regions.

That assumption may not hold up, but I am not super worried at the moment because I don't believe anything uses the "cloud query dynamo table by name" path at all, so it's only included for fulfilling the typescript types.

depombo: got it

depombo (approved):

src/modules/0.0.21/aws_dynamo/index.ts
```diff
@@ -146,7 +156,10 @@ class DynamoTableMapper extends MapperBase<DynamoTable> {
           } as WaiterOptions<DynamoDB>,
           { TableName: e.tableName },
         );
-        const newTable = this.dynamoMapper(res.TableDescription);
+        const newTable = this.dynamoMapper(
+          res.TableDescription,
+          (e.region as AwsRegions)?.region ?? (e.region as string),
```
aguillenv, line 161: why `e.region` is not enough here? is it not a string?

aguillenv (commented):
what will happen on `update` if someone changes the region of a resource? would not be necessary to delete the resource in the old region and recreate it in the new one in the `update` path?

src/modules/0.0.21/aws_dynamo/index.ts
dfellis: So, I was unable yesterday to get TypeORM to generate a foreign key relation to the `aws_regions` table and also have a default value without some weird trickery with the Typescript type.

The *behavior* of TypeORM, even if you specified only `AwsRegions` as the type, is to return a `string`, but if you set the type to `string` then it fails to generate the foreign key relation for some reason, so I did a weird `string | AwsRegions` type and now the code above has to probe for the `region` string to make Typescript happy.

I do want to try to simplify this again, since this "pattern" could end up in a lot of modules and it is really clunky, but I am not sure I can make it any better than it currently is without manually editing the migration files, which is also suboptimal.

dfellis:
> what will happen on `update` if someone changes the region of a resource? would not be necessary to delete the resource in the old region and recreate it in the new one in the `update` path?

I think it would try to execute an update call in the wrong region and fail hard. I will write a test for this and fix it.

src/modules/0.0.21/aws_dynamo/index.ts
aguillenv: it is not optimal but could be a trigger after insert/update, if no region, set default. let me see if I can think about something else

dfellis: No, this really does work and it has Postgres set the default value exactly like we want.

We're in the codebase all the time and have the mental model in our heads, the users will presumably interact with IaSQL, even in the best of cases, a far smaller amount of time, so the more weirdness exposed to them the harder it will be to adopt.

I am digging through the TypeORM decorator's actual implementation to see if I can find another way.

aguillenv: did a quick search and seems to be an open issue not expected to be solved https://github.com/typeorm/typeorm/issues/2427

aguillenv (approved):

PR #1309, opened 2022-09-21
Multi-Region 'aws_elasticache'

dfellis:


src/modules/0.0.21/aws_elasticache/index.ts
```diff
@@ -17,7 +17,7 @@ class CacheClusterMapper extends MapperBase<CacheCluster> {
   equals = (a: CacheCluster, b: CacheCluster) =>
```
aguillenv, line 17: don't we need to add the region to the equals method too?

depombo (approved):

src/modules/0.0.21/aws_elasticache/index.ts
dfellis: Oh, yes. It probably didn't do anything in the update path here.

PR #1310, opened 2022-09-22
[wip] multi-region memory db

aguillenv:


src/modules/0.0.21/aws_memory_db/entity/subnet_group.ts
```diff
@@ -20,4 +20,13 @@ export class SubnetGroup {
 
   @Column('varchar', { array: true, nullable: true })
   subnets?: string[];
+
+  // This column is joined to `aws_regions` manually via hooks in the `../sql` directory
+  @Column({
+    primary: true,
```
dfellis, line 26: Why is this being marked as a primary key?

src/modules/0.0.21/aws_memory_db/entity/memory_db_cluster.ts
```diff
@@ -56,13 +56,20 @@ export class MemoryDBCluster {
...
-  })
+  @JoinColumn([
+    {
+      name: 'subnet_group',
+      referencedColumnName: 'subnetGroupName',
+    },
+    {
+      name: 'region',
+      referencedColumnName: 'region',
+    },
```
dfellis, line 71: Does this automatically create a `region` column under the hood, or do you need to also define a `region` column in the class? If it's the former, *how* do you get access to the `region` here? Always through the `subnetGroup`? And if that's the answer, what do we do if the join is nullable?

aguillenv: All assumptions are correct and it will fail or not work for nullable joins. It will also fail in the case where you have multiple joins because it will be hard to follow which one is the source of truth. 

Maybe a better solution is always to have the `region` defined as an FK to `aws_regions`, and additionally, we add a manual constraint that forces the region equality between all the related n entities.

dfellis: What does `TypeORM` generate in the migration file if you just copy-paste the `region: string` definition like you have in the `SubnetGroup` class along with this definition? Perhaps it creates both FK relations?

aguillenv:
Going to work on this one from scratch in a new PR

PR #1311, opened 2022-09-22
Auto vacuum `iasql_rpc`

### LINKED ISSUE #1192: Implement custom functions RFC

aguillenv:
Resolves #1192


src/modules/0.0.21/iasql_functions/sql/create_fns.sql
```diff
@@ -29,6 +29,27 @@ exception when others then
...
+begin
+  DELETE FROM iasql_rpc
+  WHERE NOT (
+    opid = any(
+      array(
+        SELECT opid
+        FROM iasql_rpc
+        WHERE start_date >= CURRENT_DATE - INTERVAL '6 months'
+        ORDER BY start_date DESC
+        LIMIT 4999
```
dfellis, line 46: Any particular reason for this limit?

dfellis (approved):

src/modules/0.0.21/iasql_functions/sql/create_fns.sql
aguillenv: No particular reason. The idea, for now, is to have a max of 5000 records in the table, and since this is being executed pre-rpc run, so one less here. If I remember correctly some RDBMS always add a limit of the first couple of thousand records when you inspect data. If we want to increase the records we want to maintain in the database we will need to see how the performance is affected or change the query approach if necessary.

dfellis (approved):

yrobla (approved):

PR #1313, opened 2022-09-23
add prettier sql plugin

### LINKED ISSUE #1178: SQL style rules
It would be great to add on to [this](https://github.com/iasql/iasql-engine/pull/1157), and get style rules for the SQL files in the engine

aguillenv:
Resolves #1178 

It is not perfect since it cannot parse the content of the `plpgsql` functions, technically that's another format different from `sql`, but it at least gives us some consistency in these files.


aguillenv:
The `changes-in-latest-version` check is failing but I think we can make an exception here since it is formatting the code.

dfellis:
Did you try out [this plugin](https://www.npmjs.com/package/@potygen/prettier-plugin-pgsql)?

dfellis (approved):

depombo (approved):

Issue #867, opened 2022-06-17
Security Group Rules support for other Security Groups as a target for the rule (instead of CIDR IP address directives)


dfellis:

dfellis:
Nope. I did an impromptu demo last night and discovered that not only do we not support it, we accidentally create an invalid rule record in the database instead if the AWS account is using a rule with this feature.

PR #1319, opened 2022-09-24
 fix: add source security group as option for security rules

yrobla:
The security rules can be created either by ip or by pointing
    to a security group. Add this option to the module
    
    Closes: #867


test/modules/aws-security-group-integration.ts
```diff
@@ -57,7 +57,7 @@ describe('Security Group Integration Testing', () => {
 
   it('installs the security group module', install(modules));
 
-  it(
+  /*it(
```
dfellis: These tests are going to be restored before merging, right?

yrobla: Sure ! I just commented to accelerate the testing, but i will readd . The way of mapping in the entity and the whole workflow makes sense to you ?

src/modules/0.0.21/aws_security_group/entity/index.ts
```diff
@@ -97,4 +106,13 @@ export class SecurityGroupRule {
...
   description?: string;
+
+  @Check(
+    'Check_security_or_ip_permissions',
+    `("source_security_group" IS NULL AND ("from_port" IS NOT NULL OR "to_port" IS NOT NULL OR "cidr_ipv4" IS NOT NULL OR "cidr_ipv6" IS NOT NULL)) OR ("source_security_group" IS NOT NULL AND ("from_port" IS NULL AND "to_port" IS NULL AND "cidr_ipv4" IS NULL AND "cidr_ipv6" IS NULL))`,
+  )
+  @Column({
+    nullable: true,
+  })
+  sourceSecurityGroup?: string;
```
dfellis, line 117: Why is this not a `SecurityGroup` entity here?

yrobla: That's the first option i tried.. but check https://github.com/iasql/iasql-engine/actions/runs/3125493049/jobs/5069921967#step:7:16399 
For some reason, it is not populating the SourceSecurityGroup (it doesn't even show the field as undefined). I was wondering if that could be caused by having 2 fks to the same table... so after quite a lot of tries, i ended just passing the name as string. Do you know what can be causing that?

aguillenv: when you tried was the `eager` option set to `true`?
```
@Column({
    nullable: true,
    eager: true,
  })
  sourceSecurityGroup?: SecurityGroup;
```

yrobla: i updated to use eager:true and then it brings the entity. So far i am not facing any problems with recursion

src/modules/0.0.21/aws_security_group/entity/index.ts
```diff
@@ -97,4 +106,14 @@ export class SecurityGroupRule {
     nullable: true,
   })
   description?: string;
+
+  @Check(
+    'Check_security_or_ip_permissions',
+    `("source_security_group" IS NULL AND ("from_port" IS NOT NULL OR "to_port" IS NOT NULL OR "cidr_ipv4" IS NOT NULL OR "cidr_ipv6" IS NOT NULL)) OR ("source_security_group" IS NOT NULL AND (("from_port" IS NULL OR "from_port"=-1) AND ("to_port" IS NULL OR "to_port"=-1) AND ("cidr_ipv4" IS NULL OR "cidr_ipv4"='0.0.0.0/0') AND ("cidr_ipv6" IS NULL)))`,
```
aguillenv, line 112: two points here that confuse me:
- in the first part when `"source_security_group" IS NULL`. Is it possible to have only `from_port` defined and not the rest shouldn't they be `AND`s instead of `OR`s?
- in the second part, when checking the `cidr_ipv4` `NULL` or `0.0.0.0/0`  aren't they completely the opposite? is null or everything can pass?

aguillenv (commented):
security groups can self-reference to themselves, right? Maybe we can add another test to have that case handled.

src/modules/0.0.21/aws_security_group/entity/index.ts
yrobla: ok so first part, i think that i could refine a bit more, and set from_port IS NOT NULL AND to_port IS NOT NULL AND (cidr_ipv4 IS NOT NULL OR cidr_ipv6) IS NOT NULL. I will test that way
so 0.0.0.0/0 is the default value returned by AWS when there is a rule with a source security group. They are sending -1 on the ports, and 0.0.0.0/0 on cidr_ipv4

test/modules/aws-security-group-integration.ts
```diff
@@ -19,53 +29,81 @@ beforeAll(async () => await execComposeUp());
...
+    query(
+      `
     INSERT INTO aws_credentials (access_key_id, secret_access_key)
     VALUES ('${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}')
-  `, undefined, false));
+  `,
+      undefined,
+      false,
+    ),
+  );
```
dfellis, line 47: Why do these test files always get this ugly formatting with the new formatter?

test/modules/aws-security-group-integration.ts
```diff
@@ -131,36 +196,349 @@ describe('Security Group Integration Testing', () => {
...
+    'check no security group for source remain',
+    query(
+      `
+    SELECT *
+    FROM security_group
+    WHERE group_name = '${prefix}sgsourcetest' OR group_name='${prefix}sgforsource';
+  `,
+      (res: any[]) => expect(res.length).toBe(0),
+    ),
+  );
```
dfellis, line 514: I would like to propose one more set of security group tests. Often security group rules are written such that security group A is only accessible to security group B *and* vice versa, such that there is a recursive-looking intertwined situation between them.

It is possible to create these in the database by creating security group A, then security group B, then security group A's rule points at B and security group B's rule points at A, and to delete them in the reverse order, but making sure that the engine doesn't trip up on this is important.

Can we have a set of six `it` statements like this?

1. A query with four separate `INSERT`s to create these self-referential security groups.
2. An `apply` (confirming that it isn't tripped up by this structure)
3. A `SELECT` query to confirm these records do exist with `expect` assertions on it (likely something like `SELECT sg.id, sgr.source_security_group FROM security_group sg INNER JOIN security_group_rule sgr ON sg.id = sgr.security_group_id WHERE sgr.source_security_group IS NOT NULL`, assert that there should be two records, and that the `id` and `source_security_group` fields are opposite paired values between the two)
4. Four `DELETE` query to delete the security group rules *and* the security groups
5. Another `apply` (confirming that it can likewise drop these rules "atomically" and you don't have to delete one-at-a-time manually.
6. Another `SELECT` confirming that they are really gone.

test/modules/aws-security-group-integration.ts
```diff
@@ -532,6 +534,103 @@ describe('Security Group Integration Testing', () => {
 
   it('applies the vpc removal', apply());
 
+  // tests cycle in security rules
+  it(
+    'adds a new security group A',
```
dfellis, line 539: Nit.

```suggestion
    'adds a new security groups A and B',
```

test/modules/aws-security-group-integration.ts
```diff
@@ -532,6 +534,103 @@ describe('Security Group Integration Testing', () => {
...
+  `),
+  );
+  it(
+    'adds security group rules pointing to the other ones (B)',
+    query(`
+  INSERT INTO security_group_rule(description, security_group_id, source_security_group, is_egress) 
+  VALUES ('${prefix}sgtestB', (SELECT id FROM security_group WHERE group_name='${prefix}sgtestB'),
+  (SELECT id FROM security_group WHERE group_name='${prefix}sgtestA'), false);
+  `),
+  );
```
dfellis, line 561: Also, this could be condensed to just

```suggestion
  it(
    'adds self-referential security groups A and B',
    query(`  
    INSERT INTO security_group (description, group_name)
    VALUES ('Security Group Test A', '${prefix}sgtestA'), ('Security Group Test B', '${prefix}sgtestB');
    
    INSERT INTO security_group_rule(description, security_group_id, source_security_group, is_egress) 
  VALUES ('${prefix}sgtestA', (SELECT id FROM security_group WHERE group_name='${prefix}sgtestA'),
  (SELECT id FROM security_group WHERE group_name='${prefix}sgtestB'), false);
  
  INSERT INTO security_group_rule(description, security_group_id, source_security_group, is_egress) 
  VALUES ('${prefix}sgtestB', (SELECT id FROM security_group WHERE group_name='${prefix}sgtestB'),
  (SELECT id FROM security_group WHERE group_name='${prefix}sgtestA'), false);
  `),
  );
```

test/modules/aws-security-group-integration.ts
```diff
@@ -532,6 +534,103 @@ describe('Security Group Integration Testing', () => {
...
+
+  it('creates the source groups and rules', apply());
+
+  it(
+    'check that security group and rules are correctly created - tcp, udp, icmp',
+    query(
+      `
+      SELECT sg.id, sgr.source_security_group FROM security_group sg INNER JOIN security_group_rule sgr ON sg.id = sgr.security_group_id WHERE sg.group_name='${prefix}sgtestA' AND sgr.source_security_group=(SELECT id FROM security_group WHERE group_name='${prefix}sgtestB')`,
+      (res: any[]) => {
+        expect(res.length).toBe(3);
```
dfellis, line 571: Why is the length 3? For this particular query I would only expect one matching row?

yrobla: so when you create a rule with source security, aws expands it to 3 rules... one for tcp, another for udp and another for icmp

test/modules/aws-security-group-integration.ts
```diff
@@ -532,6 +534,103 @@ describe('Security Group Integration Testing', () => {
...
+      },
+    ),
+  );
+
+  it(
+    'deletes the source security group rule for A',
+    query(
+      `DELETE FROM security_group_rule WHERE source_security_group = (SELECT id FROM security_group WHERE group_name='${prefix}sgtestA')`,
+    ),
+  );
```
dfellis, line 591: Why are you deleting these security group rules before `iasql_apply()`ing them into the cloud? That doesn't actually test anything we need to test.

dfellis (approved):

PR #1329, opened 2022-09-26
Tune Django example: Add the checks to prevent container from exiting on ECS

mtp1376:
Since the `infra` database is not being used on the target ECS container, I've excluded it when the corresponding environment variables are not available.

### DIFF
--- a/examples/ecs-fargate/django/app/app/settings.py
+++ b/examples/ecs-fargate/django/app/app/settings.py
@@ -83,7 +83,10 @@ DATABASES = {
         'ENGINE': 'django.db.backends.sqlite3',
         'NAME': BASE_DIR / 'db.sqlite3',
     },
-    'infra': {
+}
+
+if ENV.get_value('DB_USER', default=None):  # being run to provision resources, not on the deployed ECS container
+    DATABASES['infra'] = {
         'ENGINE': 'django.db.backends.postgresql',
         'NAME': ENV('DB_NAME'),
         'USER': ENV('DB_USER'),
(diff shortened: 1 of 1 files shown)


examples/ecs-fargate/django/app/app/settings.py
```diff
@@ -83,15 +83,21 @@
...
+    ENV.get_value('DB_USER', default=None),
+    ENV.get_value('DB_PASSWORD', default=None)
+]):  # being run to provision resources, not on the deployed ECS container
+    DATABASES['infra'] = {
         'ENGINE': 'django.db.backends.postgresql',
         'NAME': ENV('DB_NAME'),
         'USER': ENV('DB_USER'),
         'PASSWORD': ENV('DB_PASSWORD'),
         'HOST': ENV.get_value('DB_HOST', default='db.iasql.com'),
         'PORT': ENV.get_value('DB_PORT', cast=int, default=5432),
```
aguillenv, line 99: Since these are not used in the CI and are meant to be used for final users in production I think we should restore them:
```suggestion
        'HOST': 'db.iasql.com',
        'PORT': '5432',
```

mtp1376: Actually they are being used in the CI job.
https://github.com/iasql/iasql-on-iasql/pull/110/files#diff-c0d9756ec1e666a5c4b1ab6fa27d5662156838384a67802e705e4de44d6ab119R107-R108

aguillenv (approved):

Issue #1328, opened 2022-09-26
Support attaching existing policies or permissions to IAM users


depombo:
Similarly to https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.21/aws_iam/entity/role.ts#L34

yrobla:
There is no policies for IAM users. I will need to add the PermissionsBoundary entity: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-iam/interfaces/attachedpermissionsboundary.html#permissionsboundaryarn

PR #1332, opened 2022-09-27
feat: add attached policies support to users

yrobla:
Allow to attach and detach policies for users on creation or deletion

Closes: #1328


yrobla:
i added the same level of support that is at role level. But it does not include support for update, just creation/deletion. Please let me know if i should add support for update policies as well, both on role and user

yrobla:
starting from a clean point

PR #1336, opened 2022-09-27
Multi-region 'aws_rds'

dfellis:
Also made `@cloudId` work for multiple columns. It *should* be backwards compatible with entities that have only a single `@cloudId` (or basically every test will fail).


src/modules/interfaces.ts
```diff
@@ -260,7 +260,13 @@ export class Mapper2<E> {
           .map(c => c.propertyName)
           .shift() ?? '';
       // Using + '' to coerce to string without worrying if `.toString()` exists, because JS
-      this.entityId = (e: E) => ((e as any)[cloudColumn] ?? (e as any)[primaryColumn]) + '';
+      this.entityId = (e: E) => {
+        if (cloudColumns && !(cloudColumns instanceof Error)) {
+          return cloudColumns.map(col => (e as any)[col]).join('|');
```
aguillenv, line 265: can we assume the order will be always the same? or should we add some kind of sorting logic to the decorator?

dfellis: So the order of the array is determined by the order of the execution of the decorators, which is static and done only once, so yes, the order is the same every time we run here.

Even if the order is different between runs of the engine process itself, there would be no `memo` cache data to contradict it, and these `entityId`s are never stored in the database, so there is no vulnerability possible here.

src/modules/0.0.21/aws_rds/entity/rds.ts
```diff
@@ -99,8 +99,24 @@ export class RDS {
...
+  @JoinColumn([
+    {
+      name: 'parameter_group_id',
+      referencedColumnName: 'id',
+    },
+    {
+      name: 'region',
+      referencedColumnName: 'region',
+    },
+  ])
```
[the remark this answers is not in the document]
dfellis: I added the `region` column to implicitly restrain it, to make sure that you don't accidentally join an RDS instance to the wrong parameter group.

src/modules/interfaces.ts
dfellis: Granted, digging in here is a good idea, though, since there is *something* wrong in this PR that I haven't figured out, yet.

src/modules/0.0.22/aws_rds/entity/parameter_group.ts
```diff
@@ -58,9 +58,14 @@ export enum ParameterGroupFamily {
 }
 
 @Entity()
+@Unique('paragrp_name_region', ['name', 'region'])
+@Unique('paragrp_id_region', ['id', 'region'])
 export class ParameterGroup {
-  @PrimaryColumn()
+  @PrimaryGeneratedColumn()
+  id?: number;
```
mtp1376, line 65: `@PrimaryGeneratedColumn` should always have a value, shouldn't it? But it's nullable here. And it's having a `@ Unique` constraint together with `region` which is strange since `id` should always be unique, and combining it with anything would make that pair unique as well.

src/modules/0.0.22/aws_rds/entity/parameter_group.ts
```diff
@@ -83,4 +88,13 @@ export class ParameterGroup {
     nullable: true,
   })
   parameters?: Parameter[];
+
+  // This column is joined to `aws_regions` manually via hooks in the `../sql` directory
```
mtp1376, line 92: Just a minor comment, you can use Alejandro's approach to avoid having `{before,after}_install.sql` scripts.
https://github.com/iasql/iasql-engine/pull/1364/files#diff-610483bfe761aea785477f5c21be25ad1405dabde226aafe32780d230bd71145R133-R141

src/modules/0.0.22/aws_rds/entity/rds.ts
```diff
@@ -14,15 +15,12 @@ import { AvailabilityZone } from '../../aws_vpc/entity';
 import { ParameterGroup } from './parameter_group';
 
 @Entity()
+@Unique('UQ_identifier_region', ['dbInstanceIdentifier', 'region'])
```
mtp1376, line 18: It's unlikely that the `dbInstanceIdentifier`s collide despite having multi-region, but I think it's a good idea to keep it unique.

src/modules/0.0.22/aws_rds/entity/parameter_group.ts
dfellis: Oh, right! Will make that update.

src/modules/0.0.22/aws_rds/entity/parameter_group.ts
dfellis: So two things:

1. We mark it as nullable here despite it always coming back with a value from TypeORM because we construct these objects with our mapper code and they won't have database IDs, which is why that's that way.
2. Both of those `@Unique` decorators being there must be some sort of accidental rebase glitch? I originally did `['id', 'region']` then remembered it should have been `['name', 'region']` and changed it, and I am just as surprised as you that `['id', 'region']` somehow got back into the codebase.

src/modules/0.0.22/aws_rds/entity/rds.ts
dfellis: So users choose the [`dbInstanceIdentifier` names themselves](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-rds/interfaces/createdbinstancecommandinput.html#dbinstanceidentifier) which means the user accidentally causing a collision is something we should guard against, I think?

What we also need, though, is a check constraint with all of those weird rules on what values are valid. Not sure if it should be in this PR or not, though.

src/modules/0.0.22/aws_rds/entity/parameter_group.ts
dfellis: Wait, no. I take that back, but I will add a comment on this. There's a really weird Postgres unique constraint error if I don't do this: https://github.com/iasql/iasql-engine/actions/runs/3216578406/jobs/5258602677#step:5:7128 because of how the RDS entity joins on the ParameterGroup entry.

yrobla (approved):

src/modules/0.0.22/aws_rds/entity/rds.ts
mtp1376: Yes. You're right. I thought it's being chosen by the cloud.

src/modules/0.0.22/aws_rds/entity/parameter_group.ts
mtp1376: Got it. So if you do the join by those two columns, you have to explicitly define a unique constraint. Somehow weird but in some sense it also makes sense :D .

mtp1376 (approved):

src/modules/0.0.22/aws_rds/index.ts
```diff
@@ -260,26 +268,30 @@ class RdsMapper extends MapperBase<RDS> {
...
-        const rdses = await this.getDBInstances(client.rdsClient);
         const out = [];
-        for (const rds of rdses) {
-          out.push(await this.rdsMapper(rds, ctx));
+        for (const region of enabledRegions) {
+          const client = (await ctx.getAwsClient(region)) as AWS;
+          const rdses = await this.getDBInstances(client.rdsClient);
+          for (const rds of rdses) {
+            out.push(await this.rdsMapper(rds, ctx, region));
+          }
```
aguillenv, line 285: let's do this in parallel with `await.Promise.all()` like here https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_api_gateway/index.ts#L79L91

src/modules/0.0.22/aws_rds/index.ts
```diff
@@ -443,33 +457,47 @@ class ParameterGroupMapper extends MapperBase<ParameterGroup> {
...
-        const parameterGroups = await this.getDBParameterGroups(client.rdsClient);
         const out = [];
-        for (const pg of parameterGroups) {
-          out.push(this.parameterGroupMapper(pg));
+        for (const region of enabledRegions) {
+          const client = (await ctx.getAwsClient(region)) as AWS;
+          const parameterGroups = await this.getDBParameterGroups(client.rdsClient);
+          for (const pg of parameterGroups) {
+            out.push(this.parameterGroupMapper(pg, region));
+          }
```
aguillenv, line 483: let's do this in parallel with `await.Promise.all()` like here https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_api_gateway/index.ts#L79L91

src/modules/0.0.22/aws_rds/index.ts
dfellis: That inner `.map` should be a `.forEach`, but got it.

aguillenv (approved):

PR #1348, opened 2022-09-28
Restart scheduler on `upgrade`

aguillenv:
Issue: After the upgrade, the engine tried calling new RPC methods with an old database connection.
The upgrade test was not catching the issue because it was not calling a new RPC after the upgrade finished.


src/services/scheduler.ts
```diff
@@ -152,6 +152,7 @@ export async function start(dbId: string, dbUser: string) {
       },
       rpc: async (payload: any) => {
         const { params, opid, modulename, methodname } = payload;
+        const rpcConn = await TypeormWrapper.createConn(dbId, { name: uuidv4() });
```
depombo, line 155: hmm but doesn't `upgrade` call `start` at all? making sure that it does seems like a more resource-efficient way of fixing this?

aguillenv: no, upgrade just uninstalls and installs

depombo: perhaps we should have upgrade stop and start the scheduler too? creating a new connection for every call is not what we did on the old path and I am somewhat concerned that it would impose limitations at scale. could we do this another way?

aguillenv: I've been trying that start/stop path today, since creating a new connection we need to drop it at the end and that was causing some long-running/infinite runs for some cases... But I'm not fully sure why it is not working, is like it is not getting called...

src/services/scheduler.ts
```diff
@@ -16,12 +16,18 @@ import MetadataRepo from './repositories/metadata';
...
+enum IasqlOperationType {
+  APPLY = 'APPLY',
+  SYNC = 'SYNC',
+  INSTALL = 'INSTALL',
+  UNINSTALL = 'UNINSTALL',
+  PLAN_APPLY = 'PLAN_APPLY',
+  PLAN_SYNC = 'PLAN_SYNC',
+  LIST = 'LIST',
+  UPGRADE = 'UPGRADE',
+}
```
aguillenv, line 30: I need to do this to avoid cycle importing files

src/services/telemetry.ts
```diff
@@ -2,19 +2,20 @@ import * as Amplitude from '@amplitude/node';
...
+enum IasqlOperationType {
+  APPLY = 'APPLY',
+  SYNC = 'SYNC',
+  INSTALL = 'INSTALL',
+  UNINSTALL = 'UNINSTALL',
+  PLAN_APPLY = 'PLAN_APPLY',
+  PLAN_SYNC = 'PLAN_SYNC',
+  LIST = 'LIST',
+  UPGRADE = 'UPGRADE',
+}
```
aguillenv, line 18: I need to do this to avoid cycle importing files

src/services/scheduler.ts
```diff
@@ -261,22 +281,37 @@ async function getContext(conn: TypeormWrapper, Modules: any): Promise<Context>
...
+    await runner.stop();
+  } catch (e) {
+    logger.warn(
+      `Graphile workers for ${dbId} has already been stopped. Perhaps Kubernetes is going to restart the process?`,
+      { e },
+    );
+  }
+}
+
+async function stopConnection(conn: any, dbId: string) {
```
depombo, line 303: nit - perhaps rename to `stopConnServer`?

depombo (approved):

dfellis (approved):

PR #1349, opened 2022-09-29
Restore symmetry to install/uninstall hooks and use it to fix migration issues

dfellis:


dfellis:
# More Hooks!

Captain Hook

yrobla (approved):

PR #1350, opened 2022-09-29
 feat: add codedeploy module coverage

yrobla:
Create initial version of AWS Codedeploy module, that will be
    capable of deploying to ec2
    
    Closes: #1258


src/modules/0.0.22/aws_codedeploy/entity/revision.ts
```diff
@@ -0,0 +1,52 @@
...
+export enum RevisionType {
+  S3 = 'S3',
+  GITHUB = 'Github',
+}
+
+@Entity()
+export class CodedeployRevision {
+  @PrimaryGeneratedColumn()
+  @cloudId
+  id: number;
```
aguillenv, line 15: yes, this I'm not so sure about, If I'm not wrong, basically for every loop in the `apply` call, for every record in the db it will think it needs to be created, because it did not encounter its counterpart, and same for every cloud record, it will think it needs to be deleted.

dfellis (approved):

src/modules/0.0.22/aws_codedeploy/entity/deploymentGroup.ts
```diff
@@ -0,0 +1,71 @@
...
+@Entity()
+export class CodedeployDeploymentGroup {
+  @PrimaryColumn()
+  @cloudId
+  name: string;
+
+  @Column({
+    nullable: true,
+  })
+  id?: string;
```
aguillenv, line 29: can we rename this? we have been using `id` for the autogenerated primary db id which could create confusion.

aguillenv (approved):

src/modules/0.0.22/aws_codedeploy/entity/deployment.ts
```diff
@@ -0,0 +1,83 @@
...
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+
+import { cloudId } from '../../../../services/cloud-id';
+import { CodedeployApplication, ComputePlatform } from './application';
+import { CodedeployDeploymentGroup } from './deploymentGroup';
+import { CodedeployRevision } from './revision';
+
+export enum DeploymentStatusEnum {
+  BAKING = 'Baking',
```
depombo, line 17: baking?

src/modules/0.0.22/aws_codedeploy/sql/after_install.sql
```diff
@@ -0,0 +1,140 @@
+ -- prevents update on revision table
```
depombo, line 1: interesting, I think historically we leave this up to the mapper so on apply it puts the record back based on what is in the cloud. I'm curious, why did you decide to do it this way?

test/modules/aws-codedeploy-integration.ts
```diff
@@ -0,0 +1,618 @@
...
+  },
+});
+
+const revisionLocationv1 = JSON.stringify({
+  revisionType: 'GitHub',
+  gitHubLocation: {
+    repository: 'iasql/iasql-codedeploy-example',
+    commitId: '165582e107955f0b114a9d9d74cd2e4f198454a7',
+  },
+});
```
depombo, line 119: how is authentication working atm?

src/modules/0.0.22/aws_codedeploy/entity/deployment.ts
yrobla: baking yes! :) 
Ask AWS...

test/modules/aws-codedeploy-integration.ts
yrobla: no authentication, we can just use public repos

src/modules/0.0.22/aws_codedeploy/sql/after_install.sql
yrobla: i have no way to identify a revision... there is no single id, arn, name... i can use. So i thought it would be better to prevent at SQL , because i do not have a trusted way to do it from our mappers.

src/modules/0.0.22/aws_codedeploy/sql/after_install.sql
```diff
@@ -0,0 +1,140 @@
...
+        RAISE EXCEPTION 'Revision cannot be deleted'
+        USING detail = 'Revisions can only be added';
+        RETURN OLD;
+    END;
+$block_codedeploy_revision_delete_function$ LANGUAGE plpgsql;
+
+CREATE TRIGGER
+  block_codedeploy_revision_delete_trigger BEFORE DELETE ON codedeploy_revision
+EXECUTE
+  FUNCTION block_codedeploy_revision_delete_function ();
```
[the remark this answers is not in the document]
yrobla: it cannot delete rows created by the user, it only allows deletion on cascade. So the rows would only be deleted when the application is. I had to do that way because i do not have a simple way to identify an unique revision. They do not have any ID, any arn, name... you can have 2 revisions totally identical, the only thing that could be different is the creation date.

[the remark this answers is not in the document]
dfellis: But this would put the database into a broken state where apply and sync don't function anymore, correct?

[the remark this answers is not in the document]
yrobla: not sure if i follow you... people can only add new revisions if they are associated with an application. I also do not reconcile on updates, just on inserts. In the application i only reconcile when a new revision is added to the cloud, i check the length of the revisions array. However i'm happy to hear about suggestions on how to improve it , or on new tests i could add to improve coverage..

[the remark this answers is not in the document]
yrobla: mm are you worried about the atomicity of these rules? if so, do you know a better way to block deletes but allow deletes on cascade?

yrobla:
I refactored the module, to make revision part of a deployment. This is more clean now

aguillenv (approved):

src/modules/0.0.22/aws_codedeploy/sql/after_install.sql
```diff
@@ -0,0 +1,18 @@
...
+    RETURN '#!/bin/bash
+sudo apt update
+sudo apt -y install ruby-full
+sudo apt -y install wget
+cd /home/ubuntu
+wget https://aws-codedeploy-' || region || '.s3.' || region || '.amazonaws.com/latest/install
+chmod +x ./install
+sudo ./install auto
+sudo service codedeploy-agent start';
+  END case;
```
dfellis, line 16: Should we add a default case that [raises an exception](https://www.postgresql.org/docs/current/plpgsql-errors-and-messages.html) that it is not implemented?

dfellis (approved):

PR #1354, opened 2022-09-30
Define `default_aws_region` before install

aguillenv:


mtp1376 (approved):
Confirmed this will fix the `default_aws_region()` does not exist error.

dfellis:
I thought @aguillenv had an example of how to define the `region` schema change in the entities, which would allow us to remove it from the `after_install.sql` files and making this moot?

Also, this restores the "insert random schema removal code you have to manually delete" behavior that we're trying to avoid?

aguillenv:
> I thought @aguillenv had an example of how to define the `region` schema change in the entities, which would allow us to remove it from the `after_install.sql` files and making this moot?
> 
> Also, this restores the "insert random schema removal code you have to manually delete" behavior that we're trying to avoid?

Oh, we need both, actually. We need this first because the `default_aws_region()` does not exist if the after_install is not executed and then we need to add what I did with the decorators to avoid the after-install FKs and the garbage generated for other modules.

dfellis:
> > I thought @aguillenv had an example of how to define the `region` schema change in the entities, which would allow us to remove it from the `after_install.sql` files and making this moot?
> > Also, this restores the "insert random schema removal code you have to manually delete" behavior that we're trying to avoid?
> 
> Oh, we need both, actually. We need this first because the `default_aws_region()` does not exist if the after_install is not executed and then we need to add what I did with the decorators to avoid the after-install FKs and the garbage generated for other modules.

Oh, we should just move the function definition to the `before_install.sql` file and leave the trigger in the `after_install.sql` file to fix that issue. I demonstrated (I think I put it in the Discord channel) that you can write a function that references non-existent schema and Postgres will accept it, give you an error if you call it then, the give you what you expect after the schema is created.

aguillenv:
> Oh, we should just move the function definition to the `before_install.sql` file and leave the trigger in the `after_install.sql` file to fix that issue. I demonstrated (I think I put it in the Discord channel) that you can write a function that references non-existent schema and Postgres will accept it, give you an error if you call it then, the give you what you expect after the schema is created.

That would work, let me test it and create a PR for that.

dfellis (approved):

yrobla (approved):

PR #1356, opened 2022-10-01
Multi-region support for `aws_acm_import`

mtp1376:


dfellis (approved):

src/modules/0.0.22/aws_acm_import/entity/certificate_import.ts
```diff
@@ -14,4 +14,12 @@ export class CertificateImport {
...
   @Column({ nullable: true })
   chain?: string;
+
+  // This column is joined to `aws_regions` manually via hooks in the `../sql` directory
+  @Column({
+    type: 'character varying',
+    nullable: false,
+    default: () => 'default_aws_region()',
+  })
+  region: string;
```
aguillenv, line 24: nit: we could avoid the `after_install.sql` and `before_uninstall.sql` defining the region like this https://github.com/iasql/iasql-engine/pull/1364/files#diff-610483bfe761aea785477f5c21be25ad1405dabde226aafe32780d230bd71145R133-R141

aguillenv (approved):

PR #1367, opened 2022-10-05
Make region read calls parallel + standardize multi-region approach

aguillenv:
This should speed up the tests a little bit


yrobla (approved):

src/modules/0.0.22/aws_cloudwatch/index.ts
```diff
@@ -58,23 +58,25 @@ class LogGroupMapper extends MapperBase<LogGroup> {
     },
     read: async (ctx: Context, id?: string) => {
       const enabledRegions = (await ctx.getEnabledAwsRegions()) as string[];
-      if (id) {
-        for (const region of enabledRegions) {
+      if (!!id) {
+        const [groupName, region] = id.split('|');
```
depombo, line 62: maybe this should be a stateless/static getter function in the cloud id service?

src/modules/0.0.22/aws_elasticache/index.ts
```diff
@@ -112,36 +109,38 @@ class CacheClusterMapper extends MapperBase<CacheCluster> {
...
+        );
         return out;
       }
     },
     update: async (clusters: CacheCluster[], ctx: Context) => {
       // if user has modified state, restore it. If not, go with replace path
       const out = [];
       for (const cluster of clusters) {
-        const cloudRecord = ctx?.memo?.cloud?.CacheCluster?.[cluster.clusterId ?? ''];
+        const cloudRecord = ctx?.memo?.cloud?.CacheCluster?.[this.entityId(cluster)];
```
depombo, line 143: oh I didn't know this was a thing. this extracts the entity?

test/modules/aws-secret-integration.ts
```diff
@@ -143,19 +143,6 @@ describe("Secrets Manager Integration Testing", () => {
     )
   );
 
-  it('tries to move the secret to another region', query(`
```
depombo, line 146: why did this behavior change?

src/modules/0.0.22/aws_cloudwatch/index.ts
dfellis: Agreed. We should have it as a built-in auto-defined method like `entityId` that returns an object of field -> value pieces. This is completely doable since the fields are stored in an array in the same order as the `id.split('|')` output [here](https://github.com/iasql/iasql-engine/blob/main/src/services/cloud-id.ts#L1).

Perhaps it could be called `this.idFields(id: string)`?

src/modules/0.0.22/aws_elasticache/index.ts
dfellis: Yep, if you use `@cloudId`, under the hood it is implementing the `entityId` method for you.

src/modules/0.0.22/aws_cloudwatch/index.ts
aguillenv: agree! I can add that in this PR

dfellis (approved):

PR #1374, opened 2022-10-08
Add multi-region support for `aws_acm_list` and `aws_acm_request`

mtp1376:


mtp1376:
Seems like the integration tests can't be done just by using the `acm_list` module. I'll add more when submitting the PR for `acm_request`.

src/modules/0.0.22/aws_acm_list/index.ts
```diff
@@ -94,18 +103,24 @@ class CertificateMapper extends MapperBase<Certificate> {
...
-        for (const rawCert of rawCerts) {
-          const cert = this.certificateMapper(rawCert);
-          if (cert) out.push(cert);
+        for (const region of enabledRegions) {
+          const client = (await ctx.getAwsClient(region)) as AWS;
+          const rawCerts = (await this.getCertificates(client.acmClient)) ?? [];
+          for (const rawCert of rawCerts) {
+            const cert = this.certificateMapper(rawCert, region);
+            if (cert) out.push(cert);
+          }
```
aguillenv, line 123: let's do this in parallel with `await.Promise.all()` like here https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_api_gateway/index.ts#L79L91

test/modules/aws-acm-request-integration.ts
```diff
@@ -190,6 +190,42 @@ describe('AwsAcmRequest Integration Testing', () => {
...
+  it('applies the deletion of the certificate in the non-default region', apply());
+
+  it('checks the deletion of the certificate in the non-default region', query(`
+      SELECT *
+      FROM certificate
+      WHERE domain_name = '${domainName}'
+        AND status = 'ISSUED'
+        AND region = 'us-east-1';
+  `, (res: any[]) => expect(res.length).toBe(0)));
+
```
aguillenv, line 228: does it make sense to have a test changing a certificate to another region?

src/modules/0.0.22/aws_acm_list/index.ts
mtp1376: Cool way. I'll switch to that.

test/modules/aws-acm-request-integration.ts
mtp1376: Changing a certificate region would be "to delete" and "to create" a certificate. Since ACM list module has nothing to do with the creation step, I think that should be done in ACM request module. The strange part is that the ACM request module removes the entry upon creation. So it should be in ACM list :))).

Basically when the `region` column in ACM list is changed, we "can" go and create an ACM certificate request for the user, but I'm not sure if it's a good idea. We might leave it to be manually done by the user. What's your own opinion on this?

aguillenv: Oh, that's true. Yes, maybe it is ok as-is now and we can improve or add some feedback in the future.

mtp1376:
> Not entirely sure why, but the second certificate issued in another region [is _deleted_ by the engine immediately after it is created](https://github.com/iasql/iasql-engine/actions/runs/3210853278/jobs/5248610401#step:7:20768) which is why the test suite then fails. I suspect something wrong with the equality checking?

I tried this test suite on my own domain on route53 twice, and it succeeded without a problem. I believe that problem was a noise in the tests.

mtp1376:
Seems like there's a problem with the `acm_request` module and it causes some residue DNS records to be left after the certificate is issued. Currently investigating it.

mtp1376:
OK seems like that bug fix has made the checks go green! 🚀

yrobla (approved):

src/modules/0.0.22/aws_acm_list/index.ts
```diff
@@ -19,19 +20,22 @@ class CertificateMapper extends MapperBase<Certificate> {
     Object.is(a.domainName, b.domainName) &&
     Object.is(a.inUse, b.inUse) &&
     Object.is(a.renewalEligibility, b.renewalEligibility) &&
-    Object.is(a.status, b.status);
+    Object.is(a.status, b.status) &&
+    Object.is(a.region, b.region);
```
dfellis, line 24: I noticed that the `entityId` method is only using the ARN for the cloud ID, so this addition here is correct, but are you sure you want `update` to run on a record that now spans two regions? The code to deal with that is more complex because you need to get two AWS clients, one for the old region and one for the new, to implement the `replace` functionality correctly.

dfellis: Oh, I see we just shove the state in the database back to whatever AWS has defined in the cloud, so it looks like that should still work fine here. Definitely not the "usual" pattern, though.

dfellis (approved):

src/modules/0.0.22/aws_acm_list/index.ts
```diff
@@ -1,4 +1,5 @@
 import { ACM, CertificateDetail, paginateListCertificates } from '@aws-sdk/client-acm';
+import { parse as parseArn } from '@aws-sdk/util-arn-parser';
```
aguillenv, line 2: interesting addition, I think in other modules we are manually splitting the ARN and this could be a good improvement.

aguillenv (approved):

PR #1385, opened 2022-10-10
Complete multi-region work for 'aws_vpc'

dfellis:


dfellis:
There is one last behavior that I am not sure if it should be handled for the user in the mapper or if it's related to a future refactor an existing TODO is already mentioning and should be tackled in a follow-up PR.

If you "move" an Endpoint Gateway between regions, the auto-populated `routeTableIds` array needs to be `NULL`ed out or it will fail because it is referencing things in the old region and then fails. This PR could be modified to filter out invalid route table IDs instead of failing on it, but that could cover up other mistakes by the user?

There's a TODO stating that the route table ought to be turned into an entity in its own right, and that would make this go away (because we could constrain on the region and force a DB error if you don't detach them before changing the region, and we can also guarantee that the values in there are "real" in a way we currently can't). It should eventually be done, but whether the current behavior is good enough or not is still debatable.

src/modules/0.0.22/aws_vpc/mappers/endpoint_gateway.ts
```diff
@@ -181,29 +182,43 @@ export class EndpointGatewayMapper extends MapperBase<EndpointGateway> {
...
         return out;
       }
     },
     updateOrReplace: (a: EndpointGateway, b: EndpointGateway) => {
-      if (!(Object.is(a.vpc?.vpcId, b.vpc?.vpcId) && Object.is(a.service, b.service))) return 'replace';
+      if (
+        !(
+          Object.is(a.vpc?.vpcId, b.vpc?.vpcId) &&
+          Object.is(a.service, b.service) &&
+          Object.is(a.region, b.region)
```
aguillenv, line 211: nit: since region is defined as part of the `cloudId` this would never happen

aguillenv (approved):

depombo (approved):

mtp1376 (approved):

Issue #1388, opened 2022-10-12
Cannot delete an s3 bucket when it has contents


yrobla:
The delete call needs that the bucket is empty before deleting it, otherwise it cannot delete.
We need to have a way to delete bucket objects in order to delete non-empty buckets

PR #1389, opened 2022-10-12
fix: add rpc to clean s3 buckets

yrobla:
It was not possible to delete a bucket with contents on it so we are creating an rpc function to delet the contents

Closes: #1388

### DIFF
--- a/src/modules/0.0.22/aws_s3/index.ts
+++ b/src/modules/0.0.22/aws_s3/index.ts
@@ -9,6 +9,7 @@ import { policiesAreSame } from '../../../services/aws-diff';
 import { AWS, crudBuilder2, crudBuilderFormat } from '../../../services/aws_macros';
 import { Context, Crud2, MapperBase, ModuleBase } from '../../interfaces';
 import { Bucket } from './entity';
+import { S3CleanBucketRpc } from './rpcs';
 
 class BucketMapper extends MapperBase<Bucket> {
   module: AwsS3Module;
@@ -148,12 +149,14 @@ class BucketMapper extends MapperBase<Bucket> {
   }
 }
 
-class AwsS3Module extends ModuleBase {
+export class AwsS3Module extends ModuleBase {
   bucket: BucketMapper;
+  s3CleanBucket: S3CleanBucketRpc;
 
   constructor() {
     super();
     this.bucket = new BucketMapper(this);
+    this.s3CleanBucket = new S3CleanBucketRpc(this);
     super.init();
   }
 }
--- a/src/modules/0.0.22/aws_s3/rpcs/s3_clean_bucket.ts
+++ b/src/modules/0.0.22/aws_s3/rpcs/s3_clean_bucket.ts
@@ -1,0 +1,33 @@
+import { AwsS3Module } from '..';
+import { Context, RpcBase, RpcResponseObject } from '../../../interfaces';
+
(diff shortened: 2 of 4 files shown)


test/modules/aws-s3-integration.ts
```diff
@@ -40,168 +46,261 @@ const updatedPolicyJSON = {
...
+      (res: any[]) => expect(res.length).toBe(1),
+    ),
+  );
+
+  it(
+    'cleans the bucket',
+    query(`SELECT * FROM s3_clean_bucket('${s3Name}')`, (res: any[]) =>
+      expect(res[0].status).toStrictEqual('OK'),
+    ),
+  );
```
aguillenv, line 183: maybe add a test checking the response of the call without bucket name

src/modules/0.0.22/aws_s3/rpcs/s3_clean_bucket.ts
```diff
@@ -0,0 +1,75 @@
...
+    res => res?.Contents ?? [],
+  );
+  deleteBucketObject = crudBuilder2<S3, 'deleteObject'>('deleteObject', (bucketName, key) => ({
+    Bucket: bucketName,
+    Key: key,
+  }));
+
+  call = async (
+    dbId: string,
+    dbUser: string,
```
aguillenv, line 27: nit: if they are not in use
```suggestion
    _dbId: string,
    _dbUser: string,
```

src/modules/0.0.22/aws_s3/rpcs/s3_clean_bucket.ts
```diff
@@ -0,0 +1,75 @@
...
+    ctx: Context,
+    ...params: string[]
```
aguillenv, line 29: I think we are already passing the params spread, so I think you could directly do
```suggestion
    bucketName: string
```

src/modules/0.0.22/aws_s3/index.ts
```diff
@@ -148,13 +149,15 @@ class BucketMapper extends MapperBase<Bucket> {
...
 
   constructor() {
     super();
     this.bucket = new BucketMapper(this);
+    this.s3CleanBucket = new S3CleanBucketRpc(this);
     super.init();
   }
 }
-export const awsS3Module = new AwsS3Module();
+export const awsS3 = new AwsS3Module();
```
aguillenv, line 163: needed due to https://github.com/iasql/iasql-engine/issues/1390

src/modules/0.0.22/aws_s3/index.ts
```diff
@@ -9,6 +9,7 @@ import { policiesAreSame } from '../../../services/aws-diff';
 import { AWS, crudBuilder2, crudBuilderFormat } from '../../../services/aws_macros';
 import { Context, Crud2, MapperBase, ModuleBase } from '../../interfaces';
 import { Bucket } from './entity';
+import { S3CleanBucketRpc } from './rpcs';
 
 class BucketMapper extends MapperBase<Bucket> {
```
dfellis, line 14: Do we want to move the Mapper into a `/mapper` directory with this change since the `/rpcs` directory has also been made?

Or perhaps move the singular `rpc` function in here, too?

dfellis (approved):

aguillenv (approved):

PR #1391, opened 2022-10-12
Multi-region `aws_memorydb`

aguillenv:


src/modules/0.0.22/aws_memory_db/sql/after_install.sql
```diff
@@ -27,3 +27,43 @@ UPDATE
...
+UPDATE
+  ON subnet_group FOR EACH ROW
+EXECUTE
+  FUNCTION check_subnet_group_subnets ();
+
+-- We have a weird case were we need to add 2 FKs to a join table column, and typeorm is not that flexible to define that
+ALTER TABLE
+  "memory_db_cluster_security_groups"
+ADD
+  CONSTRAINT "FK_memory_db_cluster_security_groups_region" FOREIGN KEY ("security_group_id", "region") REFERENCES "security_group" ("id", "region") ON DELETE CASCADE ON UPDATE CASCADE;
```
dfellis, line 69: This is actually not true. We figured out how to do it. We need to add the `@unique` on `id` and `region` in the `SecurityGroup` entity and then we can do the `@JoinColumn([...])` syntax.

aguillenv: I tried that, but it is a join table with join columns and Typeform does not like them having the same name. It renames it like `region_1` and `region_2`. I also tried adding two different columns each one referencing its region but was having trouble with "duplicated" inserts in the join table and couldn't figure out why 🤔

yrobla (approved):

PR #1392, opened 2022-10-12
drop TypeORM from SQL-only modules

depombo:


dfellis (approved):

src/modules/interfaces.ts
```diff
@@ -464,9 +464,14 @@ export class ModuleBase {
...
   init() {
+    this.loadBasics();
+    this.loadTypeORM();
+  }
+
+  loadBasics() {
     if (!this.dirname) {
       this.dirname =
-        path.dirname(callsite()?.[1]?.getFileName?.()) ??
+        path.dirname(callsite()?.[2]?.getFileName?.()) ??
```
aguillenv, line 474: why this change?

aguillenv (approved):

src/modules/interfaces.ts
depombo: this looks at the stack trace and since I added another function call it changed

PR #1393, opened 2022-10-12
be able to wait until codebuild build is done

depombo:


src/modules/0.0.22/aws_codebuild/entity/build.ts
```diff
@@ -73,4 +73,11 @@ export class CodebuildBuildImport {
     name: 'project_name',
   })
   project: CodebuildProject;
+
```
yrobla, line 76: why not always wait? that is the default behaviour for other modules...

aguillenv (approved):

src/modules/0.0.22/aws_codebuild/entity/build.ts
yrobla: in the acm request certificate, i made it wait until the cert was valid. I also did the same for codedeploy, and now with codepipeline...

aguillenv: that's true, rds also takes a lot of time and we wait for it

depombo: will change the default behavior!

dfellis (approved):

aguillenv (approved):

yrobla (approved):

PR #1396, opened 2022-10-12
allow continued management of user-generated tables via group role

depombo:


dfellis (approved):

src/services/iasql.ts
```diff
@@ -1287,6 +1291,12 @@ export async function upgrade(dbId: string, dbUser: string) {
           name: dbId,
           database: dbId,
         });
+        // TODO remove once 0.0.21 is deprecated
+        // If upgrading from a version older v0.0.22 add a group role
+        // that gets added on connect
+        if (['0.0.17', '0.0.18', '0.0.20', '0.0.21'].includes(versionString)) {
```
aguillenv, line 1297: 0.0.21 will not run in this file, maybe we are missing a change in `0.0.21/iaslq_functions/iasql.ts`?

src/services/db-manager.ts
```diff
@@ -81,23 +88,55 @@ export function grantPostgresRoleQuery(user: string) {
...
-    REVOKE INSERT ON ALL TABLES IN SCHEMA public FROM ${user};
-    REVOKE UPDATE ON ALL TABLES IN SCHEMA public FROM ${user};
-    REVOKE DELETE ON ALL TABLES IN SCHEMA public FROM ${user};
-    REVOKE EXECUTE ON ALL FUNCTIONS IN SCHEMA public FROM ${user};
-    REVOKE EXECUTE ON ALL PROCEDURES IN SCHEMA public FROM ${user};
-    REVOKE USAGE, SELECT ON ALL SEQUENCES IN SCHEMA public FROM ${user};
-    REVOKE CREATE ON SCHEMA public FROM ${user};
-    REVOKE CONNECT ON DATABASE ${dbId} FROM ${user};
+export function grantPostgresGroupRoleQuery(user: string, dbId: string, versionString: string) {
+  if (['0.0.17', '0.0.18', '0.0.20', '0.0.21'].includes(versionString)) {
```
yrobla, line 92: this is repeated everywhere, and is also something that looks dynamic... so maybe you can just add it into a constant and reuse?

src/services/iasql.ts
depombo: not quite, connect/disconnect/runsql still happen in the top level iasql service. this gets called on runsql from an updated engine on a old version. I created a new method so I didn't have to touch the 0.0.21 method

depombo: so if upgrading for a version older the v0.0.22 the group role doesn't exist for the database and needs to be created. for new databases this group role gets created on connect

aguillenv: I guess my confusion is that `versionString` will never be 0.0.21 in this file because this `upgrade` will be executed for databases in 0.0.20 or older.

aguillenv (approved):

PR #1399, opened 2022-10-13
Fix `iasql_functions` upgrade

### LINKED ISSUE #1387: Fix upgrade from 0.0.21 to 0.0.22

aguillenv:
Resolves #1387 

The change in version 0.0.21 is needed because it is actually what causes the issue, so that test will always fail.


src/modules/0.0.21/iasql_functions/iasql.ts
```diff
@@ -897,21 +899,17 @@ export async function upgrade(dbId: string, dbUser: string, context: Context) {
     if (!db) return 'Database no found (somehow)';
     await MetadataRepo.dbUpgrading(db, true);
     (async () => {
+      // We need to sleep for a bit to let the scheduler response execute before we start messing up with the db
+      await new Promise(r => setTimeout(r, 5000));
```
depombo, line 903: would it be possible to explicitly wait on something here or no? just mindful of introducing flakiness if the schedule is not ready, but I understand this is the nature of upgrade if there is nothing else

dfellis (approved):
Approved pending all tests but one passing. :)

src/modules/0.0.22/iasql_functions/iasql.ts
```diff
@@ -17,6 +16,9 @@ import MetadataRepo from '../../../services/repositories/metadata';
 import * as scheduler from '../../../services/scheduler-api';
 import { TypeormWrapper } from '../../../services/typeorm';
 
+// tslint:disable-next-line:no-var-requires
+const upgradedIasql = require(`../../${config.modules.latestVersion}/iasql_functions/iasql`);
```
depombo, line 20: maybe a comment as to what this is doing

depombo (approved):

test/upgrade/run-upgrade.sh
```diff
@@ -5,129 +5,140 @@ set -vex
...
+    if [ "${AWSACCOUNTMODULE}" == "aws_account@${LATESTVERSION}" ]; then
+      ISUPGRADED=true
+      UPGRADECHECKCOUNT=0
+    fi
+    sleep 1
+    UPGRADECHECKCOUNT=$((${UPGRADECHECKCOUNT} - 1))
+  done
+
+  # The check that the upgrade successfully loads up the new version for the account
+  if [ "${ISUPGRADED}" == "false" ]; then
```
yrobla, line 91: can we get a list of the modules that failed to upgrade? would be helpful

test/upgrade/run-upgrade.sh
```diff
@@ -5,129 +5,140 @@ set -vex
...
+  AWSACCOUNTTABLE=$(psql postgres://postgres:test@127.0.0.1:5432/to_upgrade -AXqtc "
+      SELECT \"table\" FROM iasql_tables WHERE \"table\" = 'aws_account';
+    ")
+  if [ "${AWSACCOUNTTABLE}" == "aws_account" ]; then # It's the old style
+    psql postgres://postgres:test@127.0.0.1:5432/to_upgrade -c "
+        INSERT INTO aws_account (access_key_id, secret_access_key, region)
+        VALUES ('${AWS_ACCESS_KEY_ID}', '${AWS_SECRET_ACCESS_KEY}', 'us-east-1');
+      "
+  else
+    psql postgres://postgres:test@127.0.0.1:5432/to_upgrade -c "
```
yrobla, line 48: would make sense to store this connection string on a variable, it's repeated everywhere

test/upgrade/run-upgrade.sh
aguillenv: well if one fails it will exit, so we will know exactly which one failed

src/modules/0.0.21/iasql_functions/iasql.ts
aguillenv: I get your point but I do not think there's anything else we can do here more than wait.

yrobla (approved):

PR #1409, opened 2022-10-17
add initial version of s3 multiregion

yrobla:


src/modules/0.0.22/aws_s3/mappers/bucket.ts
```diff
@@ -38,74 +75,132 @@ export class BucketMapper extends MapperBase<Bucket> {
...
   }
   updateBucketPolicy = crudBuilder2<S3, 'putBucketPolicy'>('putBucketPolicy', input => input);
+
+  async createBucketPolicy(client: S3, bucket: Bucket, ctx: Context) {
+    const input: PutBucketPolicyCommandInput = {
+      Bucket: bucket.name,
+      Policy: JSON.stringify(bucket.policyDocument ?? {}),
+    };
+    await this.updateBucketPolicy(client, input);
+    await this.module.bucket.db.update(bucket, ctx);
```
dfellis, line 85: Nit: Do we need to do the `db.update` twice?

src/modules/0.0.22/aws_s3/mappers/bucket.ts
```diff
@@ -38,74 +75,132 @@ export class BucketMapper extends MapperBase<Bucket> {
...
   cloud = new Crud2<Bucket>({
     // TODO: There are lots of useful permission controls on create to be added to this model
     create: async (es: Bucket[], ctx: Context) => {
-      const client = (await ctx.getAwsClient()) as AWS;
       const out = [];
+      const enabledRegions = (await ctx.getEnabledAwsRegions()) as string[];
+      let client;
+
       for (const e of es) {
+        // if the bucket name exists into another region we cannot create it, throw error
```
dfellis, line 110:
```suggestion
        // if the bucket name exists in another region we cannot create it, throw error
```

src/modules/0.0.22/aws_s3/rpcs/s3_clean_bucket.ts
```diff
@@ -38,29 +45,58 @@ export class S3CleanBucketRpc extends RpcBase {
...
-      return [
-        {
-          bucket: bucketName,
-          status: 'OK',
-          response_message: 'All bucket objects have been deleted',
-        },
-      ];
+    // first determine bucket region
+    let region = await this.getBucketLocation(client.s3Client, bucketName);
+    if (!region) region = 'us-east-1';
```
dfellis, line 52: Can make this const by making it a one-liner.

```suggestion
    const region = (await this.getBucketLocation(client.s3Client, bucketName)) ?? 'us-east-1';
```

src/modules/0.0.22/aws_s3/rpcs/s3_clean_bucket.ts
```diff
@@ -38,29 +45,58 @@ export class S3CleanBucketRpc extends RpcBase {
...
+    const enabledRegions = (await ctx.getEnabledAwsRegions()) as string[];
+
+    if (region) {
+      // check if it is on enabled regions
+      if (enabledRegions.includes(region)) {
+        const client1 = (await ctx.getAwsClient(region)) as AWS;
```
dfellis, line 58: Why is this called `client1`? I don't see a `client2`?

dfellis (approved):
All minor comments.

src/modules/0.0.22/aws_s3/rpcs/s3_clean_bucket.ts
yrobla: There was a client, using the default zone, and then i create another client, specific for the bucket region, to clean the bucket

src/modules/0.0.22/aws_s3/mappers/bucket.ts
```diff
@@ -38,74 +75,131 @@ export class BucketMapper extends MapperBase<Bucket> {
...
-      for (const index in buckets) {
-        if (buckets.hasOwnProperty(index)) {
-          const bucket = buckets[index];
+      if (!!id) {
+        const { bucketName, region } = this.idFields(id);
+        if (enabledRegions.includes(region)) {
+          const allBuckets = await this.getBuckets(client.s3Client);
+          rawBuckets = allBuckets.filter(b => !bucketName || b.Name === bucketName).filter(b => !!b.Name);
+        }
+      } else {
```
aguillenv, line 140: if this method is being called with an `id`, we have to return just one element and not a list of them, so I think we should `.find` the `bucketName` received from the id, get its policy, map it and return it.

src/modules/0.0.22/aws_s3/mappers/bucket.ts
```diff
@@ -38,74 +75,131 @@ export class BucketMapper extends MapperBase<Bucket> {
...
-        if (isUpdate) {
-          e.createdAt = cloudRecord.createdAt;
-          e.policyDocument = await this.module.bucket.createBucketPolicy(client.s3Client, e, ctx);
-          out.push(e);
-        } else {
-          // Replace if name has changed
-          await this.module.bucket.db.update(cloudRecord, ctx);
-          ctx.memo.db.Bucket[cloudRecord.name] = cloudRecord;
-          out.push(cloudRecord);
+        if (enabledRegions.includes(e.region)) {
```
aguillenv, line 182: oh we have not been doing this but I think it makes sense 🤔

src/modules/0.0.22/aws_s3/mappers/bucket.ts
```diff
@@ -38,74 +75,131 @@ export class BucketMapper extends MapperBase<Bucket> {
...
+          const cloudRecord = ctx?.memo?.cloud?.Bucket?.[this.entityId(e)];
+          const isUpdate = Object.is(this.module.bucket.cloud.updateOrReplace(cloudRecord, e), 'update');
+          if (isUpdate) {
+            e.createdAt = cloudRecord.createdAt;
+            e.policyDocument = await this.module.bucket.createBucketPolicy(client.s3Client, e, ctx);
+            out.push(e);
+          } else {
+            // we cannot modify bucket name or region of the bucket, replace it
+            await this.module.bucket.db.update(cloudRecord, ctx);
+            ctx.memo.db.Bucket[cloudRecord.name] = cloudRecord;
```
aguillenv, line 193: We should always treat the memo with the same entityId! Maybe this should be something like
```suggestion
            ctx.memo.db.Bucket[this.entityId(cloudRecord)] = cloudRecord;
```

src/modules/0.0.22/aws_s3/mappers/bucket.ts
```diff
@@ -135,7 +129,8 @@ export class BucketMapper extends MapperBase<Bucket> {
         const { bucketName, region } = this.idFields(id);
```
aguillenv, line 129: this won't be working since the entity property with the cloud id is `name` and not `bucketName`

src/modules/0.0.22/aws_s3/mappers/bucket.ts
```diff
@@ -135,7 +129,8 @@ export class BucketMapper extends MapperBase<Bucket> {
...
         if (enabledRegions.includes(region)) {
           const allBuckets = await this.getBuckets(client.s3Client);
-          rawBuckets = allBuckets.filter(b => !bucketName || b.Name === bucketName).filter(b => !!b.Name);
+          const foundBucket = allBuckets.find((b: BucketAWS) => b.Name === bucketName);
+          if (foundBucket) rawBuckets.push(foundBucket);
```
aguillenv, line 133: this is not solving my previous comment. This way you would still be returning a list of buckets instead of a single bucket when the `cloud.read` method is called asking just for one `id`, it is not consistent.

aguillenv (approved):

PR #1411, opened 2022-10-17
Multi-region EC2 metadata

dfellis:


mtp1376 (approved):

src/modules/0.0.22/aws_ec2_metadata/index.ts
```diff
@@ -43,6 +43,7 @@ class InstanceMetadataMapper extends MapperBase<InstanceMetadata> {
     out.ebsOptimized = instance.EbsOptimized ?? false;
     out.rootDeviceName = instance.RootDeviceName ?? '';
     out.rootDeviceType = (instance.RootDeviceType as RootDeviceType) ?? RootDeviceType.EBS;
+    out.region = region;
     return out;
```
aguillenv, line 47: we need to change here the instance read call to use the instanceId + the region

dfellis: Oh, good catch.

aguillenv (approved):

PR #1421, opened 2022-10-19
Merge the ACM modules into one 'aws_acm' module

### LINKED ISSUE #1295: Merge ACM modules into one
Requires https://github.com/iasql/iasql-engine/issues/1294 so the existing modules can be deprecated and a new one called `aws_acm` can be introduced

dfellis:
Resolves #1295


test/modules/aws-acm-import-integration.ts
```diff
@@ -148,7 +102,7 @@ describe('AwsAcmImport Integration Testing', () => {
     .then(...finish(done)));
 });
 
-describe('AwsAcmImport install/uninstall', () => {
+describe('AwsAcm Import install/uninstall', () => {
```
aguillenv, line 105: this last describe would not be necessary since it is not a separate module anymore

test/modules/aws-acm-request-integration.ts
```diff
@@ -229,7 +164,7 @@ describe('AwsAcmRequest Integration Testing', () => {
   it('deletes the test db', done => void iasql.disconnect(dbAlias, 'not-needed').then(...finish(done)));
 });
 
-describe('AwsAcmRequest install/uninstall', () => {
+describe('AwsAcm Request install/uninstall', () => {
```
aguillenv, line 167: Same here, would not be necessary since request is not a module anymore.

test/modules/aws-acm-request-integration.ts
```diff
@@ -53,37 +53,30 @@ describe('AwsAcmRequest Integration Testing', () => {
...
-
-  it(
-    'adds a new certificate to request with incorrect domain',
+    'adds a new certificate to request with a fake domain',
     query(`
-    INSERT INTO certificate_request (domain_name)
-    VALUES ('fakeDomain.com');
-  `),
+      SELECT * FROM certificate_request('fakeDomain.com', 'DNS', '${process.env.AWS_REGION}', '');
+    `, console.log),
```
aguillenv, line 78: why this `console.log`? Also, isn't this test the same as the one immediately above?

dfellis: Oh, I am surprised the linter didn't catch that. I will remove the debug logging there.

It is repeated, but to test this path failing in two different ways. This one fails "hard" because Route53 is not installed, then we install it and it fails "soft" because AWS cannot validate the fake domain.

aguillenv (approved):

yrobla (commented):
some questions

src/modules/0.0.22/aws_acm/rpcs/common.ts
```diff
@@ -0,0 +1,7 @@
+export const safeParse = (json: string) => {
```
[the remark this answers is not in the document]
dfellis: I have considered popping this out to the `/src/services/` directory, but I didn't because I wasn't sure how widespread it would be in the codebase.

src/modules/0.0.22/aws_acm/rpcs/request.ts
```diff
@@ -10,18 +10,27 @@ import {
...
+import { Context, RpcBase, RpcResponseObject } from '../../../interfaces';
+import { awsRoute53HostedZoneModule } from '../../aws_route53_hosted_zones';
+import { HostedZone, ResourceRecordSet } from '../../aws_route53_hosted_zones/entity';
+import { modules } from '../../iasql_functions/iasql';
+import { Certificate } from '../entity';
+import { safeParse } from './common';
+
+enum ValidationMethod {
+  DNS = 'DNS',
+  EMAIL = 'EMAIL',
```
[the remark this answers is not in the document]
dfellis: Hmm. I didn't want to make too many changes with this RPC-ification of the ACM request module. Can we do this in a follow-up PR?

PR #1417, opened 2022-10-19
fix: include sample buildspec.yml to consume from codepipeline

yrobla:
Signed-off-by: Yolanda Robla <info@ysoft.biz>

PR #1422, opened 2022-10-20
Revert "fix: include sample buildspec.yml to consume from codepipeline"

yrobla:
Reverts iasql/iasql-engine#1417


aguillenv (approved):
it was not necessary?

PR #1425, opened 2022-10-20
Handle tables and functions provisioned by sql-only modules

### LINKED ISSUE #1424: `aws_ecs_simplified` does not appear in `iasql_tables` after installation
the module is being installed but does not appear when you check the `iasql_tables` or if you check the UI modules or the schema. The problem is that interfaces.ts `loadBasics` is not loading properly the provides tables and fns properties and then is not being inserted in `iasql_tables`

aguillenv:
The problem: SQL-only modules like `aws_ecs_simplified` does not correctly handle the tables and functions that should be  provisioned. As a consequence, after module installation, the module was not listed in `iasql_tables`.

The result after installing the module successfully in main branch:
image

How the schema looks like after installation in main branch:
image

The result after installing the module successfully in this branch:
image

How the schema looks like after installation in this branch:
image

Resolves #1424

### DIFF
--- a/src/modules/interfaces.ts
+++ b/src/modules/interfaces.ts
@@ -499,28 +499,6 @@ export class ModuleBase {
       tables: [],
       functions: [],
     };
-    // Should we also be doing this for `beforeInstallSql`?
-    try {
-      const afterInstallSqlParsed = parse(afterInstallSql);
-      const tableStmts = afterInstallSqlParsed.filter((s: any) =>
-        Object.keys(s.RawStmt?.stmt ?? {}).includes('CreateStmt'),
-      );
-      this.provides.tables.push(
-        ...tableStmts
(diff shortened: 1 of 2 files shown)


dfellis (approved):
Wow, far more generic than I expected.

This doesn't accidentally double-count "normal" modules, though, does it?

aguillenv:
> Wow, far more generic than I expected.
> 
> This doesn't accidentally double-count "normal" modules, though, does it?

Since the `init` method was split into two methods `loadBasics` and `loadTypeORM` they should not. `loadBasics` creates the object with empty arrays and pushes the necessary tables and functions parsed from `afterInstallSql`, then `loadTypeORM` gets the migration files, parsing them and pushes the tables and functions found.

depombo:
Oh so this bug was introduced when I removed TypeORM from `ecs_simplified` last week. I didn't notice. Perhaps we can add a quick unit test to avoid regressions?

PR #1432, opened 2022-10-22
Continuous Two-Way Synchronization RFC

dfellis:


rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+
+To emulate manual mode would basically be to temporarily turn off that cron, let the user make whatever changes they want to batch together, and then re-enable that cron. Because we can't actually support the current `iasql_apply` for the three `code-` services, and because entering this batching mode that could potentially be unwound, it looks a *lot* like a transaction, so the proposed new syntax would be to emulate that:
+
+```sql
+CALL iasql_begin(); -- Disables the cron. Wrap a function call inside of a procedure so we can avoid the confusing `SELECT iasql_begin();` syntax here.
```
aguillenv, line 54: what could happen if we call `iasql_begin()` while `iasql_commit()` is being executed in another thread?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+
+SELECT iasql_drop_track_all(); -- Restores normal behavior
+```
+
+While entering the transaction mode introduces a synchronous point (such that `CALL iasql_commit();` blocks until the apply/sync operation is complete), there may be times where a user made changes in "normal" two-way mode but they now need to wait for resources to actually be created. A proposed helper function would trigger a blocking call until a user condition is met. One possible syntax for this is:
+
+```sql
+SELECT iasql_wait_for('table_name', 5, 'cloud_id IS NOT NULL'); -- Default version
+SELECT iasql_wait_for('table_name', 5, 'cloud_id IS NOT NULL', 30); -- Optional argument for maximum # of seconds to wait before erroring (defaulting to 5 minutes or something)
+```
```
aguillenv, line 104: hmm, but isn't the `iasql_commit` call blocking? I'd imagine it like this: call commit, execute all the changes, enable the cron, and exit.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+SELECT iasql_track('table_name', 5); -- Makes a particular row read-only.
+
+SELECT iasql_track_all(); -- Makes the DB read-only
+
+SELECT iasql_drop_track('table_name'); -- Drops tracking for the table. Similar issue to above: should this affect row-level tracking in the same table?
+
+SELECT iasql_drop_track('table_name', 5); -- Drops tracking for a particular row.
+
+SELECT iasql_drop_track_all(); -- Restores normal behavior
+```
```
aguillenv, line 97: Wow, this would really give some flexibility, but also, it is quite tricky for us? Could a user mess up using these statements and create locks or infinite loops because we need to update a table that is set as write-only?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+```sql
+SELECT iasql_enforce('table_name'); -- Enforce an entire table onto the cloud.
+
+SELECT iasql_enforce('table_name', 5); -- Enforce a particular row from a table onto the cloud. We need to use the database ID here because enforcement *will* cause `@cloudId` churn if it needs to recreate the resource.
+
+SELECT iasql_enforce_all(); -- Essentially makes the DB write-only
+
+SELECT iasql_drop_enforce('table_name'); -- Drop the enforcement of the table. Should it also automatically drop specific row-level enforcement is something to decide. I'm leaning towards "yes" but it could be unexpected.
+
+SELECT iasql_drop_enforce('table_name', 5); -- Drop the enforcement of a particular row. This one, at least, is unambiguous.
```
aguillenv, line 82: we will need to have the version accepting an array of ids too.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
aguillenv: ohh forget it, I understand now that the `isql_wait_for` is for the cases when no transaction like was executed, but normal 2way mode

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+- The entire test suite
+- Most of the documentation
+
+## Expected Timeline
+
+This will take a few weeks to get across the line. The major steps (with estimated times) are:
+
+1. Implement the `iasql_commit` function and port test suite to using it. (1 week)
+2. Implement pausable cron and transaction functionality, make `apply/sync` no-ops, and update the test suite again. (1 week)
+3. Implement Enforce/Track/WaitFor and add new tests for them (3-4 days)
```
aguillenv, line 159: I think the Enforce/Track section will be more difficult to implement or the one that could give us more problems.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+## Affected Components
+
+- The `iasql_functions` module
+- The core RPC scheduler
+- The entire test suite
+- Most of the documentation
+
+## Expected Timeline
+
+This will take a few weeks to get across the line. The major steps (with estimated times) are:
```
aguillenv, line 155: I'm missing here (and in the RFC) the audit logs. I think they are the foundational part for this RFC to work properly and maybe need a paragraph on how they are expected to work or look and its own timeline in this section. It will not only be the table creation but also add the triggers for all tables.

aguillenv (commented):
I really like the transaction-like look!

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: Oh, I forgot to mention that. We need to add some extra metadata to the metadata DB and/or a new table to add a locking mechanism so they don't run in parallel, both begin and commit obey this locking mechanism, where the actual cron type usage is essentially calling the two one after the other. Begin blocks on lock acquisition and commit won't run if it isn't locked and immediately unlocks before it starts, I think will work.

We may need two levels of locking, though. I need to work through the timing diagram, probably.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: Yep :)

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+
+To accomplish this, let's assume under normal operations, there is a singular apply/sync-combined command to perform the "Two-Way"-style action. It works similarly to the current sync and apply, but also uses the audit log to decide for each record which behavior to follow. This piece can be operated in a "manual" style, invoked explicitly, but we'll instead invoke it through a cron-like interval, every few minutes, to make the automatic mode. In this way, there is no special syntax for users to work with normally. They just `SELECT/INSERT/UPDATE/DELETE` records they see and eventually this is reflected in their cloud account.
+
+To emulate manual mode would basically be to temporarily turn off that cron, let the user make whatever changes they want to batch together, and then re-enable that cron. Because we can't actually support the current `iasql_apply` for the three `code-` services, and because entering this batching mode that could potentially be unwound, it looks a *lot* like a transaction, so the proposed new syntax would be to emulate that:
+
+```sql
```
yrobla, line 53: is there a way to make it atomic? or , if we detect that the transaction has stopped in the middle for x reason , rollback to a safe state ?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: I figure write-only would work like the current apply and still allow mutation to include cloud generated IDs and the like during the loop?

But there is trickiness here due to mutual exclusion and what precisely should be done with the nesting nature of all/table/row rules.

Probably should expand this section and lay out the possibilities.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: So we already have audit logs implemented. We simply need to use them. I think it should simply be a collection of select queries that we use, potentially joining in the user group/role onto the table, but I not sure that is necessary.

I can expand this with example queries that I am thinking of here.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: That may be the case. I just figured the test suite changes for the first transition would take more "grunt work" effort while this would be potentially more complicated mentally but more contained?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+-- If we don't like what we saw
+
+CALL iasql_rollback(); -- Executes `iasql_sync`, then re-enables the cron
+
+-- If we like what we saw
+
+CALL iasql_commit(); -- Re-enables the cron, but also performs a blocking execution of the new logic so a `SELECT` on the newly-created records has the desired cloud IDs
+```
+
+To emulate the two One-Way automatic modes, we can take a page from Postgres' [Table and Row-Level Locking](https://www.postgresql.org/docs/15/explicit-locking.html). We can create sets of functions to put a whole table or just specific rows into either of these modes or put them back to normal operations. These functions would write to a new metadata table that would override the decision the audit log makes, which makes them pretty easy additions to the functionality while gaining significant power and flexibility for users (and could be pure SQL stored procedures that don't go through the engine's RPC path at all, since they just manipulate a config table).
```
yrobla, line 69: is audit log secure enough to be used as source of truth? are there different options for audit logs, that we should investigate? can it be the case that we take more time to process the audit log, than the period of cronjob? how are we going to tackle that?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
yrobla: yay.... how to avoid deadlocks there, and avoid being in a broken state if there is some problem with engine, network, etc?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
depombo: I think it should be possible via a lock (it can simply be a column in the metadata table akin to what we do for updates) that the cron checks

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+
+SELECT * FROM iasql_preview(); -- Because there's no distinction between apply and sync, this function name is simpler than before
+
+-- If we don't like what we saw
+
+CALL iasql_rollback(); -- Executes `iasql_sync`, then re-enables the cron
+
+-- If we like what we saw
+
+CALL iasql_commit(); -- Re-enables the cron, but also performs a blocking execution of the new logic so a `SELECT` on the newly-created records has the desired cloud IDs
```
depombo, line 66: We should note that after an `iasql_rollback`, the `iasql_commit` will fail (or be a no-op) unless we call  `iasql_begin` since there is no current transaction

rfcs/004 - Continuous Two-Way Synchronization RFC.md
depombo: David mentions it here https://github.com/iasql/iasql-engine/pull/1432/files#r1003607851

rfcs/004 - Continuous Two-Way Synchronization RFC.md
depombo: Yes! We should also have a `iasql_wait_for_all()`? Probably from the get-go?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+
+```sql
+SELECT count(*) > 0 FROM `table_name` WHERE id = `5` AND `cloud_id IS NOT NULL`
+```
+
+The parts in backticks are from the user-provided values. This is SQL injection, but it's their own database and we already allow arbitrary SQL statements to run on the database, so it doesn't affect the security situation in the slightest, even if it does immediately cause heart palpitations while looking at it. ;)
+
+The final recommendation is to keep `iasql_apply` and `iasql_sync` as functions that can be called, but make them no-ops that warn that they don't do anything anymore, and that at some time in the future these functions will be dropped. The reasoning behind this is that this transition will take some time and there may be IaSQL snippets out there that do continue working after this transition assuming they didn't call either of those functions. We can track the number of actual invocations in these functions by having the engine-side implementation simply fire off an alert to us. If we see little-to-no triggering by end users, we could drop them sooner rather than later.
```
depombo, line 114: we do need to drop `iasql_apply_preview` and `iasql_sync_preview` ahead of the v0.0.1 release since they won't be backwards compatible, right?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,160 @@
...
+
+However, it is possible for "Two-Way" mode to emulate the other three in a way that is also mostly-backwards-compatible with the current manual mode. Furthermore, a singular IaSQL database would be able to have different "parts" of the database emulating different modes at the same time.
+
+To accomplish this, let's assume under normal operations, there is a singular apply/sync-combined command to perform the "Two-Way"-style action. It works similarly to the current sync and apply, but also uses the audit log to decide for each record which behavior to follow. This piece can be operated in a "manual" style, invoked explicitly, but we'll instead invoke it through a cron-like interval, every few minutes, to make the automatic mode. In this way, there is no special syntax for users to work with normally. They just `SELECT/INSERT/UPDATE/DELETE` records they see and eventually this is reflected in their cloud account.
```
depombo, line 49: I think there are no implications to migration usage and in fact it simplifies the UX for it, but it might be worth a mention

rfcs/004 - Continuous Two-Way Synchronization RFC.md
depombo: I would split this off into three different steps:
3. Implement WaitFor and add tests for them
4. Implement Enforce/Track per db
5. Implement granular Enforce/Track per table

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: What would `wait_for_all` do?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: So I am pretty sure we locked down the audit log to only be mutable by the trigger functions we attach to every mapper-driven-table, so as a source of truth for mutations in the database it should work.

Distinguishing between "user-intended change" and "change by the engine" is important, though. We already attach the postgres username to each row the in the changelog, so there should be enough info there, though we may need to join that on the postgres roles to be sure we filter out rows we don't want for this comparison.

As for cron job overlap, I will be more explicit about the need for a lock in the metadata database that the engine will use. If the cron run from a prior iteration overlaps the next one, the next one should see a lock failure and simply not run, allowing the iteration after that to try again.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: If we release v0.1 without the two-way mode RFC work implemented, we *probably* still want them around for tutorials to use them, right?

Then it would make sense to have these functions just `raise exception` that they no longer work and that they should read the docs on IaSQL transactions.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
depombo: I mean wait for all apply/sync transactions to complete

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: Addressed.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: I haven't addressed the broken state issue, yet, will get to that next, but I consider these two "one-way modes" more power-user-type stuff, so the rough answer is "it's on the user who broke it" but I'll try to be more specific.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: Dropped this entire approach for a better, more SQL-y one :)

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -66,52 +66,53 @@ CALL iasql_rollback(); -- Executes `iasql_sync`, then re-enables the cron
...
-
-SELECT iasql_enforce_all(); -- Essentially makes the DB write-only
+While entering the transaction mode introduces a synchronous point (such that `CALL iasql_commit();` blocks until the apply/sync operation is complete), there may be times where a user made changes in non-transactional two-way mode but they now need to wait for resources to actually be created. A proposed helper function would trigger a blocking call until a user condition is met. One possible syntax for this is:
 
-SELECT iasql_drop_enforce('table_name'); -- Drop the enforcement of the table. Should it also automatically drop specific row-level enforcement is something to decide. I'm leaning towards "yes" but it could be unexpected.
+```sql
+SELECT iasql_wait_for('table_name', 5, 'cloud_id IS NOT NULL'); -- Default version
 
-SELECT iasql_drop_enforce('table_name', 5); -- Drop the enforcement of a particular row. This one, at least, is unambiguous.
+SELECT iasql_wait_for('table_name', 5, 'cloud_id IS NOT NULL', 30); -- Optional argument for maximum # of seconds to wait before erroring (defaulting to 5 minutes or something)
```
depombo, line 78: what does the `5` mean? 

We should have this be a follow-up if/when users ask for it? `iasql_wait_for_all` covers most of the use cases and this is probably more of a nice-to-have

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,163 @@
...
+
+These checks should also pay attention to the timestamp and ignore the missing `end` record condition if it has been "too long" and therefore the engine was likely restarted in the meantime. The exact time will depend on what limit we ourselves put on the new commit logic, but perhaps 30 minutes?
+
+While entering the transaction mode introduces a synchronous point (such that `CALL iasql_commit();` blocks until the apply/sync operation is complete), there may be times where a user made changes in non-transactional two-way mode but they now need to wait for resources to actually be created. A proposed helper function would trigger a blocking call until a user condition is met. One possible syntax for this is:
+
+```sql
+SELECT iasql_wait_for('table_name', 5, 'cloud_id IS NOT NULL'); -- Default version
+
+SELECT iasql_wait_for('table_name', 5, 'cloud_id IS NOT NULL', 30); -- Optional argument for maximum # of seconds to wait before erroring (defaulting to 5 minutes or something)
```
depombo, line 78: what does the `5` mean? 

We should have this be a follow-up if/when users ask for it? `iasql_wait_for_all` is enough to unblock the `code-*` modules and covers most of the use cases such that this is probably more of a nice-to-have?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,163 @@
...
++-----------------------------------------+
+| iasql_mode                              |
++-----------------------------------------+
+| table_name NULLABLE VARCHAR             |
+| record_id  NULLABLE INT                 |
+| mode       ENUM(NORMAL, ENFORCE, TRACK) |
++-----------------------------------------+
+| UNIQUE (table_name, record_id)          |
+| FK (table_name => iasql_tables.table)   |
++-----------------------------------------+
```
depombo, line 110: can we no longer change the behavior of the cron / commit with a `iasql_enforce_all` and `iasql_track_all` functions?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### Implement `iasql_commit` as replacement for `iasql_apply`/`iasql_sync` only
+
+Stick to manual mode and just replace `apply` and `sync` with `iasql_commit`. This allows us to implement the `code-` modules and requires less work, but keeps the harder to explain syntax in the forefront, rather than being a power user tool. It also precludes our ability to support Enforce/Track behavior described above, which narrows our set of advantages.
+
+This may still be a point along-the-way in the conversion to the Full Two-Way Mode, though, but has been rejected as the stopping point.
+
+#### Implement Two-Way Mode with the transaction-like syntax only
+
+Basically just the first half of the proposal, stopping before the Enforce/Track section. This is probably also a point along-the-way, but has been rejected as the stopping point because the remaining features are simpler to implement and provide significant functionality.
```
depombo, line 141: I think this is outdated to reflect the new implementation proposal via the start/end timestamp check and we should probably call the old version out too?

yrobla (commented):
considering cron vs events approach.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: Sure we can, just do:

```sql
INSERT INTO iasql_mode (mode) VALUES ('ENFORCE');
```

If you have `NULL` for the `table_name` and the `record_id` it would be interpreted as a wildcard "select all" in that category type of thing. I mention this in the paragraph below the schema table, but let me know if it's confusing and why?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: I will double-check that these are referring to the same sections as they used to, but I don't follow "the old version" here?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,163 @@
...
+-- If we don't like what we saw
+
+CALL iasql_rollback(); -- Executes `iasql_sync`, then re-enables the cron
+
+-- If we like what we saw
+
+CALL iasql_commit(); -- Re-enables the cron, but also performs a blocking execution of the new logic so a `SELECT` on the newly-created records has the desired cloud IDs
+```
+
+Backing this would be a minor addition to the existing audit log. If we have the new cron logic and the rollback/commit calls add rows to the audit log when it starts and ends, they can check if the most recent of these records is a `start` one, implying that the engine is currently working on applying and synchronizing changes with the cloud, and would alter their behavior slightly. The `cron` would simply early exit in this case, as it implies there is an existing cron that has gone overtime, and so it should wait until the next execution of the cron to try again. `iasql_begin` would block until a paired `end` record exists and then continue with its work. `iasql_rollback` and `iasql_commit` should therefore never see this state after `iasql_begin` is called, but for defensiveness they should as well block until an `end` record exists and then do their own work.
```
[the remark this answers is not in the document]
dfellis: I think that approach simplifies in some ways and makes more complicated in others.

It seems to eliminate the need for a locking mechanism in the "normal" usage of Two-Way mode, but we would still need some sort of locking mechanism if the user calls `iasql_begin()`, so that basically shifts the lock up from the database and into the engine. That *can* work at the moment when we have one process doing the work, but we would have to re-centralize the lock once we go multi-process.

Meanwhile, the cron job could actually be powered by `pg_cron` and run on the database itself, and we could have it simply check the state of the audit log to decide if it should run or not and not involve the engine at all in this process, and `iasql_begin()`'s implementation in that scenario could be pure SQL as well, simply inserting a `start_commit` into the audit log once the most recent `start`/`end` pair has completed (then the same logic used to deal with long-running cron jobs would also block the cron job when the user wants the activity to be transactional).

Finally, we can take a look at how these two deal with various failure modes:


```
+------------------+-----------------------------------+-----------------------------------+
| Failure Mode     | Cron                              | Event                             |
+------------------+-----------------------------------+-----------------------------------+
| Stuck job >45min | The graphile worker library kills | The graphile worker library kills |
|                  | the job after 45 minutes and then | the job after 45 minutes and then |
|                  | the cron starts work again, which | no event for the next run is      |
|                  | may resolve correctly this time   | triggered, which causes this DB   |
|                  | or it may be stuck again, but any | to be soft-locked until there is  |
|                  | new changes by the user to the DB | an engine redeploy.               |
|                  | will be considered.               |                                   |
+------------------+-----------------------------------+-----------------------------------+
| Engine dies      | The cron keeps trying and failing | Nothing happens at all.           |
|                  | to call the engine. The cron SQL  |                                   |
|                  | function inserts start and end    |                                   |
|                  | audit records immediately after   |                                   |
|                  | each other, and does nothing.     |                                   |
+------------------+-----------------------------------+-----------------------------------+
| Database dies    | Nothing happens at all.           | The inner cron of the engine will |
|                  |                                   | keep trying to communicate to the |
|                  |                                   | DB and fail. This may or may not  |
|                  |                                   | be handled correctly by our code. |
+------------------+-----------------------------------+-----------------------------------+
| Engine deployed  | The cron creates the graphile job | The two engines both have their   |
|                  | and one of the two parallel       | own inner cron unaware of each    |
|                  | versions of the engine takes the  | other and very likely they will   |
|                  | work (as this approach requires a | both try to execute at the same   |
|                  | SQL RPC call to start the engine) | time for 7-10 minutes. What they  |
|                  | and the other doesn't work on it. | will do is undefined unless we    |
|                  | If it is the old version that     | thoroughly test it, but likely    |
|                  | takes it, it may fail and no work | they will "fight" with each other |
|                  | will happen for 45 minutes, then  | undoing each other's work and     |
|                  | it will continue.                 | potentially corrupting the DB.    |
+------------------+-----------------------------------+-----------------------------------+
```

Basically, I think the event approach has the harder-to-handle failure scenarios, though there are definitely problems that come up with both.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,163 @@
...
+
+```sql
+SELECT iasql_wait_for('table_name', 5, 'cloud_id IS NOT NULL'); -- Default version
+
+SELECT iasql_wait_for('table_name', 5, 'cloud_id IS NOT NULL', 30); -- Optional argument for maximum # of seconds to wait before erroring (defaulting to 5 minutes or something)
+
+SELECT iasql_wait_for_all(); -- This uses special querying of the audit log to make sure all changes up to the time this function is invoked are represented in the cloud
+```
+
+Similar to the enforce and track functions, it is essentially executing (on another postgres thread/process) a particular select statement in a loop with a sleep thrown in until success or it errors out if the condition is never met. That select statement would be something like:
```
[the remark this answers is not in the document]
dfellis: Will fix.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -66,52 +66,53 @@ CALL iasql_rollback(); -- Executes `iasql_sync`, then re-enables the cron
...
 
-Naming these functions to be clearly understood will be important so "Anti-Chaos-Monkey" can't be what we go with here. ;) The current proposal for the name for the "Write-Only Automatic" style functions is "Enforce" as the user is asking IaSQL to enforce these particular rows or tables to always match the database, and the "Read-Only Automatic" style becomes "Track" as the user is asking IaSQL to track the upstream cloud. The proposed new syntax for these is:
+These checks should also pay attention to the timestamp and ignore the missing `end` record condition if it has been "too long" and therefore the engine was likely restarted in the meantime. The exact time will depend on what limit we ourselves put on the new commit logic, but perhaps 30 minutes?
```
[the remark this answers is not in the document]
dfellis: I matched it to the 45 minute level @depombo mentioned.

For "normal" two-way mode, I don't think configuring it would matter much or at all. If it does not completely modify the cloud in a single step, the next iteration of the cron job would continue where it left off after comparing the desired and current states of the cloud.

I suppose it could still be desirable in the transaction-like syntax, but a database transaction taking over 45 minutes seems like it would break a lot of CI systems that have idle timeouts?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: I added a bit more explanatory paragraphs here. I hope that helps?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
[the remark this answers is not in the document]
dfellis: I turned this explanation into a section of the RFC.

yrobla (approved):

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,209 @@
...
+SELECT iasql_wait_for_all(); -- This uses special querying of the audit log to make sure all changes up to the time this function is invoked are represented in the cloud
+```
+
+The `iasql_wait_for` functions essentially execute (on another postgres thread/process) a particular select statement in a loop with a sleep thrown in until success or it errors out if the condition is never met. That select statement generated would be something like:
+
+```sql
+SELECT count(*) > 0 FROM `table_name` WHERE id = `5` AND `cloud_id IS NOT NULL`
+```
+
+The `iasql_wait_for_all`
```
aguillenv, line 110: an orphan text here?

aguillenv (approved):

rfcs/004 - Continuous Two-Way Synchronization RFC.md
dfellis: oh, weird. I had written more than that. Let me see if I can recover it.

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,209 @@
...
+
+-- Do some DB changes
+
+SELECT * FROM iasql_preview(); -- Because there's no distinction between apply and sync, this function name is simpler than before
+
+-- If we don't like what we saw
+
+CALL iasql_rollback(); -- Executes `iasql_sync`, then re-enables the cron
+
+-- If we like what we saw
```
depombo, line 68: I think it is obvious, but still worth calling out that you can either `rollback` *or* `commit`?

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,209 @@
...
+
+Most of the failure modes when using the cron approach lead to delays in eventual consistency between the database and the cloud, while the event-based approach can lead to more catastrophic failures.
+
+Backing the cron approach would be a minor addition to the existing audit log to go with the graphile worker job creation. If we have the new cron logic and the rollback/commit calls add rows to the audit log when it starts and ends, they can check if the most recent of these records is a `start` one, implying that the engine is currently working on applying and synchronizing changes with the cloud, and would alter their behavior slightly. The `cron` would simply early exit in this case, as it implies there is an existing cron that has gone overtime, and so it should wait until the next execution of the cron to try again. `iasql_begin` would block until a paired `end` record exists and then continue with its work. `iasql_rollback` and `iasql_commit` should therefore never see this state after `iasql_begin` is called, but for defensiveness they should as well block until an `end` record exists and then do their own work.
```
depombo, line 84: I need to think harder about this / better understand how `pg_cron` works, but to be able to represent autonomous transactions (https://github.com/iasql/iasql-engine/blob/main/CONTRIBUTING.md#how-iasql-postgresql-functions-work) the cron would need to support dblink or I guess that code lives in SQL anyway and there is no additional work here?

dfellis: So there would be a minor difference between the `iasql_commit()` function and what the cron calls, but they would be almost identical and I think it could simply be a "pre-amble" in the pg_cron cron function to just *not* call `iasql_commit()` if there's an `start_commit` without a paired `end_commit` within the last 45 minutes.

That starts to get into deeper implementation details, but I believe it will only be a couple of extra lines of pure SQL code (beyond the overhead of the wrapper function to pass to `pg_cron`).

depombo (approved):

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Current Status
+
+### Proposed
+
+2022-10-22
+
+### Accepted
+
+YYYY-MM-DD
```
dfellis, line 11:
```suggestion
2022-10-28
```

rfcs/004 - Continuous Two-Way Synchronization RFC.md
```diff
@@ -0,0 +1,209 @@
...
+
+Attempting to `iasql_sync` in the middle of that `iasql_apply` could cause IaSQL to drop other operations it is supposed to be performing, freezing their cloud into some unexpected intermediate state. And it is impossible to "parse" the YAML to determine what resources are going to be created. Since the YAML allows arbitrary shell script execution, the script could do something like `if (Math.random() > 0.5) { writeToS3(); } else { writeToECR(); }` or even `curl https://raw.githubusercontent.com/company/project/some_binary_that_calls_aws_apis; chmod +x some_binary_that_calls_aws_apis; ./some_binary_that_calls_aws_apis;` and there's no good, general purpose way to know what the side-effects of `codebuild` will be.
```
mtp1376, line 47: lol the examples here were amongst the top 3% programming jokes I've heard. +1 for the subsequent `chmod +x` and running the script and not leaving a non-fully-runnable script.

Issue #1231, opened 2022-09-07
Add a file/object list table to `aws_s3` module


depombo:
Each listed file in the table should have a foreign key to `bucket` in `aws_s3`. We should try to follow the same convention as the ACM modules. Perhaps simply be able to list and delete, but not upload for the first pass?

aguillenv:
it could be possible candidate having the table to list/delete and the functions to create/update?

dfellis:
So I figured a table would be good for listing files and their attributes (with the ability to mutate those attributes the normal IaSQL way), but reading/writing file data would be custom functions.

PR #1442, opened 2022-10-25
feat: add support for listobjects

yrobla:
Closes: https://github.com/iasql/iasql-engine/issues/1231


src/modules/0.0.22/aws_s3/entity/bucket_object.ts
```diff
@@ -0,0 +1,61 @@
...
+  id?: number;
+
+  @PrimaryColumn({
+    nullable: false,
+    type: 'varchar',
+  })
+  @cloudId
+  key: string;
+
+  @cloudId
```
aguillenv, line 29: not sure how this is behaving since it would be pointing 2 columns instead of one 🤔

src/modules/0.0.22/aws_s3/entity/bucket_object.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Bucket } from './bucket';
+
+@Entity()
+@Unique('uq_bucketobject_id_region', ['id', 'region'])
+@Unique('uq_bucketobject_bucket_key_region', ['key', 'region'])
+export class BucketObject {
+  @PrimaryGeneratedColumn()
+  id?: number;
+
+  @PrimaryColumn({
```
aguillenv, line 22: This should be `@Column`, otherwise, I think it might be generating a composite primary key in the db with `id` + `key`?

src/modules/0.0.22/aws_s3/mappers/bucket_object.ts
```diff
@@ -0,0 +1,130 @@
...
+export class BucketObjectMapper extends MapperBase<BucketObject> {
+  module: AwsS3Module;
+  entity = BucketObject;
+
+  async bucketObjectMapper(instance: _Object, ctx: Context, bucket: String, region: string) {
+    const bo: BucketObject = new BucketObject();
+    if (!instance.Key) throw new Error('Received a bucket object without a key');
+
+    // check for bucket instance
+    if (!Object.values(ctx.memo?.cloud?.Bucket ?? {}).length) {
```
aguillenv, line 18: why do we need this conditional? Doing 
```
bo.bucket =
         (await this.module.bucket.db.read(ctx, `${bucket}|${region}`)) ??
         (await this.module.bucket.cloud.read(ctx, `${bucket}|${region}`));
```
is not enough?

src/modules/0.0.22/aws_s3/mappers/bucket_object.ts
```diff
@@ -0,0 +1,130 @@
...
+    Prefix: k,
+  }));
+
+  deleteBucketObject = crudBuilder2<S3, 'deleteObject'>('deleteObject', (b, k) => ({ Bucket: b, Key: k }));
+
+  cloud = new Crud2<BucketObject>({
+    create: async (es: BucketObject[], ctx: Context) => {
+      // we cannot create buckets
+      const out: any = [];
+      return out;
```
aguillenv, line 50: maybe we can do a `db.delete` here?

src/modules/0.0.22/aws_s3/mappers/bucket_object.ts
```diff
@@ -0,0 +1,130 @@
...
+    },
+    read: async (ctx: Context, id?: string) => {
+      const enabledRegions = (await ctx.getEnabledAwsRegions()) as string[];
+
+      if (!!id) {
+        const { bucketName, objectKey, region } = this.idFields(id);
```
aguillenv, line 56: I do not think this might be working. For the `bucketName` is what I said about the `@cloudId` for the multiple join column, not sure what it is the property that it is being decorated. The, in the entity the property decorated is `key`, not `objectKey`

aguillenv: We might need to think about how to stress these cases in the test suite for coverage and to avoid future errors when we relate this entity with another one.

src/modules/0.0.22/aws_s3/mappers/bucket_object.ts
```diff
@@ -0,0 +1,130 @@
...
+            }
+          } catch (e) {
+            // do nothing, try the next bucket
+          }
+        }
+        return out;
+      }
+    },
+    update: async (es: BucketObject[], ctx: Context) => {
+      // we can just replace, no modification allowed
```
aguillenv, line 92: This comment seems misleading, here we are just restoring, right?

src/modules/0.0.22/aws_s3/rpcs/s3_upload_object.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (!bucketName || !bucketKey || !content || !contentType) {
+      return [
+        {
+          bucket: 'none',
+          key: 'none',
+          status: 'KO',
+          response_message:
+            'Please specify bucket name, key, content and content type of the object to upload',
+        },
+      ];
```
aguillenv, line 56: Not sure if it makes sense to throw in these cases

src/modules/0.0.22/aws_s3/sql/after_install.sql
```diff
@@ -22,3 +22,31 @@ UPDATE
   )
 EXECUTE
   FUNCTION block_s3_primary_key_update ();
+
+CREATE
+OR REPLACE FUNCTION block_s3_object_primary_key_update () RETURNS TRIGGER AS $block_s3_object_primary_key_update$
+    BEGIN
+        RAISE EXCEPTION 'Object bucket, key or region cannot be modified'
+        USING detail = 'A bucket object cannot be renamed or moved to another region or bucket', 
+        hint = 'If you want to rename or change bucket object, first remove it and then you could readd it.';
```
aguillenv, line 31:
```suggestion
        hint = 'If you want to rename or change bucket object, first remove it and then you could read it.';
```

aguillenv: ohh I see, updating my suggestion then

src/modules/0.0.22/aws_s3/rpcs/s3_clean_bucket.ts
```diff
@@ -3,6 +3,7 @@ import { S3, _Object } from '@aws-sdk/client-s3';
 import { AwsS3Module } from '..';
 import { AWS, crudBuilder2, crudBuilderFormat } from '../../../../services/aws_macros';
 import { Context, RpcBase, RpcResponseObject } from '../../../interfaces';
+import { BucketObject } from '../entity';
```
dfellis, line 6: Just a comment: We haven't release v0.0.22 yet, which is also the first iteration that has this RPC function. With this PR the "normal" cleaning logic should work because deleting all of the bucket object records should trigger a delete on all of those files, so we can probably just drop this RPC function entirely?

src/modules/0.0.22/aws_s3/rpcs/s3_upload_object.ts
dfellis: Yeah, I think that might be better, too.

src/modules/0.0.22/aws_s3/rpcs/s3_clean_bucket.ts
yrobla: maybe we can do in a follow up ? right now it is used on the ci for prisma, for example, and in codepipeline as well. I prefer if we do in separate prs, so we also refactor the dependencies

dfellis (approved):

PR #1443, opened 2022-10-25
feat: add multi-region support to codepipeline

yrobla:


dfellis (approved):

src/modules/0.0.22/aws_codepipeline/mappers/pipeline_declaration.ts
```diff
@@ -157,35 +160,43 @@ export class PipelineDeclarationMapper extends MapperBase<PipelineDeclaration> {
...
-
-        for (const pipeline of pipelines) {
-          const rawPipeline = await this.getPipelineDeclarations(client.cpClient, {
-            name: pipeline.name,
-          });
-          if (rawPipeline) {
-            const newPipeline = await this.pipelineDeclarationMapper(rawPipeline, ctx);
-            if (newPipeline) out.push(newPipeline);
+      if (!!id) {
+        const { region, pipelineName } = this.idFields(id);
```
aguillenv, line 165:
```suggestion
        const { region, name } = this.idFields(id);
```

src/modules/0.0.22/aws_codepipeline/mappers/pipeline_declaration.ts
```diff
@@ -157,35 +160,43 @@ export class PipelineDeclarationMapper extends MapperBase<PipelineDeclaration> {
...
+        if (enabledRegions.includes(region)) {
+          if (supportedRegions.includes(region)) {
+            const client = (await ctx.getAwsClient(region)) as AWS;
+            const pipeline = await this.getPipelineDeclarations(client.cpClient, {
+              name: pipelineName,
```
aguillenv, line 170:
```suggestion
              name,
```

src/modules/0.0.22/aws_codepipeline/mappers/pipeline_declaration.ts
```diff
@@ -201,8 +212,9 @@ export class PipelineDeclarationMapper extends MapperBase<PipelineDeclaration> {
       return out;
```
aguillenv, line 212: cannot comment in the exact line, but `cloud.update` need to be updated. `const cloudRecord = ctx?.memo?.cloud?.PipelineDeclaration?.[pd.name ?? ''];` is not true anymore, we need to get the cloudRecord using the `this.entityId(e)`

aguillenv (commented):
If we are letting the region be possible updated, maybe add a quick test creating a pipeline in one region and then updating the region to another one.

yrobla:
> If we are letting the region be possible updated, maybe add a quick test creating a pipeline in one region and then updating the region to another one.

Actually i do not believe that users should be able to update primary key of pipeline, because when they update, the pipeline should be already executing/executed. I will rather block updates on primary key, as i've done in other modules

aguillenv (approved):

PR #1445, opened 2022-10-26
add support for posthog

depombo:


aguillenv (approved):

src/services/telemetry.ts
```diff
@@ -45,16 +49,27 @@ export async function logEvent(
   uid?: string,
   deviceId?: string,
 ) {
-  if (!singleton) return;
+  if (!singletonAmp || !singletonPh) return;
```
dfellis, line 52: Should this be `||` or `&&`? Technically would continue on if either one is undefined and it looks like not defining amplitude would cause posthog to never run even if it is defined?

depombo: hmm let me gate each separately actually

dfellis (approved):

Issue #1449, opened 2022-10-27
/run endpoint shall process transactions in separate connections


yrobla:
for each query, when the per-statement flag is set, the engine should be picking a new sql connection. Otherwise the new queries are started without the previous one being processed

PR #1450, opened 2022-10-27
fix: start new connection for statement in byStatement mode

yrobla:

Closes: #1449


aguillenv (approved):
Question for a follow-up: Does it makes sense to keep the `byStatement` flag instead of always doing it `byStatement`? 

I know that I did this change and it is my bad not leaving a comment on why we had the distinction. If I recall correctly was to keep compatibility with the dashboard we had deployed at that moment and the development we were working on (the dashboard refactor I think). So maybe it makes sense now to get rid of this complexity and always parse and return the results of each statement separately?

yrobla:
> 

you mean, not offering the byUser support then, and always iterate one by one? Are you aware of any use case that could be broken due to that?

dfellis (approved):

PR #1453, opened 2022-10-28
Add `iasql_commit`, `iasql_rollback` and `iasql_preview`

aguillenv:


src/modules/0.0.22/iasql_functions/iasql.ts
```diff
@@ -1037,3 +1039,633 @@ export async function continueUpgrade(
...
+    if (c.changeType === AuditLogChangeType.DELETE) {
+      Object.entries(c.change.original).forEach(([k, v]: [string, any]) => (entity[camelCase(k)] = v));
+    } else if ([AuditLogChangeType.INSERT, AuditLogChangeType.UPDATE].includes(c.changeType)) {
+      Object.entries(c.change.change).forEach(([k, v]: [string, any]) => (entity[camelCase(k)] = v));
+    }
+    if (changesByEntity[entityName]) {
+      changesByEntity[entityName].push(entity);
+    } else {
+      changesByEntity[entityName] = [entity];
+    }
```
dfellis, line 1205: Nit: This could just be

```suggestion
    changesByEntity[entityName] = changesByEntity[entityName] ?? [];
    changesByEntity[entityName].push(entity);
```

src/modules/0.0.22/iasql_functions/iasql.ts
```diff
@@ -1037,3 +1039,633 @@ export async function continueUpgrade(
...
+        toReplace,
+        toDelete,
+        dryRun,
+        changesByEntity,
+      );
+    } else {
+      /*
+       * If no changes found, what to do?
+       * If we sync everything and a user inserts something during the sync, could be detected and override the changes.
+       * We need to bring changes on AWS only always, ignore db only changes, but what to do with existing db records that might have changed? how to avoid overrides?
```
dfellis, line 1115: Perhaps apply the filter on records to ignore based on the audit log to *both* cloud and db? If both changed simultaneously, we should still defer to the user's choice, right?

src/modules/0.0.22/iasql_functions/iasql.ts
```diff
@@ -1037,3 +1039,633 @@ export async function continueUpgrade(
...
+        dbId,
+        rootToLeafOrder,
+        context,
+        toCreate,
+        toUpdate,
+        toReplace,
+        toDelete,
+        dryRun,
+        changesByEntity,
+      );
```
dfellis, line 1110: Interesting that you're doing the remote sync first instead of the apply of the user-generated changes. What's your reasoning here?

src/modules/0.0.22/iasql_functions/index.ts
```diff
@@ -3,8 +3,10 @@ import { ModuleBase } from '../../interfaces';
 import { IasqlOperationType } from './entity';
 import {
   IasqlApply,
```
dfellis, line 5: Will the Apply And Sync functions be deprecated in this PR, or are you thinking of doing so in a follow-up PR?

src/modules/0.0.22/iasql_functions/iasql.ts
aguillenv: oh, my bad to leave this misleading comment. I think I've solved this by [getting the latest changes](https://github.com/iasql/iasql-engine/pull/1453/files#diff-d1cec5ce18674d505cb6eddf5fe3c9421b5c2279e25df59363a3edd73060e80fR1282-R1283) every time we `db.read` during `commitSync` and then [filter the changes if necessary](https://github.com/iasql/iasql-engine/pull/1453/files#diff-d1cec5ce18674d505cb6eddf5fe3c9421b5c2279e25df59363a3edd73060e80fR1318-R1332)

src/modules/0.0.22/iasql_functions/index.ts
aguillenv: Now that you have mentioned this, there are some issues still pending in this PR:
- I cannot deprecate `apply` and `sync` until we release 0.0.22
- `install` and `upgrade` would need to be modified to use the new `commitSync`, I need to see what might have changed there.

src/modules/0.0.22/iasql_functions/iasql.ts
aguillenv: Since they are sequential actions, my idea was, on commit, to always make sure we have the DB updated before applying changes. Also, it is easier to identify which changes need to be synced and which ones need to be ignored during the `sync` phase since we know which changes need to be made in the cloud. The other way around could be trickier to apply the filters.

src/modules/0.0.22/iasql_functions/index.ts
aguillenv: Should we run a full `commit` while `install` and `update`? If so, and a commit is already in process? should we wait for completion?

src/modules/0.0.22/iasql_functions/iasql.ts
```diff
@@ -1019,7 +1019,9 @@ export async function continueUpgrade(
     INSERT INTO aws_credentials (access_key_id, secret_access_key)
     VALUES ('${creds.access_key_id}', '${creds.secret_access_key}');
   `);
+  const orm = context.orm;
   await commit(dbId, false, context, true);
+  context.orm = orm;
```
dfellis, line 1024: How do these two lines do anything?

dfellis (approved):

src/modules/0.0.22/iasql_functions/iasql.ts
aguillenv: because inside the `commit` fn I re-assign the `context.orm` with the inner `orm` object created, then that connection is closed and here it would fail to continue the upgrade.

dfellis: Got it. Can we add a comment to that effect? Or perhaps push that behavior into the `commit` function instead of requiring the callers of the function to replicate this logic?

src/modules/0.0.23/iasql_functions/iasql.ts
```diff
@@ -1039,3 +1038,878 @@ export async function continueUpgrade(
...
+          force,
+          crupdes,
+          dryRun,
+          changesToCommit,
+        );
+        if (dryRun) return applyRes;
+      }
+    } catch (e) {
+      logger.info('Something failed. Starting commit apply phase for all modules');
+      return await commitApply(dbId, installedModulesSorted, context, force, crupdes, dryRun);
```
dfellis, line 1114: This behavior feels a little bit dangerous, as it could blow away changes in the cloud account that were not touched by the user in IaSQL.

What's the reasoning behind it?

aguillenv: Well, I think we talked offline about this. This will help in cases where a dependency for a change we need to apply has been deleted or modified remotely outside of IaSQL. If we just take into account the changes, it will fail and on sync it will delete the user's change, so we do a full apply to restore resources in case of deletion in the cloud and see if the change we are trying to apply succeeds. This indeed could restore resources that might have been deleted from the cloud but we need to prioritize our DB as the SoT.

dfellis (approved):

depombo (approved):

src/modules/0.0.23/iasql_functions/iasql.ts
dfellis: I approved with this answer. I think the solution is going to be not fully satisfactory either way because a failure here is not going to 100% do what users want all of the time. I do agree the opposite behavior would be worse, though.

PR #1462, opened 2022-10-31
Unsurprising Functions and Scalability RFC

dfellis:


rfcs/005 - Unsurprising IaSQL Functions RFC.md
```diff
@@ -0,0 +1,174 @@
...
+
+**Pros:**
+
+- "Solves" the issue, by breaking connections that try to do multi-statement at all.
+
+**Cons:**
+
+- "Solves" the issue, by breaking connections that try to do multi-statement at all.
+
+##### Implement a Postgres proxy in the Engine
```
yrobla, line 150: that sounds as a reasonable approach to me... we should focus on minimizing the load + latency

rfcs/005 - Unsurprising IaSQL Functions RFC.md
```diff
@@ -0,0 +1,174 @@
...
+
+`pgbouncer` is a commonly-used service in conjunction with Postgres to allow thousands of worker processes connect to Postgres without exploding the server with too many child processes. [The `statement`-level in its `pool_mode` configuration](https://www.pgbouncer.org/config.html#pool_mode) bans requests that span more than one statement (and is likely why ArcType and other database clients simply split the requests by statement and issue them separately).
+
+**Pros:**
+
+- "Solves" the issue, by breaking connections that try to do multi-statement at all.
+
+**Cons:**
+
+- "Solves" the issue, by breaking connections that try to do multi-statement at all.
```
yrobla, line 148: pros=cons?

rfcs/005 - Unsurprising IaSQL Functions RFC.md
dfellis: I am also leaning towards it being the first thing we do, also because it lets us sidestep an issue with the current RDS approach (no SSL cert), but I have realized after writing that up that there's still going to be an issue if the user tries to write their query in a real transaction -- that would be sent to the database as a singular "statement" but would run into the same data issues as before.

rfcs/005 - Unsurprising Functions and Scalability RFC.md
```diff
@@ -0,0 +1,215 @@
...
+
+This singular docker container can also be pushed to Dockerhub and make the deployments even simpler, being just downloading a new docker image, restarting the container, and pruning the dead image(s), with deployments taking seconds rather than minutes, though with a brief moment of downtime (before we implement the routing layer, which could hold external connections open until downstream is up again, then making it zero-downtime).
+
+It would also be possible to consider docker-container-per-user-database (with multiple of them per EC2 instance), which would allow us to drop maintaining multiple module versions at the same time within the engine. `iasql_upgrade` would become an *actual* upgrade of the engine to a newer version triggered by the end user, instead, but this and the routing layer could be done progressively in the future after the main transition in the four points above.
```
depombo, line 70: I want to echo that like (4) and the previous sentence this would be follow-up work. 

I hadn't thought of this. However, to fully get rid of multiple module versions the user call to `upgrade` would also require a deploy to their respective container?

rfcs/005 - Unsurprising Functions and Scalability RFC.md
```diff
@@ -0,0 +1,215 @@
...
+- No weird injection in TypeORM to capture it's output, just configuring it to talk to a different port than normal.
+
+**Cons:**
+
+- Need to write a Rust extension, and faking the standard postgres connection may be more difficult than expected (it may need to interpret and inject into the main connection)
+- Need to expose all ports between the Postgres server and the IaSQL engine, and RDS cannot be used
+
+#### Change Postgres Behavior
+
+##### Fork Postgres, make it like MySQL
```
depombo, line 148: we could also port everything to mysql instead of postgres. in theory, typeorm is an abstraction on top of both

depombo (approved):

rfcs/005 - Unsurprising Functions and Scalability RFC.md
```diff
@@ -0,0 +1,215 @@
...
+
+This singular docker container can also be pushed to Dockerhub and make the deployments even simpler, being just downloading a new docker image, restarting the container, and pruning the dead image(s), with deployments taking seconds rather than minutes, though with a brief moment of downtime (before we implement the routing layer, which could hold external connections open until downstream is up again, then making it zero-downtime).
+
+It would also be possible to consider docker-container-per-user-database (with multiple of them per EC2 instance), which would allow us to drop maintaining multiple module versions at the same time within the engine. `iasql_upgrade` would become an *actual* upgrade of the engine to a newer version triggered by the end user, instead, but this and the routing layer could be done progressively in the future after the main transition in the four points above.
+
+### Alternatives Considered
```
yrobla, line 72: what about considering a pub/sub approach? https://github.com/pvarentsov/nestjs-pg-notify
or websockets: https://github.com/supabase/realtime

yrobla (commented):
added a pair of alternatives

rfcs/005 - Unsurprising Functions and Scalability RFC.md
```diff
@@ -0,0 +1,215 @@
...
+2. Efficiency gains as well as a reduction in complexity can be had by switching to the [`pgsql-http`](https://github.com/pramsey/pgsql-http) extension and having the RPC functions simply wait for the HTTP response from the engine, though they must *all* finish before the HTTP connection times out. This requires getting off of RDS, as this [is not one of the blessed extensions](https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-extensions.html). The HTTP timeout issue can be avoided entirely if the engine is a sidecar process to the Postgres instance and they communicate over `localhost`.
+3. Take the final few HTTP endpoints exposed to the end users and dashboard in the engine and turn most of them into RPC functions in the `metadata` table, and close off HTTP access to the outer world entirely, exposing only the database. The `/run` endpoint would turn into a lambda function for the dashboard, instead. This lets us simplify our deployment artifact into a singular Docker container with all three (Postgres, Engine, pgbouncer) communicating directly to each other.
```
aguillenv, line 65: this would mean that endpoints like connect will now be RPC calls?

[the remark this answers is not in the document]
dfellis: The purpose is to allow any sort of SQL querying at all. Web browsers can't speak Postgres' wire protocol. ;)

rfcs/005 - Unsurprising Functions and Scalability RFC.md
dfellis: Basically there would be no `iasql_upgrade` as a callable function anymore. We'd just make upgrading from earlier versions of IaSQL something it does during startup just like Postgres auto-upgrades its on-disk representation during startup if you point a new version of Postgres at an older version's data (only up to so many versions behind, though, which we would also preserve).

So the `upgrade` code wouldn't go away, but 1) it would be fully contained in a singular logical block instead of split between "old version upgrade RPC triggering" and "new version continuation function that needs to be compatible with all supported older versions" and 2) it would only run during startup before anything else is set up on the engine, so there's no weird issues of the upgrade running in parallel with an apply or install.

Upgrading in the dashboard would still exist, but it would call something else (perhaps a lambda to upgrade the database/docker in question)? And yes, this is definitely not required for the initial pass, but this particular architecture makes it possible to have per-user IaSQL DB versioning (like we currently support) but without all of the complexities to the codebase the current approach has, so I do think pursuing it will be worthwhile (no longer needing to port in-progress changes between versions if you didn't make the release cut-off, no need to guard against improper changes to prior versions, or worries about backwards incompatible changes in the non-module code necessitating things like `Mapper2` classes, etc).

rfcs/005 - Unsurprising Functions and Scalability RFC.md
dfellis: We would need to dig in a bit on that, but I'm not sure we'd have something like `dblink` to let us break stored procedure behavior?

There are also several other downsides to MySQL (for us, at least) that I think make it not worth pursuing, but let me know if you disagree on any of these:

1. MySQL's DB type system is looser, so several data-correctness enforcements we have in the database would need to be moved into the engine itself, which increases development load for us and reduces immediate feedback on bad `INSERT`s for users. For instance in MySQL [ENUMs accept garbage input and batch it all into a singular bucket instead of failing outright](https://dev.mysql.com/doc/refman/5.7/en/enum.html)
2. MySQL doesn't have an array type, so all uses of arrays currently would need to be converted to JSON (which is not typed at all). This is similar to the first point, but rather than being about weird auto-conversion of bad data into a fixed type and cause unexpected behavior, the bad data stays in its original form and could crash our code, instead.
3. Postgres has some types that help us better represent the data we're working with, like the CIDR type, and also allows us to create types to do things better, with the [composite types](https://www.postgresql.org/docs/9.6/rowtypes.html) that we can use where it makes sense for the data to be represented in a clearer way to end users.
4. MySQL is controlled by one of the more capricious corporations out there: Oracle, the company that sued to make API *definitions* copyrightable, which would have effectively destroyed software development in the US. So there is some amount of risk building in that ecosystem if we are very successful, as we don't have the legal budget of Google to fend them off.

rfcs/005 - Unsurprising Functions and Scalability RFC.md
dfellis: So pub/sub is functionally identical to the current approach:

You publish a request and then poll on the response. The only difference I can see is the message passing between the database and engine being two tables instead of one?

The project you linked to is itself built on top of another library called `pg-listen`, the mission-statement of which being that [they didn't have a reliable communication layer](https://github.com/andywer/pg-listen#why-another-package). Graphile Worker *has* been reliable for us in terms of communication between the engine and postgres, so removing it is *not* for that reason. The reasons why we want to remove it are:

1. It adds hundreds of milliseconds of latency to the RPC calls (which are generally not noticeable when AWS API calls take *minutes* but not every cloud API or service is going to be that slow)
2. The current structure causes unexpected behavior from our functions because the requested work is not happening in the connection process.

Switching to a pub/sub approach won't resolve either of those concerns.

Sockets/Websockets *are* an answer that could be considered separate. Since this is our backend database communicating to our backend engine, we could use a standard socket there, and between the dashboard and the engine we could use a websocket.

The socket approach with the database is doable, but we would have to write our own postgres extension instead of re-using an existing one (as SQL doesn't really have a concept of streams, so we'd have to create new extension functions to expose consuming a stream) and then we'd need to write SQL code to parse our streaming protocol out of the socket extension and then write back into it.

It's powerful, but since we just need an RPC where we provide up front all of the data we need to get our response, it feels like overkill.

Also needing a custom Postgres extension kicks us off of RDS, which then makes just using `pgsql_http` on an engine instance running in the same `localhost` the easier and potentially more-reliable approach.

*But* using that project to provide query support to the dashboard looks like it could be much better than the Lambda approach I mentioned. We would need to move the per-statement splitting into the dashboard entirely, but that seems simple enough.

rfcs/005 - Unsurprising Functions and Scalability RFC.md
```diff
@@ -0,0 +1,215 @@
...
+## Expected Semver Impact
+
+This would be considered a major version bump as the spin-up of new databases will change, as well as how it is deployed in every environment, but if considering only the user experience *after* the database exists, it would just be a patch bump, as it is simply fixing a bug in the RPC functions.
+
+## Affected Components
+
+- The scheduler is dropped, `pgsql-http` replaces it
+- `docker-compose.yml` is dropped, a new singular `Dockerfile` replaces it
+- `pgbouncer` becomes a hard dependency
+- The `metadata` database gains RPC support, all external Express endpoints become `metadata` RPC calls except `/run` which becomes a Lambda for the Dashboard
```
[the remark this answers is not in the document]
dfellis: Yes. I was figuring being able to increase the timeout to 15 minutes would be a decent enough mitigation, but what @yrobla pointed out [in this comment chain](https://github.com/iasql/iasql-engine/pull/1462#discussion_r1015553543) with Supabase's "realtime" Elixir Websocket-to-Postgres service looks like a more ideal solution that we don't have to build ourselves.

rfcs/005 - Unsurprising Functions and Scalability RFC.md
```diff
@@ -0,0 +1,215 @@
...
+
+On top of all of that, this direction causes our production and staging environments to become further and further removed from our local and test environments, which increases the maintenance burden and makes the likelihood of bugs getting through higher, as developer testing of certain paths differs more from user usage.
+
+The piecemeal approach to scalability would guide us to a much more complex production environment with worse latencies and a harder-to-manage codebase, and that is why it has been rejected (while we still can).
+
+## Expected Semver Impact
+
+This would be considered a major version bump as the spin-up of new databases will change, as well as how it is deployed in every environment, but if considering only the user experience *after* the database exists, it would just be a patch bump, as it is simply fixing a bug in the RPC functions.
+
+## Affected Components
```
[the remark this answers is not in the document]
dfellis: So I was debating mentioning that or not, because this repo on its own doesn't deal with any of that. Our production/staging deployment setup does. A bit of where having the RFC inside of the OSS repo breaks down a bit.

aguillenv (approved):

rfcs/005 - Unsurprising Functions and Scalability RFC.md
dfellis: I didn't add these into the alternatives considered section because they wouldn't resolve the "next SQL statement on the same connection has access to the new data", at least not on their own, but I incorporated the websocket suggestion into the `/run` replacement text, where I think it is a better fit.

rfcs/005 - Unsurprising Functions and Scalability RFC.md
```diff
@@ -42,32 +43,39 @@ To be more clear on what is going on and why in the RPC function behavior, a des
...
 
 Choosing a new mechanism to implement the RPC functionality and patching up the SPOFs over time, always being in a deployable state from one patch to the next, is a positive goal, and the path we would be much more likely forced to take if we had significant usage, but as we do not, more "ideal" architectures that we can't find a gradual transition to were also considered, and in the end this is the path recommended. In the alternatives considered, all of the other RPC mechanisms considered are covered, as well as an exploration of how the piecemeal approach hits a dead-end, though the details are sparser there as the specifics depends on the RPC mechanism being transitioned into.
 
+Additionally, there is a problem on the consumption of these functions. When connecting our engine with a client, we rely on an express endpoint, that is deployed on an AWS server and connected via load balancers with some fixed timeout.
+This approach is not suitable for communications with a client, basically due to the async nature of our engine. When a request is sent, the engine is processing this on the background, and is returning a response in a certain amount of time.
```
dfellis, line 63: Nit

```suggestion
This approach is not suitable for communications with a client, basically due to the async nature of our engine. When a request is sent, the engine is processing this on the background, and is returning a response in an unknown amount of time outside of our control, depending on the AWS cloud API calls involved.
```

rfcs/005 - Unsurprising Functions and Scalability RFC.md
```diff
@@ -42,32 +43,39 @@ To be more clear on what is going on and why in the RPC function behavior, a des
...
+This response time can be quite high depending on the queries, and will hit a maxium easily for some queries, either causing failures or not letting the user discover the result of the query.
```
dfellis, line 64: Typo

```suggestion
This response time can be quite high depending on the queries, and will hit a maximum easily for some queries, either causing failures or not letting the user discover the result of the query.
```

depombo (approved):

rfcs/005 - Unsurprising Functions and Scalability RFC.md
```diff
@@ -0,0 +1,239 @@
...
+- MySQL doesn't have an array type, so all uses of arrays currently would need to be converted to JSON (which is not typed at all). This is similar to the first point, but rather than being about weird auto-conversion of bad data into a fixed type and cause unexpected behavior, the bad data stays in its original form and could crash our code, instead.
+- Postgres has some types that help us better represent the data we're working with, like the CIDR type, and also allows us to create types to do things better, with the composite types that we can use where it makes sense for the data to be represented in a clearer way to end users.
+- MySQL is controlled by one of the more capricious corporations out there: Oracle, the company that sued to make API definitions copyrightable, which would have effectively destroyed software development in the US. So there is some amount of risk building in that ecosystem if we are very successful, as we don't have the legal budget of Google to fend them off.
```
mtp1376, line 170: Also, we won't be able to use graphile :D

mtp1376 (approved):

rfcs/005 - Unsurprising Functions and Scalability RFC.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Current Status
+
+### Proposed
+
+2022-11-04
+
+### Accepted
+
+YYYY-MM-DD
```
dfellis, line 11:
```suggestion
2022-11-11
```

PR #1490, opened 2022-11-03
merge all webcontent into one docusaurus instance

depombo:
porting over the landing page to docusaurus using react + [infima](https://infima.dev/) wasn't too bad. there are a few outdated things in the current landing page, but I figured I would port it as-is and then change it in follow up PRs. the dark mode doesn't work as nicely, but that can also be follow ups


dfellis:
Perhaps the top-level `docs` directory should be renamed `site`?

aguillenv (approved):

dfellis (approved):

PR #1507, opened 2022-11-07
Set id before save updated log group entity

aguillenv:


dfellis (approved):

src/modules/0.0.22/aws_cloudwatch/mappers/log_group.ts
```diff
@@ -83,6 +83,7 @@ export class LogGroupMapper extends MapperBase<LogGroup> {
         const newEntity = this.logGroupMapper(newObject, e.region);
         if (!newEntity) continue;
         // Save the record back into the database to get the new fields updated
+        newEntity.id = e.id;
```
mtp1376, line 86: In case we're calling the `.cloud.create` from another module, this is going to create a new entity in the database (doing what `.db.create` should've done). I think the `cloud.create` function should just create the entity in the cloud, and if a local version in the database already exists it should update the values.

I suggest you to go with a similar approach to [this one](https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_iam/index.ts#L435-L436).

aguillenv: That's the idea we are trying to achieve here. We do not need to do a `db.read` because we already have the db record, it is the `e` variable. If you see the [next line](https://github.com/iasql/iasql-engine/pull/1507/files#diff-35ab988197742cc20fbcd0e55b8ba2e2355a2aadbf8fc1aa3f05135f716b50c4R87), we are trying to `db.update` with the new information we have from the cloud (arn for example). The problem is that here the primary key of the entity is the auto-generated `id`, and `newEntity` just has the cloud information and no `.id` defined, so the `db.update` might be doing nothing, or worst, it could be creating a new record in the DB. With this fix, adding the `id` we already have from the db record to `newEntity`, the `db.update` will know exactly which record needs to update.

mtp1376: I understand your point of view. In case that this `cloud.create` is being triggered by a DB record, you're right (we already have the db record in that case). But let's suppose another module is calling the `cloud.create` with an object it is passing and expects a cloud object to be created. In that case, the `e.id` would not have a value and it will result in `db.update` to create a new database entry (unexpected behavior).

aguillenv: Well, I'm not sure if that should not happen. We rely on the idea that the database will be our Source of Truth in terms of resource creation. If from another module we want to create a new resource we might need to add the corresponding record to the db and let the engine does it work

dfellis: My opinion is that the work @aguillenv is doing towards two-way mode in the other PR will eventually make this a moot point. If you create a cloud resource and forget to save it, the next step we re-sync changes from the cloud into the database and then there it is. :)

But in the meantime, just wrapping this whole bit in `if (e.id) { ... }` should resolve your concern and also not require much extra code?

mtp1376: Yes. If we're doing an `iasql_apply` that's going to happen. But it's not always an `iasql_apply`. For example, an RPC like `ecr_build` is going to create (and eventually delete) cloud resources in one call. In that case we might want the `cloud.create` to just create the cloud entities and not the database ones.

mtp1376: - ECR build is trying to create the cloud resource [+](https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_ecr/rpcs/build.ts#L273)
- And then deleting it without worrying about the database entry [+](https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_ecr/rpcs/build.ts#L129)

mtp1376: > My opinion is that the work @aguillenv is doing towards two-way mode in the other PR will eventually make this a moot point. If you create a cloud resource and forget to save it, the next step we re-sync changes from the cloud into the database and then there it is. :)
> 
> But in the meantime, just wrapping this whole bit in `if (e.id) { ... }` should resolve your concern and also not require much extra code?

Right, no additional code necessary. I think it was more like a philosophical thing (maybe without much real-world application :))).

aguillenv: > * ECR build is trying to create the cloud resource [+](https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_ecr/rpcs/build.ts#L273)
> * And then deleting it without worrying about the database entry [+](https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_ecr/rpcs/build.ts#L129)

ahh, I see. Valid point here. let me add the conditional for this case, but yes, as David says, this should not be a problem with the 2-way

src/modules/0.0.22/aws_cloudwatch/mappers/log_group.ts
```diff
@@ -83,6 +83,7 @@ export class LogGroupMapper extends MapperBase<LogGroup> {
...
         await this.module.logGroup.db.update(newEntity, ctx);
```
aguillenv, line 87:
```suggestion
        if (e.id) {
          newEntity.id = e.id;
          await this.module.logGroup.db.update(newEntity, ctx);
        }
```

src/modules/0.0.22/aws_cloudwatch/mappers/log_group.ts
dfellis: > Yes. If we're doing an `iasql_apply` that's going to happen. But it's not always an `iasql_apply`. For example, an RPC like `ecr_build` is going to create (and eventually delete) cloud resources in one call. In that case we might want the `cloud.create` to just create the cloud entities and not the database ones.

So the `ecr_build` RPC is self-contained and (because we only allow one job to run at a time) never happens in parallel with an `apply`/`install`/etc, so as long as it cleans up after itself, or writes to the database anything that will persist after that RPC function call is done, we're fine.

In the future two-way mode, the `commit` logic can run in parallel with these RPCs, *but* whatever they're doing is immaterial, as `commit` will just periodically re-sync the database with the state of the cloud and if the record was temporarily created and then removed later doesn't really matter. But in that future the RPC function *doesn't* have to worry about creating those database entries for things that persist, which would let us remove most of the `db.create/update/delete` logic inside of the `cloud.create/update/delete` methods, which will reduce the cognitive load when writing them.

PR #1509, opened 2022-11-07
Fix how to set user and password across tests

aguillenv:


dfellis (approved):

depombo:
what is this fixing exactly?

aguillenv:
> what is this fixing exactly?

Basically, it was always using the default username and password, even when we were "passing" the `pgUser` and `pgPassword` generated at `connect`. The reason is that by the time the `runQuery` fn was called the `pgUser` and `pgPassword` were not defined yet. Jest run first the bindings first, before any test, and assigned to `query` the resulting function. In that binding call, the variables were not defined yet.

depombo (approved):

PR #1513, opened 2022-11-07
Add the `ecs_simplified` blog post

mtp1376:


site/blog/2022-11-04.mdx
```diff
@@ -0,0 +1,123 @@
...
+- deploy a load balancer as the point-of-contact for your app
+- create a target group for the load balancer and register the ECS tasks in it
+- add a new listener to your load balancer and connect it to the target group
+- create a security group, and you need to allow the port your app is listening on in that security group
+- attach the security group above to your load balancer
+- create a CloudWatch log group for your ECS task
+- create an ECS cluster, and definitely the task definition as well
+- Oh, and create an ECR repository to push your images to be run on the container
+
+I'm not going to continue this long list, since I've already got a headache. Doing those steps manually is going to give you a headache as well, so why bother doing all those steps yourself and risking different errors you might face when deploying your containerized app? You don't really have the time for the random IAM-related errors AWS demanding you to resolve. Besides, you already have your codebase ready and the `Dockerfile` is there, so why not just run a simple command doing something that should be simply done?
```
dfellis, line 16: Nit

```suggestion
I'm not going to continue this long list, since I've already got a headache. Doing those steps manually is going to give you a headache as well, so why bother doing all those steps yourself and risking different errors you might face when deploying your containerized app? You don't really have the time for the random IAM-related errors AWS is demanding you to resolve. Besides, you already have your codebase ready and the `Dockerfile` is there, so why not just run a simple command doing something that should be simply done?
```

site/blog/2022-11-04.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+In this post, we're going to discover an IaSQL module that's built to make deploying to ECS, simplified. Most of the details for deploying a container to ECS are the same (load balancers, security groups, IAM roles, etc), and we have created the `ecs_simplified` module for you so that you can give it your codebase and get your app deployed to ECS **in just 11 minutes**. All of those needed resources are going to be created automatically in your AWS account, and you'll have full access over them either through the AWS dashboard, or through your IaSQL database.
```
dfellis, line 18: I know it's not our fault, but "in just 11 minutes" feels like forever for something that *should* be able to spin up in seconds, so not sure about the emphasis on that?

site/blog/2022-11-04.mdx
```diff
@@ -0,0 +1,123 @@
...
+```sql title="Deploy the folder 'examples/ecs-fargate/prisma/app' of the 'iasql-engine' repo to ECS on my AWS account"
+-- Install the "high-level" aws_ecs_simplified IaSQL module
+SELECT iasql_install('aws_ecs_simplified');
+-- Create the load balancer, ECS service, task definition, etc. All ready afterwards for us to build and push to deploy the app.
+INSERT INTO ecs_simplified (app_name, app_port, image_tag, public_ip) VALUES ('simple-express', 8088, 'latest', true);
+SELECT iasql_apply();
+-- Use the "high-level" ecr_build function to build the folder "examples/ecs-fargate/prisma/app" from "iasql-engine" Github repo and push it to ECR
+SELECT ecr_build(
+  'https://github.com/iasql/iasql-engine/',
+  (SELECT id FROM repository WHERE repository_name = 'simple-express-repository')::varchar(255),
```
dfellis, line 38: Do you need this `varchar` cast? Postgres will auto-cast where reasonable. I think it's worth checking if we can remove that part.

site/blog/2022-11-04.mdx
```diff
@@ -0,0 +1,123 @@
...
+  './examples/ecs-fargate/prisma/app',
+  'ghp_XXX', -- replace your github personal access token here
```
dfellis, line 40: Public repos shouldn't need the github access token, right? Can this be optional?

site/blog/2022-11-04.mdx
```diff
@@ -0,0 +1,123 @@
...
+  'main'
```
dfellis, line 41: It may make sense to put comments after each of these lines explaining what the arguments are. I know this is our `main` branch, but the reader may not know that and if they're still using `master` as the primary branch name, they might copy-paste this string as-is and not understand why it's failing.

site/blog/2022-11-04.mdx
```diff
@@ -0,0 +1,123 @@
...
+);
+```
```
dfellis, line 43: Alternatively, add in a paragraph here explaining the steps in the SQL above. Then you can include clickable hyperlinks to other things, like some of our documentation, in the explanation.

site/blog/2022-11-04.mdx
```diff
@@ -0,0 +1,123 @@
...
+```sql title="Create a security group named 'allow-all' attached to the VPC with ID 'vpc-f56cff91'. Then add a rule to it allowing all incoming traffic on port '80' from '0.0.0.0/0'"
+-- Install the aws_security_group IaSQL module
+SELECT iasql_install('aws_security_group');
+-- "Create a security group named `allow-all` attached to the VPC with ID `vpc-f56cff91`"
+INSERT INTO security_group (description, group_name, vpc_id)
+VALUES ('allows all incoming traffic on port 80', 'allow-all', (SELECT id FROM vpc WHERE vpc_id = 'vpc-08c0a68025c2983a2'));
```
dfellis, line 49: You have two different `vpc_id`s referenced in this text.

I would also suggest restructuring this to select the `vpc_id` through the [`subnet`s on the `service` record](https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_ecs_fargate/entity/service.ts#L67).

Something like:

```suggestion
-- "Create a security group named `allow-all` attached to the VPC the ECS service is on"
INSERT INTO security_group (description, group_name, vpc_id)
VALUES ('allows all incoming traffic on port 80', 'allow-all', (
  SELECT id
  FROM vpc
  INNER JOIN subnet ON subnet.vpc_id = vpc.id
  INNER JOIN service ON service.subnets[1] = subnet.subnet_id
  WHERE service.name = 'simple-express-service'
 ));
```

It's a bit longer, but much less error prone, and truly runnable with the Button links without requiring any edits on the user's side, if they want.

site/blog/2022-11-04.mdx
```diff
@@ -0,0 +1,123 @@
...
+);
+SELECT iasql_apply();
+
+SELECT ecr_build(
+  'https://github.com/iasql/iasql-engine/',
+  (SELECT id
+    FROM repository
+    WHERE repository_name = 'simple-express-repository')::varchar(255),
+  './examples/ecs-fargate/prisma/app',
+  'ghp_XXX', -- replace your github personal access token here
```
dfellis, line 76: If we can eliminate this for public github repos, that would be great. Zero edits of these scripts is the ideal scenario. Let them edit it *after* they see something working.

site/blog/2022-11-04.mdx
```diff
@@ -0,0 +1,123 @@
...
+```
+- This command tells IaSQL to go clone `iasql-engine` repository, build an image on the subdirectory specified, and then push it to the ECR repository created earlier by `ecs_simplified` module. Running the above command will automatically create a CodeBuild project and the related roles, etc. Then it'll start a build, and after it's successful all the created resources are deleted to ensure there won't be any additional charges to your AWS account.
+- To access your app on the cloud, get the load balancer address and use your browser to access the live version of it:
+```sql
+SELECT load_balancer_dns
+FROM ecs_simplified
+WHERE app_name = 'simple-express';
+```
+Mine is `simple-express-load-balancer-94898832.us-east-2.elb.amazonaws.com` so I can access it by opening the following address in my browser:
+-  http://simple-express-load-balancer-94898832.us-east-2.elb.amazonaws.com:8088/health/
```
dfellis, line 123: We probably shouldn't include this statement except if we're willing to leave that URL up in perpetuity.

site/blog/2022-11-04.mdx
```diff
@@ -0,0 +1,123 @@
...
+- Now the ECS is waiting for an image to be pushed to your ECR repository to run it. You can get the URI for the ECR repository by running the `SELECT repository_uri FROM ecs_simplified WHERE app_name = 'simple-express'` query.
+  - In the next step, we'll automatically build an image for the code in Github repo and then push it to this URI (all through SQL and using another high-level function named `ecr_build`).
+```sql
+SELECT ecr_build(
+  'https://github.com/iasql/iasql-engine/',
+  (SELECT id
+    FROM repository
+    WHERE repository_name = 'simple-express-repository')::varchar(255),
+  './examples/ecs-fargate/prisma/app',
+  'ghp_XXX', -- replace your github personal access token here
```
dfellis, line 111: Re-emphasizing here. It could just be `NULL` and then we don't mess with the github token, but can still have an obvious call-out in the comments about adding a github personal access token if it's a private repo.

site/blog/2022-11-04.mdx
mtp1376: I tried using the query without that cast, but it doesn't work :(.

site/blog/2022-11-04.mdx
mtp1376: Oh that's a really messy AWS CodeBuild stuff. I tried a bunch of times without any personal access tokens and it cannot even pull the public repos. God knows why it needs a PAT for just cloning a repo. I'll try again to see if it can be done without a PAT (in case it was because of IAM role or something else and I've misjudged).

mtp1376: In the worst case, we can sign up for a Github account and create a PAT (with only access to the public repos) and put it here? Or maybe put it on the engine?

site/blog/2022-11-04.mdx
mtp1376: This example didn't meant to be related to the `ecs_simplified` context. I was trying to tell the users that they can do both the high-level and low-level stuff using IaSQL. But if it's not self-explanatory enough, I should add some more info showing that intention?

site/blog/2022-11-04.mdx
mtp1376: Wait, I have another idea: we can set the source provider in the CodeBuild to "No Source" and do a normal `git clone` at the first of the `buildspec.yaml` file. In case a PAT is provided by the user, we can go with the normal flow.

mtp1376: I tested that idea and it worked. I'll change the `ecr_build` to use that behavior.

site/blog/ecs-simplified.mdx
```diff
@@ -0,0 +1,125 @@
...
+SELECT iasql_apply();
+-- Use the "high-level" ecr_build function to build the folder "examples/ecs-fargate/prisma/app" from "iasql-engine" Github repo and push it to ECR
+SELECT ecr_build(
+  'https://github.com/iasql/iasql-engine/',
+  (SELECT id FROM repository WHERE repository_name = 'simple-express-repository')::varchar(255),
+  './examples/ecs-fargate/prisma/app',
+  'ghp_XXX', -- replace your github personal access token here
+  'main'
+);
+```
```
aguillenv, line 45: since this piece of code is meant to be an example I would remove the comments asking actions like `-- replace your github personal access token here` and even maybe rename the app and use more placeholders. Taking a quick view I got confused between this code and the one in the next section and thought it was some kind of duplication.

site/blog/ecs-simplified.mdx
```diff
@@ -0,0 +1,125 @@
...
+```sql title="Create a security group named 'allow-all' attached to the VPC with ID 'vpc-f56cff91'. Then add a rule to it allowing all incoming traffic on port '80' from '0.0.0.0/0'"
+-- Install the aws_security_group IaSQL module
+SELECT iasql_install('aws_security_group');
+-- "Create a security group named `allow-all` attached to the VPC with ID `vpc-f56cff91`"
+INSERT INTO security_group (description, group_name, vpc_id)
+VALUES ('allows all incoming traffic on port 80', 'allow-all', (SELECT id FROM vpc WHERE vpc_id = 'vpc-08c0a68025c2983a2'));
+-- "Then add a rule to it allowing all incoming traffic on port `80` from `0.0.0.0/0`."
+INSERT INTO security_group_rule (is_egress, ip_protocol, from_port, to_port, cidr_ipv4, description, security_group_id)
+VALUES (false, 'tcp', 80, 80, '0.0.0.0/8', 'allow all to 80', (SELECT id FROM security_group WHERE group_name = 'allow-all'));
+```
```
aguillenv, line 55: Also, maybe we can move the structure of this section to put the SQL examples closer to the point they're describing. Not a blocker, just a suggestion.

I mean, talk about high-level modules and show the example, then talk about the low-level module and show the example.

site/blog/2022-11-04.mdx
dfellis: Then I am completely confused by the example being in there, to be honest? It feels very off-topic from the "deploy my `Dockerfile`-managed git repo to AWS easily" subject if it is not used in the actual flow in any way, and could similarly confuse people into believing it is necessary.

If you want to go into the high-level/low-level distinctions in this blog post, perhaps it would make more sense explaining how `aws_ecs_simplified` depends on `aws_ecs_fargate`, `aws_ecr`, etc, and listing out all of the SQL statements it's `INSERT` trigger executes for you? That would showcase how *much* the simplified module does for you.

Then you could take it further, if you wanted to, showing how after you created the deployment via the simplified module you can still `UPDATE` some default that you have a special need to be different, getting the power to adjust things to your precise needs that a PaaS would instead force you to work around because it cannot be changed.

site/blog/ecs-simplified.mdx
```diff
@@ -0,0 +1,124 @@
...
+SELECT iasql_apply();
+
+SELECT ecr_build(
+  'https://github.com/iasql/iasql-engine/', -- the Github repo URL
+  (SELECT id
+    FROM repository
+    WHERE repository_name = 'simple-express-repository')::varchar(255), -- ECR repo for the image to be pushed
+  './examples/ecs-fargate/prisma/app', -- the subdirectory in Github repo
+  'main', -- the Github branch or ref
+  '' -- Github personal access token - can be omitted if public repository
```
dfellis, line 79: Now that it's the last argument, can we write `NULL` instead of an empty string, or does that still fail?

site/blog/ecs-simplified.mdx
```diff
@@ -0,0 +1,124 @@
...
+- In the next step, we'll automatically build an image for the code in Github repo and then push it to this URI (all through SQL and using another high-level function named `ecr_build`).
+```sql
+SELECT ecr_build(
+  'https://github.com/iasql/iasql-engine/', -- the Github repo URL
+  (SELECT id
+    FROM repository
+    WHERE repository_name = 'simple-express-repository')::varchar(255), -- ECR repo for the image to be pushed
+  './examples/ecs-fargate/prisma/app', -- the subdirectory in Github repo
+  'main',  -- the Github branch or ref
+  '' -- Github personal access token - can be omitted if public repository
```
dfellis, line 114: Same comment: can we make this `NULL` instead of an empty string? That would be more idiomatic in SQL, I think.

mtp1376: It does work. I've changed to `NULL`.

site/blog/ecs-simplified.mdx
mtp1376: It works with `NULL`. I've changed all occurrences to use `NULL` argument.

site/blog/ecs-simplified.mdx
```diff
@@ -125,4 +90,21 @@ SELECT load_balancer_dns
 FROM ecs_simplified
 WHERE app_name = 'simple-express';
 ```
-Then you can check if the server is running on the `url:8088/health` address.
\ No newline at end of file
+Then you can check if the server is running on the `url:8088/health` address.
+
+## Low-level Access to Resources
+
+So the `aws_ecs_simplified` module simplifies things, right? But what if you still want the level of control you had when you were doing all the steps manually? You can't grow your app beyond limitations if you don't have access to all the small details. Let's say you want your ECS container to be able to use the AWS CLI to provision an EC2 instance, and for that purpose its IAM role needs `AmazonEC2FullAccess` policy to work properly. `aws_ecs_simplifiy` is good, but wasn't it better if we could just manage the lower-level details as well?
```
dfellis, line 97: Nit: want -> need. Hopefully people who *want* to just spend time fiddling with stuff are not using a tool to reduce the amount of stuff they fiddle with. ;)

Also restructuring a little bit that's a bit awkward-reading to me. Feel free to disregard that point, though.

```suggestion
So the `aws_ecs_simplified` module simplifies things, right? But what if you still need the level of control you had when you were doing all the steps manually? The traditional PaaS trade-off is that you can't grow your app beyond limitations as you don't have access to all the small details. The IaSQL approach is not limited in that way.

Let's say you want your ECS container to be able to use the AWS CLI to provision an EC2 instance, and for that purpose its IAM role needs `AmazonEC2FullAccess` policy to work properly. `aws_ecs_simplified` does not have a column to configure such a thing, but that doesn't mean we're stuck.
```

site/docs/concepts/high-level-vs-low-level.md
```diff
@@ -13,6 +13,18 @@ The first statement, is described in a high-level abstraction layer. The higher
...
+
+Now comes in the `aws_ecs_simplified` high-level module. It takes the advantage of the CRUD operations handled by the `aws_security_group` module. So when you insert a new record into the `ecs_simplified` table (to create an ECS app with all the needed resources), that `INSERT` command [triggers](https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_ecs_simplified/sql/after_install.sql#L159-L165) a function that [inserts records for a new security group with two rules](https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.22/aws_ecs_simplified/sql/after_install.sql#L64-L73) into the tables managed by `aws_security_group`. So in the next `iasql_apply()`, the `aws_security_group` module will provision that security group and add those two rules to it.
```
dfellis, line 22: The only nit is that all of these links to source code should be changed to the `v0.0.22` tag instead of `main` once `v0.0.22` is released to prevent link-rot.

dfellis (approved):

site/blog/ecs-simplified.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+In this post, we're going to discover an IaSQL module that's built to make deploying to ECS, simplified. Most of the details for deploying a container to ECS are the same (load balancers, security groups, IAM roles, etc), and we have created the `ecs_simplified` module for you so that you can give it any Github repo with a `Dockerfile` and get your app deployed to ECS in the fastest time possible, with scalability available! All the needed resources are going to be created automatically in your AWS account, and you'll have the full access to the details while you're gaining the benefit of a higher-level simple deployment.
+
+## ** An Example of `aws_ecs_simplified` Module **
+
+Let's say we are going to deploy [this simple Express.js app](https://github.com/iasql/iasql-engine/tree/main/examples/ecs-fargate/prisma/app) to the ECS. It has a `Dockerfile` ready which `npm install`s the `express` package and then starts the Express server which listens on the port `8088`.
```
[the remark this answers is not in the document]
dfellis: The URL changes to reference `v0.0.22` is to make sure this blog post always points at valid URLs.

site/blog/ecs-simplified.mdx
```diff
@@ -0,0 +1,112 @@
...
+  'https://github.com/iasql/iasql-engine/', -- the Github repo URL
+  (SELECT id
+    FROM repository
+    WHERE repository_name = 'simple-express-repository')::varchar(255), -- ECR repo for the image to be pushed
+  './examples/ecs-fargate/prisma/app', -- the subdirectory in Github repo
+  'main',  -- the Github branch or ref
+  NULL -- Github personal access token - can be omitted if public repository
+);
+```
+- This command tells IaSQL to go clone `iasql-engine` repository, build an image on the subdirectory specified, and then push it to the ECR repository created earlier by `ecs_simplified` module. Running the above command will automatically create a CodeBuild project and the related roles, etc. Then it'll start a build, and after it's successful all the created resources are deleted to ensure there won't be any additional charges to your AWS account.
```
dfellis, line 86:
```suggestion
- This command tells IaSQL to go clone `iasql-engine` repository, build an image on the subdirectory specified, and then push it to the ECR repository created earlier by the `aws_ecs_simplified` module. Running the above command will automatically create a CodeBuild project and the related roles, etc. Then it'll start a build, and after it's successful all the created resources are deleted to ensure there won't be any additional charges to your AWS account.
```

site/blog/ecs-simplified.mdx
```diff
@@ -0,0 +1,112 @@
...
+- To access your app on the cloud, get the load balancer address and use your browser to access the live version of it:
+```sql
+SELECT load_balancer_dns
+FROM ecs_simplified
+WHERE app_name = 'simple-express';
+```
+Then you can check if the server is running on the `url:8088/health` address.
```
dfellis, line 93: Might be slightly clearer as:

```suggestion
Then you can check if the server is running on the `<load_balancer_dns value>:8088/health` address.
```

site/blog/ecs-simplified.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+## Low-level Access to Resources
+
+So the `aws_ecs_simplified` module simplifies things, right? But what if you still need the level of control you had when you were doing all the steps manually? The traditional PaaS trade-off is that you can't grow your app beyond limitations as you don't have access to all the small details. The IaSQL approach is not limited in that way.
```
dfellis, line 97:
```suggestion
So the `aws_ecs_simplified` module simplifies things, right? But what if you still need the level of control you had when you were doing all the steps manually? The traditional PaaS trade-off is that you can't grow your app beyond the built-in limitations as you don't have access to all the small details. The IaSQL approach is not limited in that way.
```

site/blog/ecs-simplified.mdx
```diff
@@ -0,0 +1,112 @@
...
+The good news is that you still have the full control over all resources in the deepest details. Let's fix your app's IAM role access by attaching the needed policy to its IAM role:
+```sql
+UPDATE iam_role
+SET attached_policies_arns = attached_policies_arns ||
+                             'arn:aws:iam::aws:policy/AmazonEC2FullAccess' -- attached_policies_arns is of text[] type
+WHERE role_name = 'simple-express-ecs-task-exec-role';
+
+```
+
+You want additional rules for the container's security group? No problem! just write the SQL and execute it, and it will be applied to the cloud within seconds. You want 3 copies of your container to be kept running with a round-robin load balancing on them? It's already there, just do an `UPDATE ecs_simplified SET desired_count = 3 WHERE app_name = 'simple_express';` and it's there for you.
```
dfellis, line 110:
```suggestion
You want additional rules for the container's security group? No problem! Just write the SQL and execute it, and it will be applied to the cloud within seconds. You want 3 copies of your container to be kept running with a round-robin load balancing on them? It's already there, just do an `UPDATE ecs_simplified SET desired_count = 3 WHERE app_name = 'simple_express';` and it's there for you.
```

dfellis: Also, maybe this instead:

```suggestion
You want additional rules for the container's security group? No problem! Just write the SQL, execute it, and `SELECT iasql_apply()` it and it will be applied to the cloud within seconds. You want 3 copies of your container to be kept running with a round-robin load balancing on them? It's already there, just do an `UPDATE ecs_simplified SET desired_count = 3 WHERE app_name = 'simple_express';`, `SELECT iasql_apply()` it and it's there for you.
```

site/blog/ecs-simplified.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+I'm not going to continue this long list, since I've already got a headache. Doing those steps manually is going to give you a headache as well, so why bother doing all those steps yourself and risking different errors you might face when deploying your containerized app? You don't really have the time for the random IAM-related errors AWS is demanding you to resolve. Besides, you already have your codebase ready and the `Dockerfile` is there, so why not just run a simple command doing something that should be simply done?
+
+In this post, we're going to discover an IaSQL module that's built to make deploying to ECS, simplified. Most of the details for deploying a container to ECS are the same (load balancers, security groups, IAM roles, etc), and we have created the `ecs_simplified` module for you so that you can give it any Github repo with a `Dockerfile` and get your app deployed to ECS in the fastest time possible, with scalability available! All the needed resources are going to be created automatically in your AWS account, and you'll have the full access to the details while you're gaining the benefit of a higher-level simple deployment.
```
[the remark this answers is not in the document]
depombo: technically it is called `ecs_simplified` at the moment, but we should probably rename it?

site/blog/ecs-simplified.mdx
```diff
@@ -52,7 +52,7 @@ That's it! Now wait for some time and your app is deployed! While your app is be
 ```sql
 SELECT iasql_install('aws_ecs_simplified');
 ```
-- This command installs the `aws_ecs_simplified` high-level module. We– at IaSQL– have created that module to make it easy to deploy containerized apps to ECS. The code for it is [here](https://github.com/iasql/iasql-engine/blob/main/src/modules/0.0.23/aws_ecs_simplified/sql/after_install.sql). But IaSQL is so flexible that anyone can create their own high-level (and of course, low-level) modules and add it to IaSQL.
+- This command installs the `aws_ecs_simplified` high-level module. We– at IaSQL– have created that module to make it easy to deploy containerized apps to ECS. The code for it is [here](https://github.com/iasql/iasql-engine/blob/v0.0.22/src/modules/0.0.23/aws_ecs_simplified/sql/after_install.sql). But IaSQL is so flexible that anyone can create their own high-level (and of course, low-level) modules and add it to IaSQL.
```
dfellis, line 55: This URL is broken. You need the `0.0.22` in both parts for it to work.

site/docs/concepts/high-level-vs-low-level.md
```diff
@@ -0,0 +1,70 @@
...
+- `ecr_build` is an RPC function under the `aws_ecr` module, and [it's written in Node.js](https://github.com/iasql/iasql-engine/blob/v0.0.22/src/modules/0.0.22/aws_ecr/rpcs/build.ts#L92). Its relation with the low-level modules (such as `aws_iam`) can't be altered, and its code cannot reside anywhere outside the main `iasql-engine` repository.
+
+So, if you're doing something that can be done using the pure-SQL commands working with the low-level modules in IaSQL, your code can reside in your own Github repository. But if you're doing something that can't be done just by using the building blocks available, you can create a new one yourself. Just add the RPC, or the module to the `iasql-engine` codebase, and submit a PR. Reading our [contribution guide](https://github.com/iasql/iasql-engine/blob/v0.0.22/CONTRIBUTING.md) can help you and if you have questions regarding the development or the usage of our system, feel free to ask it in our [Discord channel](https://discord.com/invite/machGGczea).
```
dfellis, line 70:
````suggestion
# Low-Level vs High-Level Modules

[Modules in IaSQL](./module) allow one to manipulate tables and call database functions that impact the state of their cloud infrastructure. Some of these modules may be described as "low level" and others as "high level", but rather than (just) referring to the features available and the difficulty of using them, these terms have a specific meaning within the IaSQL context.

Consider the following stack diagram:

```
     O
    -|-                      User
    / \
     |
     |______________
     |              |
     v              |
   ____________     |
  | High-level |    |
  |  Modules   |    |        IaSQL
  |____________|____V_____
  | Low-level Modules     |
  |_______________________|
              |
              |
              |
              v
             __
         ___/  \__
        /  Cloud  \              Infrastructure
        \_    _   /
          \__/ \_/
```

The user can use both high-level and low-level modules to manipulate their cloud infrastructure, but the high-level modules are built on top of the low-level modules, and only the low-level modules *actually* interact with their cloud infrastructure.

The high-level modules operate on top of the abstraction layer the low-level modules provide: SQL tables and functions, therefore the high-level modules *are pure SQL*.

Therefore, they cannot, on their own, perform arbitrary code execution, but can only perform operations that can be glued together by the low-level modules.

The `aws_ecs_simplified` module is one such example, with the vast majority of code [being pure SQL code](https://github.com/iasql/iasql-engine/blob/v0.0.22/src/modules/0.0.22/aws_ecs_simplified/sql/after_install.sql) defining a new table and trigger functions to mutate the low-level module tables that it abstracts away for you.

For more information on just how nice it is to use a high-level module like `aws_ecs_simplified`, check out [this blog post on "ECS, Simplified"](https://iasql.com/blog/ecs-simplified).

As pure SQL, the barrier to entry in authoring a high-level module is lower, so anyone with the desire can write a high-level module and share it with others, and we hope to make such third-party module installation much simpler soon.

The low-level modules are tables and functions that are created by and interface with the IaSQL engine, where module code is written to execute the necessary cloud API calls, such as [this part of the `aws_ecs_fargate` module for the `cluster` table](https://github.com/iasql/iasql-engine/blob/v0.0.22/src/modules/0.0.22/aws_ecs_fargate/mappers/cluster.ts).

These low-level modules convert your database changes into API calls, and need to be [thoroughly tested](https://github.com/iasql/iasql-engine/blob/v0.0.22/test/modules/aws-ecs-integration.ts) for correctness and vetted for trustworthiness that they won't utilize your cloud spending account for nefarious purposes.

For that reason, they *must* be committed into the [IaSQL engine repository](https://github.com/iasql/iasql-engine) before they can be installed or used, and no third-party low-level modules are allowed.

This makes Low-Level modules in IaSQL similar to the built-in functions and libraries of programming languages: something that you'll often use in your day-to-day, but that most will never contribute to, and that's okay.

As a slight wrinkle, there are also Low-Level and High-Level Postgres Functions in IaSQL. There is a similar distinction between them, where Low-Level Functions are directly powered by the IaSQL Engine, while High-Level Functions are pure SQL built on top of other functions, tables, etc. However, some Low-Level Functions can be quite "high level" in user experience.

An "obvious" kind of low-level function comes from the `aws_s3` module, where [the `s3_upload_object` function](https://github.com/iasql/iasql-engine/blob/v0.0.22/src/modules/0.0.22/aws_s3/rpcs/s3_upload_object.ts) allows users to upload new "objects" (files) to an S3 bucket through a SQL statement.

That provides a clear, composable function that you could use in ways never envisioned by the author of the function, such as uploading the results of a SQL query about your infrastructure into an S3 file to be consumed by an internal dashboard.

In contrast, the `aws_ecr` module defines the [`ecr_build` function](https://github.com/iasql/iasql-engine/blob/v0.0.22/src/modules/0.0.22/aws_ecr/rpcs/build.ts) that allows one to very simply build any public Git repo (or private Github repo) with a `Dockerfile` defined and store the results in your ECR repository of choice. It is a low-level module but provides an abstracted interface (over a couple of different AWS services) in a "high-level" way. It's purpose and arguments being so particular to the task it is designed for it is unlikely to ever be used in a way the author did not envision.

As an example high-level function, we can look at the low-level `aws_account` module (yes, you can define high-level functions in low-level modules, but not vice-versa), where the [`default_aws_account` function](https://github.com/iasql/iasql-engine/blob/v0.0.22/src/modules/0.0.22/aws_account/sql/after_install.sql#L1-L8) has been defined.

It is both a convenience function for users to alter their default region without running into a constraint on the `aws_regions` table with multiple regions marked as default, but also as the default value to insert into the `region` column in multiple tables across many low-level modules if the user does not specify which region the resource belongs to. The high-level function simply builds on top of the functionality the low-level `aws_regions` table provides.

For most users, the high-level vs low-level divide between modules and functions is unimportant and totally opaque to them, and the "high-level vs low-level" distinction they need is simply the trade-off between configurability and ease-of-use. *In general* this will line up with the actual technical distinction, as any cloud functionality *not* exposed by the low-level modules can never be recovered with the high-level modules, so the low-level modules *must* be as configurable and detailed as possible, while the high-level modules can cut whatever corners they like to provide a more tailored experience.

But that does not excuse accidental, needless complexity, and so great care is taken on the design of [the database schema these modules produce](https://iasql.com/schema/). How would someone write a high-level module if they don't understand the low-level modules in the first place?
````

PR #1528, opened 2022-11-09
optional gh_pat from @mtp1376

depombo:


dfellis (approved):

dfellis:
I pulled the logs from the test account before they are wiped out

Screenshot from 2022-11-06 19-00-23

It's not authenticating. :/ @mtp1376 can you take a look? It's possible something was changed when we cherry-picked it, but if not, how did you get it to pass locally?

mtp1376:
> I pulled the logs from the test account before they are wiped out
> 
> Screenshot from 2022-11-08 20-46-13
> 
> It's not authenticating. :/ @mtp1376 can you take a look? It's possible something was changed when we cherry-picked it, but if not, how did you get it to pass locally?

I think I've found the issue. When we pass the `NULL` value, the RPC engine seems to ignore that value and use the next value (in this case `main`) as the `githubPersonalAccessToken`. And the unexpected stuff happens.

aguillenv:
> > I pulled the logs from the test account before they are wiped out
> > Screenshot from 2022-11-08 20-46-13
> > It's not authenticating. :/ @mtp1376 can you take a look? It's possible something was changed when we cherry-picked it, but if not, how did you get it to pass locally?
> 
> I think I've found the issue. When we pass the `NULL` value, the RPC engine seems to ignore that value and use the next value (in this case `main`) as the `githubPersonalAccessToken`. And the unexpected stuff happens.

Should we swap the `githubRef` and the `githubPersonalAccessToken` arguments? Will be more probable to want to pass the ref rather than the ghPAT and we would avoid that PAT as `NULL` issue?

aguillenv (approved):

PR #1532, opened 2022-11-09
feat: adds vpc support to lambda functions

yrobla:
Closes: #1515


dfellis:
This error looks legit https://github.com/iasql/iasql-engine/actions/runs/3473294268/jobs/5811531172#step:7:25011

src/modules/0.0.23/aws_lambda/mappers/lambda_function.ts
```diff
@@ -230,6 +336,27 @@ export class LambdaFunctionMapper extends MapperBase<LambdaFunction> {
     delete: async (es: LambdaFunction[], ctx: Context) => {
       for (const e of es) {
         const client = (await ctx.getAwsClient(e.region)) as AWS;
+        const region = e.region;
+
+        // Update function configuration
```
aguillenv, line 341: Maybe expand this comment that we are doing the update to avoid network interfaces get locked by aws.

src/modules/0.0.23/aws_lambda/entity/lambda_function.ts
```diff
@@ -130,6 +141,15 @@ export class LambdaFunction {
   })
   tags?: { [key: string]: string };
 
+  @Column('varchar', { array: true, nullable: true })
+  subnets?: string[];
```
aguillenv, line 145: in other tests where we add subnets like this, we also add an `after_install.sql` and `before_uninstall.sql` to add/drop a constraint to make sure all the string inserted as a subnet actually belongs to the subnet table.

test/modules/aws-lambda-integration.ts
```diff
@@ -673,6 +673,12 @@ describe('Lambda Integration Testing', () => {
 
     DELETE FROM security_group WHERE group_name = '${prefix}lambdanotdefault' AND region='${region}';
 
+    DELETE FROM security_group_rule
+    WHERE security_group_id = (
+      SELECT id
+      FROM security_group
+      WHERE security_group.vpc_id=vpc.id);
```
aguillenv, line 681: I think you do not have vpc.id here available?

dfellis (approved):

dfellis (approved):

PR #1533, opened 2022-11-09
Test and fix full bucket removal

dfellis:


depombo (approved):

src/modules/0.0.22/aws_s3/mappers/bucket_object.ts
```diff
@@ -34,10 +34,11 @@ export class BucketObjectMapper extends MapperBase<BucketObject> {
     return res;
   };
 
-  getBucketObjects = paginateBuilder<S3>(paginateListObjectsV2, 'Contents', undefined, undefined, (b, k) => ({
-    Bucket: b,
-    Prefix: k,
-  }));
+  getBucketObjects = paginateBuilder<S3>(paginateListObjectsV2, 'Contents', undefined, undefined, (b, k) => {
+    const out: any = { Bucket: b };
+    if (k) out.Prefix = k;
```
mtp1376, line 39: Just curious what `k` was here. Currently the `out.Prefix` would be `undefined`. So it was something like `''` or `false`? :D

mtp1376 (approved):
LGTM.

src/modules/0.0.22/aws_s3/mappers/bucket_object.ts
dfellis: There's unfortunately some *weirdness* with Javascript and `undefined`.

Screenshot from 2022-11-09 17-34-46

Most of the time, this distinction doesn't matter, but tools that do meta-work on the data structures themselves, like the underlying JSON-to-XML stuff that the AWS SDK does, may end up producing weird XML output like casting an `undefined` into an empty string, which could have follow-on results.

I made that change out of an abundance of caution. Not sure if it actually was also a secondary cause of listing failures or not, but I don't see a reason to undo it, personally. ;)

mtp1376: Got it. Thanks for that tip.

PR #1536, opened 2022-11-10
Fix secrets manager to allow the same secret name to exist in multiple regions

dfellis:


aguillenv (approved):

test/modules/aws-secret-integration.ts
```diff
@@ -158,8 +158,19 @@ describe("Secrets Manager Integration Testing", () => {
     expect(res[0].region).toBe('us-east-1');
   }));
 
+  it('creates the same secret back in the original region at the same time', query(`
+    INSERT INTO secret (name, description, value)
+    VALUES ('${secretName}', 'description', '${secretValue}');
+  `));
```
dfellis, line 164: I forgot to add an extra `apply` here, just a sec.

PR #1538, opened 2022-11-10
Add Coverage for VpcPeeringConnection

mtp1376:


mtp1376:
A problem I need to solve is that when a DB record is deleted, the `cloud.delete` will be called and therefore a delete request will be sent to AWS to delete that PeeringConnection. But it doesn't actually get deleted and its `state` changes to `deleted`. So in the next `cloud.read` it's still there and the engine keeps deleting it.
Maybe I should not include the items that have `state=deleted` in the result of `cloud.read`. EC2 has the same functionality so maybe looking at the `aws_ec2` module's code can help.

mtp1376:
> A problem I need to solve is that when a DB record is deleted, the `cloud.delete` will be called and therefore a delete request will be sent to AWS to delete that PeeringConnection. But it doesn't actually get deleted and its `state` changes to `deleted`. So in the next `cloud.read` it's still there and the engine keeps deleting it. Maybe I should not include the items that have `state=deleted` in the result of `cloud.read`. EC2 has the same functionality so maybe looking at the `aws_ec2` module's code can help.

Yes the EC2 module is not including terminated ones in the result.
image

mtp1376:
- Current version only supports the cases that two VPCs are within the connected account. Should we add the support for the case that a VPC from other AWS account is to be peered? I initially thought it's a rare case so I didn't go for it.
- Regarding the routes and route tables: since we don't have that entity defined and the database tables are not available for easy management of those, we have two possible approaches:
  - Add the `RouteTable` and `Route` entity and create the mappers for it: great for long-term, but all the other usages should be refactored to use the new approach
  - Just manually add the routes for the two VPCs to be able to see each other by invoking the AWS calls directly.

mtp1376:
There are some complications:
1. VPCs can have multiple CIDRs. Currently we just support single-CIDR VPCs and if there's more than one CIDR for a VPC, we just keep the first one in the database.
2. When we create a peering connection, we should add routes to *every* subnet's route table in those two VPCs in order for them to be able to connect to each other. 
   - Let's say they each have 2 public subnets and 2 private ones. We will need to insert:
   - VPC A - subnet 1 route table: send traffic to VPC B's CIDR through `pcx-xxx` peering connection
   - VPC B - subnet 1 route table: send traffic to VPC A's CIDR through `pcx-xxx` peering connection
   - ... (repeat the steps above for each `subnet * CIDR` for VPC A and B (`|subnet| * |CIDR| * |VPCs|` cardinal multiplication).

mtp1376:
Interface Endpoints have some differences with Gateway Endpoints which affects our work here:
- Interface Endpoints use ENIs, SecurityGroups and Subnets to work.
   - image
- Gateway Endpoints should be configured using RouteTable.

mtp1376:
So the code should be changed to do these as well:
- For each subnet `s` in VPC A:
   - `rt = get_route_table_for_subnet(s)` (explicitly associated RouteTable, otherwise VPC's main RouteTable)
   - `rt.add(VPC_B.CIDR, through='pcx-xxx`)
Same for VPC B

src/modules/0.0.23/aws_vpc/entity/peering_connection.ts
```diff
@@ -0,0 +1,51 @@
...
+  EXPIRED = 'expired',
+  FAILED = 'failed',
+  INITIATING_REQUEST = 'initiating-request',
+  PENDING_ACCEPTANCE = 'pending-acceptance',
+  PROVISIONING = 'provisioning',
+  REJECTED = 'rejected',
+}
+
+@Entity()
+export class PeeringConnection {
```
aguillenv, line 17: Are peerings region specific or global?

mtp1376: They are region specific, but I thought keeping an extra `region` column would be redundant since `requester.region` and `accepter.region` for the VPCs are available. What's your idea?

mtp1376: Actually the 2 VPCs **can** be from two different regions.

aguillenv: oh, nevermind then. So the peering has to be in the requester one? or wherever?

mtp1376: The peering creation should be on the `requester` region, but the accept call should be originating from the `accepter` region.

src/modules/aws_vpc/mappers/peering_connection.ts
```diff
@@ -0,0 +1,261 @@
...
+      // accepter or requester vpc no longer exist
+      return undefined;
+    }
+    out.state = peeringConnection.Status?.Code as PeeringConnectionState;
+    out.tags = this.convertTagsFromAws(peeringConnection.Tags);
+
+    return out;
+  }
+
+  private convertTagsFromAws(inputTags?: Tag[]) {
```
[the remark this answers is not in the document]
mtp1376: We can have it as a follow-up PR.

src/modules/aws_vpc/entity/peering_connection.ts
```diff
@@ -0,0 +1,51 @@
+import { Column, Entity, JoinColumn, ManyToOne, PrimaryGeneratedColumn } from 'typeorm';
+
+import { cloudId } from '../../../services/cloud-id';
+import { Vpc } from './vpc';
+
+export enum PeeringConnectionState {
```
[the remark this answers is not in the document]
mtp1376: Sure. I'll add them now.

depombo (approved):

src/modules/aws_vpc/mappers/peering_connection.ts
```diff
@@ -0,0 +1,290 @@
...
+        }),
+      );
+
+      if (!!id) {
+        return out.find(pc => pc.peeringConnectionId === id);
+      }
+
+      return out;
+    },
+    update: async (es: PeeringConnection[], ctx: Context) => {
```
aguillenv, line 215: Would be good to add here the `updateOrReplace` method and return `replace` when the vpcIds are different.

PR #1541, opened 2022-11-11
Create the group role only if it doesn't yet exist, which it will for databases that are upgraded

dfellis:


dfellis:
Screenshot from 2022-11-10 18-44-04

With this diff added on I was able to get the staging database upgraded, so confirmation that it works.

mtp1376 (approved):
What was the problem with dropping the role if it existed? Just asking to gain a better insight on the postgres stuff.

dfellis:
> What was the problem with dropping the role if it existed?

Other accounts also associated with that role would lose access to it, which would negatively impact the main account we set up and forget the password for, leaving it only to the user.

PR #1543, opened 2022-11-12
Singular Docker container

dfellis:
Part of the work for better scalability.


dfellis:
Ready for review, but not going to land it, yet.

Dockerfile
```diff
@@ -1,19 +1,37 @@
-FROM node:16-bullseye
+FROM debian:bullseye
```
aguillenv, line 1: interesting, any reason for this change?

package.json
```diff
@@ -19,8 +19,9 @@
     "gen-sql": "IASQL_ENV=ci ts-node ./node_modules/.bin/typeorm migration:generate -n",
     "migrate": "IASQL_ENV=ci ts-node ./node_modules/.bin/typeorm migration:run",
     "gen-module": "IASQL_ENV=ci ./.module_migration.sh",
-    "docker-compose:debug": "IASQL_ENV=ci docker-compose -f docker-compose.yml -f docker-compose.debug.yml up --build",
-    "docker-compose": "IASQL_ENV=local docker-compose up --build",
+    "docker-build": "docker build -t iasql:latest .",
+    "docker-run": "docker run -e IASQL_ENV=local -p 5432:5432 -p 8088:8088 --name iasql iasql:latest",
+    "docker-run-vol": "docker run -e IASQL_ENV=local -p 5432:5432 -p 8088:8088 -v $(dirname $PWD)/db:/var/lib/postgresql/13/main --name iasql iasql:latest",
```
aguillenv, line 24: for persistence?

Dockerfile
dfellis: Well, we need Node and Postgres and pgbouncer all in the same container, and while the Node image was based on the same version of Debian as this, it ripped out things from Debian that Node does not need.

I figured going from a more "standard" base that I can look up problems online and be more sure they'll actually apply would reduce the total workload and maintenance burden longer-term.

.github/workflows/ci.yml
```diff
@@ -209,38 +211,38 @@ jobs:
 
           # Setup Account
           echo "\nInstalling aws_account..."
-          psql postgres://postgres:test@localhost:5432/iasql -c "
+          psql "postgres://postgres:test@localhost:5432/iasql?ssl=true&sslmode=require" -c "
```
yrobla, line 214: maybe you can just have a var for postgres_connection_string, and reuse in all the following calls?

.github/workflows/django.yml
```diff
@@ -72,9 +74,9 @@ jobs:
           export AWS_ACCESS_KEY_ID=$(echo "${ACCESS_KEY_IDS}" | jq -r ".[${ACCOUNT_INDEX}]")
           export AWS_SECRET_ACCESS_KEY=$(echo "${SECRET_ACCESS_KEYS}" | jq -r ".[${ACCOUNT_INDEX}]")
           echo "\nDelete all records..."
-          psql postgres://postgres:test@localhost:5432/iasql -c "select * from delete_all_records();"
+          psql "postgres://postgres:test@localhost:5432/iasql?ssl=true&sslmode=require" -c "select * from delete_all_records();"
```
yrobla, line 77: same comment about the reused var

.github/workflows/prisma.yml
```diff
@@ -88,9 +90,9 @@ jobs:
           export AWS_ACCESS_KEY_ID=$(echo "${ACCESS_KEY_IDS}" | jq -r ".[${ACCOUNT_INDEX}]")
           export AWS_SECRET_ACCESS_KEY=$(echo "${SECRET_ACCESS_KEYS}" | jq -r ".[${ACCOUNT_INDEX}]")
           echo "\nDelete all records..."
-          psql postgres://postgres:test@localhost:5432/iasql -c "select * from delete_all_records();"
+          psql "postgres://postgres:test@localhost:5432/iasql?ssl=true&sslmode=require" -c "select * from delete_all_records();"
```
yrobla, line 93: same

test/upgrade/run-upgrade.sh
```diff
@@ -4,7 +4,7 @@
 set -vex
 
 # Make sure the test will work at all
-if [ ! -f "${PWD}/docker-compose.yml" ]; then
+if [ ! -f "${PWD}/CONTRIBUTING.md" ]; then
```
yrobla, line 7: why this change?

.github/workflows/ci.yml
dfellis: Sounds good

.github/workflows/django.yml
dfellis: 👍

.github/workflows/prisma.yml
dfellis: 👍

test/upgrade/run-upgrade.sh
dfellis: The `docker-compose.yml` file has been deleted in this branch, so I decided to go with a file that is highly unlikely to ever be deleted and can indicate that we are in the root of the repo.

test/upgrade/run-upgrade.sh
```diff
@@ -14,22 +14,37 @@ LATESTVERSION=$(./node_modules/.bin/ts-node src/scripts/latestVersion.ts)
...
 i=0
 while [ $i -lt $(($len - 1)) ]; do
+  # Clear mutations, if any
+  git reset --hard HEAD
   echo "Upgrading from version ${availableVersions[$i]} to ${LATESTVERSION}"
   # Check out the older version of the codebase and launch the engine with a local postgres
   git checkout v${availableVersions[$i]}
-  yarn docker-compose &
+  mkdir -p ${PWD}/../db
+  if [ -f "${PWD}/docker-compose.yml" ]; then
```
aguillenv, line 30: should this be changed with the CONTRIBUTING.md?

dfellis: No, here we are figuring out if the repo has been switched back to an earlier version that still uses Docker compose or not.

src/scripts/postgresql.conf
```diff
@@ -0,0 +1,799 @@
+# -----------------------------
```
aguillenv, line 1: once you rebase with the `pg_cron` change we will need to add in this file 2 lines:
```
shared_preload_libraries=pg_cron
cron.database_name=iasql_metadata
```

Dockerfile
```diff
@@ -1,19 +1,37 @@
...
+# RUN ["bash", "-c", "echo 'deb http://apt.postgresql.org/pub/repos/apt/ bullseye-pgdg main' > /etc/apt/sources.list.d/postgresql.list"]
+# RUN apt update
+# RUN apt upgrade -y
+# RUN apt install postgresql-client-14 postgresql-14 pgbouncer -y
+# COPY ./src/scripts/postgresql.conf /etc/postgresql/14/main/postgresql.conf
+# COPY ./src/scripts/pg_hba.conf /etc/postgresql/14/main/pg_hba.conf
+
+RUN apt upgrade -y
+# From https://unix.stackexchange.com/a/366722
+RUN apt install postgresql-client-13 postgresql-13 pgbouncer locales-all -y
```
aguillenv, line 24: here we would need to install `pg_cron` too, I was doing it with:
`RUN apt-get update && apt-get -y install postgresql-13-cron`

Maybe with this change would work:
```suggestion
RUN apt install postgresql-client-13 postgresql-13 postgresql-13-cron pgbouncer locales-all -y
```

aguillenv (approved):

Dockerfile
dfellis: Done :)

src/scripts/postgresql.conf
dfellis: Done :)

src/scripts/setup-iasql-db.sh
```diff
@@ -8,36 +8,37 @@ echo "\nCreating an iasql db..."
 connectres=$(curl http://localhost:8088/v1/db/connect/iasql)
 export IASQL_USERNAME=$(jq -r '.user' <<<"$connectres")
 export IASQL_PASSWORD=$(jq -r '.password' <<<"$connectres")
+echo $connectres
```
yrobla, line 11: why is that needed?

src/scripts/setup-iasql-db.sh
```diff
@@ -8,36 +8,37 @@ echo "\nCreating an iasql db..."
...
 
 # Setup Account
 echo "\nInstalling aws_account..."
```
yrobla, line 14: you could have some connection string and reuse it

yrobla (commented):
some comments

Dockerfile
mtp1376: Why do we need Node/Postgres/pgbouncer all in the same container? Can't a `docker-compose` do the trick?

mtp1376 (approved):

src/scripts/setup-iasql-db.sh
dfellis: This was debug logging on my part. Going to remove.

Dockerfile
dfellis: @mtp1376 that was discussed in [this RFC](https://github.com/iasql/iasql-engine/blob/main/rfcs/005%20-%20Unsurprising%20Functions%20and%20Scalability%20RFC.md#final-proposal) where this is the first PR to begin implementing it.

Basically by tightly coupling the engine, postgres, and pgbouncer, we can eliminate the unexpected behavior of multiple statements involving IaSQL RPC functions not having the mutated state available to follow-up statements *and* we can resolve scalability issues by letting the engine use more than one core on a machine, and eventually partition users between different machines running this setup (since no one user database should consume an entire machine, but also no user database has any interaction with any other user database so they don't need to be in the same database server).

yrobla (approved):

PR #1552, opened 2022-11-15
Make 'aws_s3' uninstallable

dfellis:
I discovered this accidentally, but we forgot to define the down path for the custom triggers in the `aws_s3` module, so uninstalling and re-installing that module might cause issues at the moment.

### DIFF
--- a/src/modules/0.0.22/aws_s3/sql/before_uninstall.sql
+++ b/src/modules/0.0.22/aws_s3/sql/before_uninstall.sql
@@ -1,0 +1,11 @@
+DROP TRIGGER
+  block_s3_object_primary_key_update;
+
+DROP FUNCTION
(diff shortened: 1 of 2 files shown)


aguillenv (approved):

dfellis:
> we should maybe add a test if these hooks are missing?

Because they're hand written it's hard to have a generalized test that wouldn't probably bite us in the future. But maybe we can do something confirming on install/uninstall of any module the schema is the same before and after?

depombo:
> Because they're hand written it's hard to have a generalized test that wouldn't probably bite us in the future. But maybe we can do something confirming on install/uninstall of any module the schema is the same before and after?

yes, something that checks the schema https://database.guide/5-ways-to-check-if-a-table-exists-in-postgresql/

PR #1557, opened 2022-11-16
[blog] launch stable diffusion

depombo:
a few road bumps that prevent a fully programmatic creation thus far:
- the ami id used by the main tutorials is part of AWS marketplace and the terms need to be agreed upon in the aws console https://aws.amazon.com/marketplace/server/procurement?productId=e6724620-3ffb-4cc9-9690-c310d8e794ef
- if no ec2 instances of significant size have been spin up thus far, the apply loop immediately hits an AWS error limit from exceeding the cpu quota and there needs to be a claim to increase the quota https://support.console.aws.amazon.com/support/home#/case/create?issueType=service-limit-increase&limitType=service-code-ec2-instances&serviceLimitIncreaseType=ec2-instances&type=service_limit_increase
- hugging face requires creating an account and providing the credentials to be able to download the model https://huggingface.co/CompVis/stable-diffusion-v1-4?text=A+high+tech+solarpunk+utopia+in+the+Amazon+rainforest

these can be addressed later or just included in the tutorial

### DIFF
--- a/site/blog/unlisted/stable-diffusion.mdx
+++ b/site/blog/unlisted/stable-diffusion.mdx
@@ -1,0 +1,55 @@
+---
+slug: stable-diffusion
+title: Deploy Stable Diffusion to your AWS account with one click
+date: 2022-11-15
+authors: [depombo]
+---
+
+# Deploy Stable Diffusion to an EC2 instance in your AWS account with one click
+
+:::caution
+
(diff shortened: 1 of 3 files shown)


depombo:
replaced by https://github.com/iasql/iasql/pull/2095

PR #1559, opened 2022-11-16
Add Coverage For `RouteTable`, `Route`, and `RouteTableAssociation"

mtp1376:


src/modules/0.0.23/aws_vpc/entity/route.ts
```diff
@@ -0,0 +1,74 @@
...
+  NetworkInterfaceId?: string;
+
+  @Column({ nullable: true })
+  @cloudId
+  VpcPeeringConnectionId?: string;
+
+  @Column({ nullable: true })
+  @cloudId
+  CoreNetworkArn?: string;
+}
```
dfellis, line 74: It may make sense to drop `@cloudId` on this particular entity and define your own custom `entityId` and `generateId` methods. That way you can use the `routeTable.routeTableId` as part of the unique key.

src/modules/0.0.23/aws_vpc/mappers/route_table.ts
```diff
@@ -0,0 +1,258 @@
...
+
+        const client = (await ctx.getAwsClient(e.region)) as AWS;
+        if (!this.eqListItems<Route>(this.eqRoute, e.routes, cloudRecord.routes)) {
+          // delete and create routes
+          const { entitiesInDbOnly, entitiesInAwsOnly } = this.findRoutesDiff(e.routes, cloudRecord.routes);
+          await Promise.all([
+            ...entitiesInAwsOnly.map(async r => await RouteTableMapper.deleteRoute(client.ec2client, e, r)),
+            ...entitiesInDbOnly.map(async r => await RouteTableMapper.createRoute(client.ec2client, e, r)),
+          ]);
+        }
```
dfellis, line 204: You may want to do this with two `for` loops instead of a `Promise.all`. The `Promise.all` will trigger the promises near-simultaneously and will be more vulnerable to rate limiting checks for larger tables.

src/modules/0.0.23/aws_vpc/mappers/route_table.ts
```diff
@@ -0,0 +1,258 @@
...
+
+        return out;
+      }
+    },
+    delete: async (es: RouteTable[], ctx: Context) => {
+      for (const e of es) {
+        const client = (await ctx.getAwsClient(e.region)) as AWS;
+        await client.ec2client.deleteRouteTable({ RouteTableId: e.routeTableId });
```
dfellis, line 212: Do you not need to delete any left-over routes in the route table?

mtp1376: Good point. I'll check the current behavior and whether the deletion cascades.

src/modules/0.0.23/aws_vpc/entity/route.ts
aguillenv: +1

src/modules/0.0.23/aws_vpc/mappers/route_table.ts
```diff
@@ -0,0 +1,278 @@
...
+      this.eqListItems<Route>(this.eqRoute, a.routes, b.routes) &&
+      eqTags(a.tags, b.tags)
+    );
+  };
+
+  eqListItems<T>(eq: (a: T, b: T) => boolean, a?: T[], b?: T[]) {
+    const subset1 = !!a?.every(a1 => !!b?.find(a2 => eq(a1, a2)));
+    const subset2 = !!b?.every(a1 => !!a?.find(a2 => eq(a1, a2)));
+    return subset1 && subset2;
+  }
```
aguillenv, line 31: should you check the length here too?

src/modules/0.0.23/aws_vpc/mappers/route_table.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  findRoutesDiff(dbEntities: Route[], cloudEntities: Route[]) {
```
aguillenv, line 33: hmm, curious why you need this here?

src/modules/0.0.23/aws_vpc/entity/subnet.ts
```diff
@@ -77,6 +78,11 @@ export class Subnet {
   })
   subnetArn?: string;
 
+  @OneToMany(() => RouteTableAssociation, rta => rta.routeTable, {
+    nullable: true,
+  })
+  explicitRouteTableAssociations?: RouteTableAssociation[];
```
aguillenv, line 84: Do we need this side of the relation explicitly for any reason?

src/modules/0.0.23/aws_vpc/entity/route.ts
aguillenv: Actually if `Route` entity will be managed by `RouteTableMapper` you don't need the `@cloudId` concept because there will be no mapper to `cloud.read` a `Route`

src/services/scheduler.ts
```diff
@@ -251,7 +251,7 @@ export async function start(dbId: string, dbUser: string) {
   }
 }
 
-async function getContext(conn: TypeormWrapper, Modules: any): Promise<Context> {
+export async function getContext(conn: TypeormWrapper, Modules: any): Promise<Context> {
```
aguillenv, line 254: where do you need this?

src/modules/0.0.23/aws_vpc/mappers/route_table.ts
mtp1376: I believe these are mathematically enough:
https://qr.ae/pr1NnC

src/services/scheduler.ts
mtp1376: in the functional tests to build the context object.

src/modules/0.0.23/aws_vpc/entity/route.ts
mtp1376: I suppose `generateId` should be used together with a mapper, right? But we don't have a mapper for this entity.

ormconfig.js
```diff
@@ -1,28 +0,0 @@
-const { SnakeNamingStrategy } = require('typeorm-naming-strategies');
```
dfellis, line 1: Can you restore this file, please? Some of the scripts require it.

src/modules/aws_vpc/entity/route_table.ts
```diff
@@ -0,0 +1,59 @@
...
+  id: number;
+
+  @Column({ nullable: true })
+  @cloudId
+  routeTableId?: string;
+
+  @ManyToOne(() => Vpc, {
+    nullable: false,
+    eager: true,
+    onDelete: 'CASCADE',
```
dfellis, line 21: Are we sure we want this cascade here? Also, does this mean deleting the RouteTable would also delete the VPC and Region records, or deleting one of those two records would delete the RouteTable?

src/modules/aws_vpc/entity/route_table_association.ts
```diff
@@ -0,0 +1,47 @@
...
+    nullable: true,
+    eager: true,
+  })
+  @JoinColumn()
+  subnet?: Subnet;
+
+  @ManyToOne(() => Vpc, {
+    nullable: false,
+    eager: true,
+    onDelete: 'CASCADE',
```
dfellis, line 35: Same question here: which direction does this cascade go?

src/modules/aws_vpc/entity/route_table.ts
mtp1376: I believe this means deletion of VPC cascades to its RouteTables and it's expected.

src/modules/aws_vpc/entity/route_table_association.ts
mtp1376: VPC deleted -> RouteTableAssociations are deleted as well.

src/modules/aws_vpc/mappers/route_table.ts
```diff
@@ -0,0 +1,266 @@
...
+      a.InstanceOwnerId === b.InstanceOwnerId &&
+      a.NatGatewayId === b.NatGatewayId &&
+      a.TransitGatewayId === b.TransitGatewayId &&
+      a.LocalGatewayId === b.LocalGatewayId &&
+      a.CarrierGatewayId === b.CarrierGatewayId &&
+      a.NetworkInterfaceId === b.NetworkInterfaceId &&
+      a.VpcPeeringConnectionId === b.VpcPeeringConnectionId &&
+      a.CoreNetworkArn === b.CoreNetworkArn
+    );
+  }
```
dfellis, line 64: You could probably replace this whole thing with [`_.isEqual`](https://docs-lodash.com/v4/is-equal/) which we use in a couple of other modules.

src/modules/aws_vpc/mappers/route_table.ts
```diff
@@ -0,0 +1,266 @@
...
+      this.eqListItems<Route>(this.eqRoute, a.routes, b.routes) &&
+      eqTags(a.tags, b.tags)
+    );
+  };
+
+  eqListItems<T>(eq: (a: T, b: T) => boolean, a?: T[], b?: T[]) {
+    const subset1 = !!a?.every(a1 => !!b?.find(a2 => eq(a1, a2)));
+    const subset2 = !!b?.every(a1 => !!a?.find(a2 => eq(a1, a2)));
+    return subset1 && subset2;
+  }
```
dfellis, line 33: This is actually subtly broken, believe it or not. The following code should show it:

```ts
function eqListItems<T>(eq: (a: T, b: T) => boolean, a?: T[], b?: T[]) {
  const subset1 = !!a?.every(a1 => !!b?.find(a2 => eq(a1, a2)));
  const subset2 = !!b?.every(a1 => !!a?.find(a2 => eq(a1, a2)));
  return subset1 && subset2;
}

const firstThreeWholeNumbers = [1, 2, 3];
const firstFourFibonacciNumbers = [1, 1, 2, 3];

console.log(eqListItems((a, b) => a === b, firstThreeWholeNumbers, firstFourFibonacciNumbers));
```

Which in `ts-node` gives me:

```sh
damocles@zelbinion:~/alantech/iasql-engine(metadata-rpcs)$ ./node_modules/.bin/ts-node
> function eqListItems<T>(eq: (a: T, b: T) => boolean, a?: T[], b?: T[]) {
... const subset1 = !!a?.every(a1 => !!b?.find(a2 => eq(a1, a2)));
... const subset2 = !!b?.every(a1 => !!a?.find(a2 => eq(a1, a2)));
... return subset1 && subset2;
... }
undefined
> const firstThreeWholeNumbers = [1, 2, 3];
undefined
> const firstFourFibonacciNumbers = [1, 1, 2, 3];
undefined
> console.log(eqListItems((a, b) => a === b, firstThreeWholeNumbers, firstFourFibonacciNumbers));
true
undefined
>
```

When those are clearly two different lists.

It *probably* doesn't matter, but first early returning if the list lengths are different, then using a sorting function to deterministically sort the two arrays, and then scan the arrays one-at-a-time using the same sorting function to confirm that they both sort into the same spot (the JS-style sorting function returns zero) would be more correct *and* have a smaller Big-O notation than this implementation (the current one being `O(n^2)` while the described implementation would be `O(n*log(n))` (though for small `n` it would definitely be slower).

Just adding a `length` comparison check on the current implementation wouldn't work, because it would still pass for `[1, 1, 2, 3]` vs `[1, 2, 2, 3]`, for instance.

But if you don't allow duplicates in your comparison, then you can probably get away with what's currently there, perhaps with a pre-check scan on both arrays that none of the arrays have any elements that match with any other elements in the array (harder to do functionally, though, but iteratively on this one with separate `i` and `j` counter variables could speed it up by letting you do the four needed comparisons at the same time, instead of in separate loops, and allow early exiting on any of the four conditions failing).

test/modules/aws-vpc-routetable-integration.ts
```diff
@@ -0,0 +1,262 @@
...
+const begin = runBegin.bind(null, dbAlias);
+const query = runQuery.bind(null, dbAlias);
+const install = runInstall.bind(null, dbAlias);
+const commit = runCommit.bind(null, dbAlias);
+const region = defaultRegion();
+
+const prefix = getPrefix();
+let username: string, password: string;
+
+describe('RouteTable Functional Testing', () => {
```
dfellis, line 36:
```suggestion
describe('RouteTable Integration Testing', () => {
```

src/modules/aws_vpc/mappers/route_table.ts
mtp1376: I was proven wrong mathematically :))

src/modules/aws_vpc/entity/route.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  @Column({ nullable: true })
+  @cloudId
+  DestinationIpv6CidrBlock?: string;
+
+  @Column({ nullable: true })
+  @cloudId
+  DestinationPrefixListId?: string;
+
+  // through
```
yrobla, line 31: what does this comment mean, can you clarify?

src/modules/aws_vpc/entity/route.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  @Column({ nullable: true })
+  @cloudId
+  NetworkInterfaceId?: string;
+
+  @Column({ nullable: true })
+  @cloudId
+  VpcPeeringConnectionId?: string;
+
+  @Column({ nullable: true })
```
yrobla, line 72: all fields are belonging to the cloud id? there is no single identifier for it?

src/modules/aws_vpc/mappers/tags.ts
```diff
@@ -21,3 +21,22 @@ export async function updateTags(client: EC2, resourceId: string, tags?: { [key:
     Tags: tgs,
   });
 }
+
+export function convertTagsFromAws(inputTags?: Tag[]) {
```
yrobla, line 25: is that different from other tags for other entities in AWS?

test/modules/aws-vpc-routetable-integration.ts
```diff
@@ -0,0 +1,262 @@
...
+          INSERT INTO aws_credentials (access_key_id, secret_access_key)
+          VALUES ('${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}')
+      `,
+      undefined,
+      false,
+      () => ({ username, password }),
+    ),
+  );
+
+  it('starts a transaction', begin());
```
yrobla, line 74: why begin/commit here without any other queries?

src/modules/aws_vpc/entity/route.ts
mtp1376: Yes. A Route has two types of values:
- Destination (`DestinationCidrBlock`, `DestinationIpv6CidrBlock`, `DestinationPrefixListId`) which one of the values should be set. They match the destination of a request.
- Through: through what entity should the request be routed? For example it might be a VPC peering connection, an instance id or a NAT gateway.

src/modules/aws_vpc/entity/route.ts
mtp1376: No unfortunately the Route entity in AWS doesn't have anything to identify it with. So we need to use a combination of all the fields to be able to identify it.

src/modules/aws_vpc/mappers/tags.ts
mtp1376: No it's the same. Do we have a function for it?

test/modules/aws-vpc-routetable-integration.ts
mtp1376: That's for syncing the region. We have it in other tests as well. I think we need to do it this way.

src/modules/aws_vpc/mappers/tags.ts
yrobla: We have a similar pattern on ec2, so it could be reused

test/modules/aws-vpc-routetable-integration.ts
yrobla: ok i see, that looked weird to me

mtp1376: Yes I've checked that. We can't do a `iasql_commit` without a `iasql_begin`.

aguillenv (approved):

test/modules/aws-vpc-routetable-integration.ts
dfellis: In the RFC for two-way mode we planned out an `iasql_wait_for` helper function that polls a record waiting for the specified column to be non-null. We haven't implemented that yet, partly because that syntax still feels a bit weird, but perhaps we should?

dfellis (approved):

src/modules/aws_vpc/entity/route.ts
```diff
@@ -0,0 +1,75 @@
...
+  @ManyToOne(() => RouteTable, {
+    onDelete: 'CASCADE',
+    onUpdate: 'CASCADE',
+  })
+  @JoinColumn()
+  routeTable: RouteTable;
+
+  // destination fields: used to determine the destination to be matched
+  @Column({ nullable: true })
+  @cloudId
```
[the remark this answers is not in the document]
mtp1376: We actually don't have an independent mapper for `Route`.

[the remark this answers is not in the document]
mtp1376: Using it here:
https://github.com/iasql/iasql-engine/pull/1559/files#diff-99141b97c9fb18adad20c31c7f845435e0ac07893679d0b5b74e0edb8c74bb6aR37-R49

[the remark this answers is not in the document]
aguillenv: oh! I've missed it 🤦🏼‍♂️

PR #1561, opened 2022-11-17
Mono-Version of IaSQL in the codebase

dfellis:
With the self-contained Docker we can conceivably just deploy different docker containers for different versions, so we don't need all of the extra complications of mantaining multiple versions at the same time. This is still a work in progress, but most of the mechanical stuff is done.


aguillenv (commented):
wow looking good!
image

src/router/index.ts
```diff
@@ -23,7 +23,7 @@ if (config.auth) {
 }
 // TODO secure with scope
 v1.use('/db', db);
-v1.get('/version/latest', (_req, res) => res.send(config.modules.latestVersion));
-v1.get('/version/oldest', (_req, res) => res.send(config.modules.oldestVersion));
+v1.get('/version/latest', (_req, res) => res.send(config.version));
+v1.get('/version/oldest', (_req, res) => res.send(config.version));
```
aguillenv, line 27: since the dashboard would need its own refactor to work with this new approach, does make sense to remove this duplication and just call `/version`?

aguillenv (approved):

src/router/index.ts
dfellis: Yes, but I figured that could be a follow-up PR paired with a dashboard change. Running this code locally with a local dashboard works with this change here.

mtp1376 (approved):
Gonna be really good. My IDE wasn't able to find the right `.ts` file to attach the breakpoints to because there were duplicate codes in previous versions. This is going to make it better.

PR #1565, opened 2022-11-17
HTTP-based RPC

dfellis:
Still early work, but switching to Postgres 14 here (not worrying about upgrading older DBs) and got a restricted level of access to the http extension with a POC function using it

- Switch to Postgres 14 and install the pgsql-http extension
- WIP: HTTP-based RPC


aguillenv (approved):

aguillenv (approved):

src/router/db.ts
```diff
@@ -202,6 +208,112 @@ db.post('/run/:dbAlias', async (req: Request, res: Response) => {
...
+      let errorMessage: string | string[] = logErrSentry(e, uid, email, dbAlias);
+      // split message if multiple lines in it
+      if (errorMessage.includes('\n')) errorMessage = errorMessage.split('\n');
+      // error must be valid JSON as a string
+      error = JSON.stringify({ message: errorMessage });
+      return res.status(400).json({ message: errorMessage });
+    } finally {
+      try {
+        const recordCount = await iasql.getDbRecCount(conn);
+        const rpcCount = 0;
```
depombo, line 284: why are we zero-ing this out now and not simply incrementing it?

depombo:
will pg bouncer be within the same Docker container eventually?

src/router/db.ts
dfellis: Oh, because I figured a follow-up PR would drop it? Now that we don't have a table holding onto the RPCs there's no impact on DB performance or disk usage that the rpc count could cause, anymore, so I figured it was dead weight.

dfellis:
> will pg bouncer be within the same Docker container eventually?

I'm already installing it, but haven't done the work to figure out how to configure it, yet. I was planning to tackle that after the metadata repo has its own RPC-like functions to replace the REST calls into the engine so we can close off the HTTP port from being exposed out of the container.

src/router/db.ts
depombo: it is not quite dead weight. it is used here https://github.com/iasql/iasql-engine/blob/main/src/services/telemetry.ts#L18. we only need to update it here https://github.com/iasql/iasql-engine/blob/main/src/entity/index.ts#L40 even if we are dropping the `iasql_rpc` table

dfellis: Discussed offline. The RPC tracking needs to be reworked with the new Postgres-oriented approach combined with the cron job making RPC calls every ~2 minutes. We can also make it better (faster, stronger) so this will be addressed in a follow-up PR.

depombo (approved):

docker-entrypoint.sh
```diff
@@ -17,6 +17,4 @@ service postgresql restart
 su - postgres -c "psql iasql_metadata -c \"CREATE EXTENSION IF NOT EXISTS pg_cron;\""
 su - postgres -c "psql iasql_metadata -c \"GRANT EXECUTE ON FUNCTION cron.schedule_in_database(text,text,text,text,text,boolean) TO postgres;\""
 
-yarn forever -f dist/services/scheduler.js &
-yarn wait-on http://localhost:14527/health/
 yarn start
```
mtp1376, line 20: Now `yarn start` would suffice to run the program. No need for running multiple processes for `iasql-engine` to start. That's a good thing.

PR #1568, opened 2022-11-18
Increase test timeout

aguillenv:
It has been affecting the `aws_ecs_simplified` test... Let's increase the timeout to 90 min, monitor the test and adjust the timeout in follow-up PR.


yrobla (approved):

aguillenv:
Merging with the known issue of codepipeline. That is fixed in https://github.com/iasql/iasql-engine/pull/1567

PR #1576, opened 2022-11-21
Issue 1570 fix

yrobla:


src/modules/0.0.23/aws_security_group/index.ts
```diff
@@ -140,31 +140,35 @@ class SecurityGroupMapper extends MapperBase<SecurityGroup> {
...
+        // if the network interface is in detached state, we can remove it
+        if (sgEniInfo.NetworkInterfaces) {
+          for (const sg of sgEniInfo.NetworkInterfaces) {
+            if (sg.Attachment?.Status === 'detached') {
+              // we can just delete it
+              try {
+                await client.deleteNetworkInterface({ NetworkInterfaceId: sg.NetworkInterfaceId });
+                continue;
+              } catch (e) {
+                throw new Error('Error deleting network interface');
```
aguillenv, line 160: It would be good to improve this message by adding more info to this error message, like for what sg it happened and maybe the network interface info.

src/modules/0.0.23/aws_security_group/index.ts
```diff
@@ -140,31 +140,37 @@ class SecurityGroupMapper extends MapperBase<SecurityGroup> {
...
+            {
+              Name: 'group-id',
+              Values: [`${instanceParams.GroupId}`],
+            },
+          ],
+        });
+        // if the network interface is in detached state, we can remove it
+        if (sgEniInfo.NetworkInterfaces) {
+          for (const sg of sgEniInfo.NetworkInterfaces) {
+            if (sg.Status === 'available' || sg.Attachment?.Status === 'detached') {
```
aguillenv, line 154: nit: would be better if it is called `eni` instead  of `sg`

aguillenv (approved):

PR #1578, opened 2022-11-22
Rfc garbage

yrobla:


rfcs/006 - Deleting blocked resources asynchronously.md
```diff
@@ -0,0 +1,82 @@
...
+time - even in the range of 24 hours - until the are totally released and can be removed from the cloud.
+This will cause some errors and inconsistencies when managing resources from our engine, and we need to have a way to manage these deletions asynchronously.
+
+## Proposal
+
+We could use the concept of `Garbage collection`, where we can schedule some of those resources for deletion at a certain point of time, and let those be collected regularly by a garbage collector, that will
+check if the resource is in a situation to be deleted and perform it.
+This can be achieved by using a table to track the resources marked for asynchronously deletion, and a periodicaly process that iterates over this table and triggers the proper deletion. Records of this table will contain:
+
+- module and entity of the resource to be deleted
```
aguillenv, line 41: maybe we can save the mapper directly here too.

rfcs/006 - Deleting blocked resources asynchronously.md
```diff
@@ -0,0 +1,82 @@
...
+- cloudId for the resource
+- expected deletion date
```
aguillenv, line 43: what do you mean with this?

rfcs/006 - Deleting blocked resources asynchronously.md
```diff
@@ -0,0 +1,82 @@
...
+The garbage collector procedure will be executed periodically and do the following:
+
+1. Query for all resources in the table with expected deletion date < current data
+2. Identify the module/entity, and the cloud id
+3. Trigger a delete for that resource using the mappers
+4. If the delete succeeds, remove it from the deletion table
+5. If the delete fails, schedule another deletion for the future. We could be using some increasing time (60 minutes, 1 day, etc...). We will also increase the number of deletions.
+6. If the resource failed and number of deletions arrived to maximum, delete from this table as well. Additionally we can throw some error/trigger some email to the user informing about this problem, asking to fix manually.
+
+This procedure will also affect the way we sync resources from the database to the cluster: if we have a resource in the db and not in the cloud, we will check if the resource is on the deletion table. If it is there, we won't trigger the creation in the cloud.
```
aguillenv, line 61: The scenario I see possible is: the resource was deleted from the database, then it error deleting from the cloud. I think what you mean is that when we sync, the cloud resources that need to be created in the database would need to be filtered using this table.

rfcs/006 - Deleting blocked resources asynchronously.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Alternatives Considered
+
```
aguillenv, line 64: another alternative would have added a "soft delete" column on each resource, to know that it was marked and deleted but it was not possible. But this alternative could bring troubles and is not a good UX.

rfcs/006 - Deleting blocked resources asynchronously.md
```diff
@@ -0,0 +1,82 @@
...
+**_Pros:_**
+
+We may be able to delete some resources with some specific per-service logic, being more agile and giving better feedback to the user.
+
+**_Cons:_**
+This is causing huge wait times for the users, eventually causing the engine to be locked, if the user retries the procedure so many times.
+
+## Affected Components
+
+- New periodical procedure `garbage-collector` needs to be implemented
```
aguillenv, line 76: could be implemented with pg_cron 😄

rfcs/006 - Deleting blocked resources asynchronously.md
yrobla: resources cannot be deleted until some expected date. Some of them in 1 hour, some of them in 24 hours,etc... so when we add the resource to the datable, we need to add a "deletion date" so the sync process can pick it when it's just on time, and avoid reading resources not ready to be deleted

aguillenv: hmm I see your point is a good one. What I'm not sure about is how to make this generic but should be doable.

rfcs/006 - Deleting blocked resources asynchronously.md
```diff
@@ -0,0 +1,94 @@
...
+This will cause some errors and inconsistencies when managing resources from our engine, and we need to have a way to manage these deletions asynchronously.
+
+## Proposal
+
+We could use the concept of `Garbage collection`, where we can schedule some of those resources for deletion at a certain point of time, and let those be collected regularly by a garbage collector, that will
+check if the resource is in a situation to be deleted and perform it.
+This can be achieved by using a table to track the resources marked for asynchronously deletion, and a periodicaly process that iterates over this table and triggers the proper deletion. Records of this table will contain:
+
+- module and entity of the resource to be deleted
+- mapper for the resource to be deleted
```
dfellis, line 42: I am a little concerned with the mapper being listed here. It should be derivable from the module + entity, but if someone maliciously tried to make these two mismatch, it could cause chaos if different parts of the eventual code used one definition or the other, right?

rfcs/006 - Deleting blocked resources asynchronously.md
```diff
@@ -0,0 +1,94 @@
...
+- cloudId for the resource
+- expected deletion date
+- number of deletion retries
+
+The procedure for deletion could be the following:
+
+1. Try to delete a resource in the regular way, capturing the `DependencyViolation` exception. If that happens, this is a candidate for the async deletion
+2. Add an entry on the deletion table, with this resource and initial deletion date of a default time (we can start with minimum period , depending on the frequency where the garbage collector will run)
+3. Do not retrigger anymore the deletion, avoiding loops and delays waiting for this resource.
```
dfellis, line 51: How do we avoid the `sync` portion of the `iasql_commit` and/or cron job from re-inserting this resource back in? The mapper-support-side of this needs to be covered in this RFC for this to work, or we'll end up with weird situations where something is marked as deleted, then restored, and then explicitly deleted later but also showing up in the normal table still.

rfcs/006 - Deleting blocked resources asynchronously.md
yrobla: this was added by a previous review, in order to maybe optimize/speed the sync procedure. I can remove, it can just be infered from entity/module

aguillenv: yes, this was my suggestion. My thought was "why to deduce something we already know?" but I get your point, is a valid one!

PR #1587, opened 2022-11-25
Catch get security group error

### LINKED ISSUE #1566: ECS fails in install with misconfigured resources
In [this attempt](https://github.com/iasql/iasql-engine/actions/runs/3488284643/jobs/5843605997#step:7:22570), the ecs service security group got deleted and the test timed out immediately after. In the [next attempt](https://github.com/iasql/iasql-engine/actions/runs/3488284643/jobs/5845456164), when trying to install all modules, ecs service failed because the security group does not exist anymore in the cloud. We should mark this as misconfigured and return undefined or ignore this resource to avoid a hard failure.

aguillenv:
Resolves #1566


src/modules/0.0.23/aws_security_group/index.ts
```diff
@@ -271,7 +271,12 @@ class SecurityGroupMapper extends MapperBase<SecurityGroup> {
         const { groupId, region } = this.idFields(id);
         if (enabledRegions.includes(region)) {
           const client = (await ctx.getAwsClient(region)) as AWS;
-          const rawSecurityGroup = await this.getSecurityGroup(client.ec2client, groupId);
+          let rawSecurityGroup;
```
yrobla, line 274: this is just going to fail silently , not throwing any error?

aguillenv: yes, but now that you question it, I think maybe it is not the best solution for what I want to solve. Let me handle it better

yrobla:
this needs a rebase?

yrobla (approved):

mtp1376 (approved):

Issue #1564, opened 2022-11-17
Add support for keypair import/generation for ec2


yrobla:
Currently users can specify the name of a keypair to use to connect to vms, but they cannot generate/import. It can be done with RPCs in a clean way.

PR #1593, opened 2022-11-28
feat: add initial support for keypairs

yrobla:
Closes: #1564


src/modules/aws_ec2/entity/key_pair.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  @Column({ nullable: true })
+  keyPairId?: string;
+
+  @Column({
+    nullable: true,
+    type: 'enum',
+    enum: KeyType,
+  })
+  type: KeyType;
```
mtp1376, line 27: nit: should we use the datatype from `aws-sdk`?

src/modules/aws_ec2/entity/key_pair.ts
```diff
@@ -0,0 +1,44 @@
...
+export class KeyPair {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @Column()
+  @cloudId
+  name: string;
+
+  @Column({ nullable: true })
+  keyPairId?: string;
```
mtp1376, line 20: Maybe this should be `@cloudId` instead?

test/modules/aws-ec2-integration.ts
```diff
@@ -17,6 +17,9 @@ import {
   runUninstall,
 } from '../helpers';
 
+const { generateKeyPairSync } = require('crypto');
+const sshpk = require('sshpk');
```
mtp1376, line 21: If the `crypto` and `sshpk` support `import` statements rather than `require`, I'd prefer the normal `import` ones here.

mtp1376 (approved):

mtp1376:
BTW your tests are failing because you have not added the `export * from './key_pair';` to the end of your `src/modules/aws_ec2/entity/index.ts` file.

src/modules/aws_ec2/entity/key_pair.ts
yrobla: I have followed this approach on other entities, to avoid redefining the types that were already there and could be used from AWS. Do you think it would give some problems?

src/modules/aws_ec2/entity/key_pair.ts
aguillenv: is the `name` unique for AWS? Usually what I do is set the cloudId as the value we use for the SDK GET method

src/modules/aws_ec2/mappers/key_pair.ts
```diff
@@ -0,0 +1,107 @@
...
+            for (const rk of rawKeyPairs) {
+              const k = await this.keyPairMapper(rk, region);
+              if (k) out.push(k);
+            }
+          }),
+        );
+        return out;
+      }
+    },
+    updateOrReplace: (a: KeyPair, b: KeyPair) => 'replace',
```
aguillenv, line 71: I think this needs to be updated, you are actually updating for the case when `!Object.is(e.fingerprint, cloudRecord.fingerprint) ||
           !Object.is(e.keyPairId, cloudRecord.keyPairId) ||
           !Object.is(e.type, cloudRecord.type)`

src/modules/aws_ec2/entity/key_pair.ts
yrobla: it is unique, i cannot create 2 keys with the same name. It is also the method i use for GET

src/modules/aws_ec2/mappers/key_pair.ts
yrobla: actually i'm not even using it... i just restore/replace

dfellis (approved):

src/modules/aws_ec2/mappers/key_pair.ts
aguillenv: well, I was not suggesting removing it. In your update method, you actually have 2 paths if I'm not wrong. You are updating and you are replacing. It is not so bad, but this method is used also for a better UX telling the user what exactly is happening. Could confuse a user that a replacement occurs but the UI tells them that it is being updated instead.

PR #1599, opened 2022-11-29
feat: add autogenerated documentation for entities

yrobla:


yrobla:
image

yrobla:
It actually misses adding comments to the classes and methods we want to document, so i will be doing that

yrobla:
image

site/docusaurus.config.js
```diff
@@ -21,20 +21,18 @@ const theButton = require('./src/rehype/thebutton');
     url: config.url,
     baseUrl: '/',
     trailingSlash: true,
-    onBrokenLinks: 'throw',
-    onBrokenMarkdownLinks: 'throw',
+    onBrokenLinks: 'warn',
+    onBrokenMarkdownLinks: 'warn',
```
aguillenv, line 25: do we want or need this for any reason?

site/docusaurus.config.js
```diff
@@ -91,17 +90,17 @@ const theButton = require('./src/rehype/thebutton');
               label: 'Dashboard',
             },
             {
-              to: 'docs',
+              href: '/docs',
```
aguillenv, line 93: this change is because the plugin does not understand the `to:` property?

site/typedoc.json
```diff
@@ -0,0 +1,11 @@
+{
+  "entryPoints": ["../src/modules/*" ],
+  "entryPointStrategy": "expand",
+  "cleanOutputDir": true,
+  "excludeExternals": true,
+  "name": "IaSQL documentation",
+  "categorizeByGroup": false,
+  "disableSources": true,
+  "readme": "./sql_reference.md",
```
aguillenv, line 9: do we have this file?

site/docusaurus.config.js
mtp1376: +1

mtp1376 (approved):

site/typedoc.json
yrobla: this is actually not used. The plugin just checks if the readme is set to some value or to "none"

aguillenv (approved):

dfellis:
> image

This is great! But could we group the tables by the module? The current alphabetically-sorted structure is going to be overwhelming, I think, and without the module mention will be confusing to users when they don't have these tables automatically in their database.

dfellis (approved):

depombo (approved):

PR #1602, opened 2022-11-30
Fix `yarn gen-module` command

mtp1376:


aguillenv (commented):
the migration script needs to be updated too

dfellis (approved):
Have you tested this by re-generating some module's migration and confirming that it creates approximately the same file as before?

mtp1376:
> Have you tested this by re-generating some module's migration and confirming that it creates approximately the same file as before?

Yes I tried with `aws_vpc` and it worked fine. The generated tables looked like the previous version and nothing random showed up.

PR #1607, opened 2022-11-30
Docker layer cache CI

mtp1376:


mtp1376:
Does it affect the container build time?
## Old Build Time
[19:21:17](https://github.com/iasql/iasql-engine/actions/runs/3585871471/jobs/6037162034#step:6:1) -> [19:26:53](https://github.com/iasql/iasql-engine/actions/runs/3585871471/jobs/6037162034#step:6:1122) (5m36s)

## New Build Time
[1m20s](https://github.com/iasql/iasql-engine/actions/runs/3587581914/jobs/6038091488#step:7:1)

But the problem is that the new structure needs an "upload artifact" (3m14s) and "download artifact + load image" (1m23s) time which are added (4m37s total). Therefore the optimization only gives us 1 minute (plus some time of compute because the Docker image is only built once -- maybe cost saving in Github Actions).

mtp1376:
Another positive point:
- In the first run, the build is done and maybe some jobs fail.
- When you do a retry, the build is not done again and the already-built image would be reused. This is going to save **4m10s** on all of the retries for failed jobs.

dfellis (approved):

.github/workflows/ci.yml
```diff
@@ -137,6 +137,24 @@ jobs:
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: false
+          load: true
+          cache-from: type=gha
+          cache-to: type=gha
```
aguillenv, line 149: what does `type=gha` mean?

aguillenv (approved):

.github/workflows/ci.yml
mtp1376: It means [Github Actions Cache](https://docs.docker.com/build/building/cache/backends/gha/) which its contents are available at [this page](https://github.com/organizations/iasql/settings/actions/caches).

Different backends are discussed [here](https://docs.docker.com/build/building/cache/backends/#backends).

PR #1609, opened 2022-11-30
[WIP] Handling transaction with SQL instead of engine functions

### LINKED ISSUE #1608: Handling transaction with SQL instead of engine functions
- Add a new property to RpcBase, let's say transactional with possible values 'not-needed' (default), 'fail-if-not-open', 'wait-and-open', 'inner-transaction'.
- Based on this property, update interfaces.ts to generate generic rpc's sql depending on the case. Waiting for a transaction and opening a new one, failing if there is no open transaction, etc

aguillenv:
Resolves #1608


aguillenv:
Talked offline and after the inability to make this PR stable due to race conditions occurring in the database, we have decided not to continue working on this path, at least for now.

PR #1623, opened 2022-12-02
iasql_metadata RPC-like functions

dfellis:
Currently only works properly in the local environment (and probably ci and test), not staging or production. Moves all of the relevant REST endpoints into SQL functions accessible from the `iasql_metadata` database that is automatically created by the engine.

When this PR is done, there will be no accessible REST port. :)


dfellis:
Here's a new database being created and the same fields originally returned as a JSON blob showing up as a record:

Screenshot from 2022-12-01 21-46-37

And here's a listing of databases owned by that user (which is only the just created one):

Screenshot from 2022-12-01 21-46-50

Here's that database being disconnected:

Screenshot from 2022-12-01 21-52-29

And no longer showing up:

Screenshot from 2022-12-01 21-52-53

aguillenv (approved):
looking good to me so far!

CONTRIBUTING.md
```diff
@@ -99,16 +99,17 @@ For Windows you can use below command:
 $env:IASQL_ENV = 'local'; docker-compose up --build
 ```
```
depombo, line 100: the entire architecture section that mentions dblink and the graphite worker also needs to be updated

depombo (approved):

.github/workflows/ci.yml
```diff
@@ -215,20 +215,20 @@ jobs:
           ACCOUNT_INDEX: ${{ steps['determine-test-account'].outputs['account_index'] }}
           ACCESS_KEY_IDS: ${{ secrets.ACCESS_KEY_IDS }}
           SECRET_ACCESS_KEYS: ${{ secrets.SECRET_ACCESS_KEYS }}
-          PSQL_CONN: postgres://postgres:test@localhost:5432/iasql?ssl=true&sslmode=require
+          PSQL_CONN: postgres://postgres:test@localhost:5432/iasql
```
aguillenv, line 218: why do we remove the ssl mode?

dfellis: So Postgres 14 is able to correctly negotiate the proper SSL level without these query params, unlike Postgres 13, so I was just cleaning things up *but* I now notice that I forgot to make the update to the `pg_hba.conf` files to only allow SSL-encrypted connections, and I also have a merge conflict, so let me rebase and also correct that.

aguillenv: oh ok, I was worried about the enforcement of SSL connections, if we are going to update the postgres config to force it I'm ok removing this.

aguillenv (approved):

Dockerfile
```diff
@@ -29,6 +29,5 @@ ARG IASQL_ENV
 ENV IASQL_ENV=$IASQL_ENV
 RUN yarn build
 
-EXPOSE 8088
```
mtp1376, line 32: One liner answer for what the PR does.

src/migration/1669932953007-rpcs.ts
```diff
@@ -0,0 +1,158 @@
...
+        end;
+        $$;
+
+        create or replace function iasql_version() returns text
+        language plpgsql security definer
+        as $$
+        declare
+          _content text;
+        begin
+          select content into _content from http.http_get('http://localhost:8088/v1/version');
```
mtp1376, line 25: Can we use a variable for the `localhost:8088`? I've already pointed an issue on Docker Desktop for Mac (same in Windows) in which the `network_mode: host` doesn't work. In order to make that work, these `localhost:8088`'s need to be changed to `host.docker.internal:8088`.

src/router/db.ts
```diff
@@ -100,24 +107,19 @@ db.post('/export', async (req: Request, res: Response) => {
...
-  } catch (e) {
-    res.status(500).end(logErrSentry(e, uid, email));
-  }
-});
-
-db.get('/disconnect/:dbAlias', async (req: Request, res: Response) => {
+db.post('/disconnect', async (req: Request, res: Response) => {
   logger.info('Calling /disconnect');
-  const { dbAlias } = req.params;
+  const { dbAlias, user } = req.body;
```
mtp1376, line 112: User can send arbitrary post data to the engine, right? In that case I'm worried if this might go in a bad way and `iasql` user database will be dropped?

dfellis: Nope. Here's why:

1. The `pgsql_http` extension is not accessible to any account but the primary account, or functions defined by the primary account with the `security definer` configuration set.
2. The `http` endpoints they can call and the JSON blobs they create are fixed (or "quasi-static") no arbitrary, unsanitized inputs from the user are allowed.
3. The `user` property, in particular, is populated by the `session_user` global variable that the user can't touch. It is the username of the Postgres account calling that function, so they only way to change the `user` is to successfully log in as another user.

src/migration/1669932953007-rpcs.ts
dfellis: That *should* be doable with a new `config` property. Do you already have a PR with that change you're mentioning? I can rebase on top of that and use that property.

mtp1376: No I don't have it as a PR, but I can add it tomorrow. We might do it as a follow-up since I think this should be merged soon?

dfellis: @mtp1376 I think I have addressed this, and also made changes to the migration. We'll see if it works in a bit!

mtp1376 (approved):

src/migration/1669932953007-rpcs.ts
```diff
@@ -0,0 +1,160 @@
...
+          from iasql_database id
+          inner join iasql_user_databases iud on id.pg_name = iud.iasql_database_pg_name
+          where iud.iasql_user_id = session_user
+        $$;
+      `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+        -- drop function iasql_db_list;
```
dfellis, line 148:
```suggestion
        drop function iasql_db_list;
```

dfellis:
That last change didn't fail in my local testing, it only fails *if* you're using the primary user account, like we do in the test suite. Funny.

PR #1624, opened 2022-12-02
Add specific documentation to classes to use new typedoc syntax

yrobla:
image


src/modules/aws_account/entity/index.ts
```diff
@@ -2,32 +2,94 @@ import { Column, Entity, PrimaryGeneratedColumn, PrimaryColumn } from 'typeorm';
...
+ * @see https://github.com/iasql/iasql-engine/blob/main/test/modules/aws-account-integration.ts#L62
+ * @see https://github.com/iasql/iasql-engine/blob/main/test/modules/aws-account-integration.ts#L95
+ *
+ */
 @Entity()
 export class AwsCredentials {
+  /**
+   * @private
+   * @param
+   * Internal ID field for storing accounts
```
dfellis, line 28: Maybe:

```suggestion
   * Auto-incremented ID field for storing accounts
```

dfellis: It's not *exactly* internal since the user can access every column in the database, but it is set for them so they shouldn't need (or want) to do it themselves.

An "internal" field for us would be any properties from the AWS SDK that we use in the mappers but never expose to the user.

src/modules/aws_acm/entity/certificate.ts
```diff
@@ -13,64 +17,123 @@ export enum certificateStatusEnum {
...
+   * Specifies if the certificate is already in use
+   */
   @Column({
     default: false,
   })
   inUse: boolean;
 
+  /**
+   * @public
+   * Region to where the certificate was
```
aguillenv, line 135: hmm incomplete sentence here?

src/modules/aws_acm/entity/certificate.ts
```diff
@@ -13,64 +17,123 @@ export enum certificateStatusEnum {
...
+ */
 export enum certificateRenewalEligibilityEnum {
   ELIGIBLE = 'ELIGIBLE',
   INELIGIBLE = 'INELIGIBLE',
 }
 
+/**
+ * Table to query for all AWS certificates in the system, managed by AWS ACM.
+ * Certificates can be read and deleted, but not created or modified. Instead certificates can be created
+ * by using RPCs.
```
aguillenv, line 44: Should we refer to the concept of RPCs here? Maybe just say like: `... instead certificates can be created by using the following postgres functions <link here to the posgres fns docs>.`

src/modules/interfaces.ts
```diff
@@ -284,7 +284,7 @@ export class MapperBase<E> {
             !Object.keys(fields).every(fk => [primaryColumn].includes(fk))
           ) {
             throw new Error(
-              `Id generation error. Valid field to generate id is: ${primaryColumn}. Receiving: ${Object.keys(
+              `Id generat ion error. Valid field to generate id is: ${primaryColumn}. Receiving: ${Object.keys(
```
aguillenv, line 287:
```suggestion
              `Id generation error. Valid field to generate id is: ${primaryColumn}. Receiving: ${Object.keys(
```

aguillenv (commented):
should this file be here? https://github.com/iasql/iasql-engine/pull/1624/files#diff-735d126242bbf8d14337b7da3ee2bdb5c3e6e1e9f4162333a343b9909ca8e7f9

site/docs/sql/enums/aws_acm_entity_certificate.certificateStatusEnum.md
```diff
@@ -0,0 +1,36 @@
...
+title: "Enumeration: certificate_status"
+sidebar_label: "certificate_status"
+custom_edit_url: null
+---
+
+Different states for a certificate. A valid certificate should be in ISSUED state
+
+## Values
+
+• **EXPIRED** = ``"EXPIRED"``
```
aguillenv, line 12: There is something weird for the enums, it thinks they are some secrets or something 
image

src/modules/aws_account/entity/index.ts
```diff
@@ -2,32 +2,89 @@ import { Column, Entity, PrimaryGeneratedColumn, PrimaryColumn } from 'typeorm';
...
 
+/**
+ * Table that will hold the user's AWS credentials. When a new connection to IaSQL is issued,
+ * the AWS credentials are stored.
+ * The keys can be generated from the AWS console for each registered user
+ *
+ * @example
+ * ```sql
+ *  INSERT INTO aws_credentials (access_key_id, secret_access_key)
+ *  VALUES ('${process.env.AWS_ACCESS_KEY_ID}', '${process.env.AWS_SECRET_ACCESS_KEY}');
```
depombo, line 13: this is technically specific to node.js so we should do something like `axdfa***`

src/modules/iasql_functions/rpcs/iasql_install.ts
```diff
@@ -3,8 +3,34 @@ import { TypeormWrapper } from '../../../services/typeorm';
...
+ * - module_name: Name of the module that was installed
+ * - created_table_name: Name of the associated table that was created
+ * - record_count: Total of registers added
+ *
+ * Accepts the following parameters:
+ * - list of modules to install
+ *
+ * @example
+ * ```sql
+ * SELECT * FROM iasql_install('aws_account');
```
depombo, line 19: the user doesn't typically call this since it is done by the dashboard so let's use another module like vpc or ec2?

src/modules/iasql_functions/rpcs/iasql_install.ts
```diff
@@ -3,8 +3,34 @@ import { TypeormWrapper } from '../../../services/typeorm';
...
+ * ```
+ *
+ * @see https://github.com/iasql/iasql-engine/blob/main/site/docs/reference/function.md
```
depombo, line 22: is this preferred over https://iasql.com/docs/function/?

src/modules/iasql_functions/rpcs/iasql_modules_list.ts
```diff
@@ -2,8 +2,32 @@ import { IasqlFunctions } from '..';
...
+ *
+ * Returns following columns:
+ * - module_name: Name of the module that was installed
+ * - module_version: Version of the modules that was installed
+ * - dependencies: complex type representing the dependencies for the module
+ *
+ *
+ * @example
+ * ```sql
+ * SELECT * FROM iasql_preview_apply()
```
depombo, line 16: typo

src/modules/aws_vpc/entity/vpc.ts
```diff
@@ -3,60 +3,133 @@ import { Column, Entity, JoinColumn, ManyToOne, PrimaryGeneratedColumn, Unique }
...
+ * Table to manage AWS VPC entries.
+ * Amazon Virtual Private Cloud (Amazon VPC) gives you full control over your virtual networking
+ * environment, including resource placement, connectivity, and security.
+ *
+ * @example
+ * ```sql
+ * INSERT INTO vpc (cidr_block, tags, enable_dns_hostnames, enable_dns_support)
+ * VALUES ('192.0.0.0/16', '{"name":"vpc-1"}', true, true);
+ * SELECT * FROM vpc WHERE cidr_block='192.0.0.0/16' AND state='available';
+ * DELETE FROM vpc WHERE cidr_block = '191.0.0.0/16';
```
depombo, line 25: should we move what is in https://iasql.com/docs/vpc here and get rid of that?

src/modules/aws_rds/entity/parameter_group.ts
```diff
@@ -58,38 +61,81 @@ export enum ParameterGroupFamily {
...
+/**
+ * Table to manage AWS RDS parameter groups
+ *
+ * @example
+ * ```sql
+ * INSERT INTO parameter_group (name, family, description) VALUES ('pg_name', 'postgres14', 'description');
+ * SELECT params ->> 'ParameterValue' as value FROM parameter_group, jsonb_array_elements(parameters) as params
+ * WHERE name = 'pg_name' AND params ->> 'DataType' = 'boolean' AND params ->> 'IsModifiable' = 'true';
+ * DELETE FROM parameter_group WHERE name = 'pg_name';
+ * ```
```
depombo, line 73: is it possible to make these `TheButton` arguments in the typedoc plugin so we can move everything from https://iasql.com/docs/aws_rds/ and then centralize all the documentation here?

depombo (approved):

site/typedoc.json
```diff
@@ -3,10 +3,13 @@
...
   "excludeExternals": true,
+  "excludePrivate": true,
+  "excludeInternal": true,
   "name": "IaSQL documentation",
   "categorizeByGroup": false,
   "disableSources": true,
   "readme": "default",
-  "exclude": ["../src/modules/**/mappers/*", "../src/modules/**/migration/*", "../src/modules/iasql_functions/*", "../src/modules/iasql_platform/*", "../src/modules/interfaces.ts", "../src/modules/subscribers.ts"],
+  "exclude": ["../src/modules/**/mappers/*", "../src/modules/**/migration/*", "../src/modules/interfaces.ts", "../src/modules/subscribers.ts",
+"../src/modules/iasql_functions/**", "../src/modules/iasql_platform/**",  ],
```
aguillenv, line 13: to be a valid json shouldn't be like this:
```suggestion
"../src/modules/iasql_functions/**", "../src/modules/iasql_platform/**"]
```

src/modules/aws_acm/entity/certificate.ts
```diff
@@ -13,64 +17,128 @@ export enum certificateStatusEnum {
...
   INELIGIBLE = 'INELIGIBLE',
 }
 
+/**
+ * Table to query for all AWS certificates in the system, managed by AWS ACM.
+ * AWS Certificate Manager (ACM) handles the complexity of creating, storing, and renewing public
+ * and private SSL/TLS X.509 certificates and keys that protect your AWS websites and applications.
+ *
+ * Certificates can be read and deleted, but not created or modified. Instead certificates can be
+ * created or imported by using the following postgres functions:
```
aguillenv, line 47: sounds a bit weird to me that we are saying that Certificates cannot be created or modified but immediately after we are saying they actually can be created using the Postgres functions.

src/modules/aws_acm/entity/certificate.ts
```diff
@@ -13,64 +17,128 @@ export enum certificateStatusEnum {
...
+ *
+ * @see https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html
+ * @see https://aws.amazon.com/certificate-manager
+ *
+ */
 @Entity()
 export class Certificate {
+  /**
+   * @internal
+   * Internal ID field for storing accounts
```
aguillenv, line 64: I think David at some point commented this, but this are "Internal IDs" but database auto-incremented ids?

Also, this seem copied from aws_account.

src/modules/aws_acm/entity/certificate.ts
```diff
@@ -13,64 +17,128 @@ export enum certificateStatusEnum {
...
   @Column({
     nullable: true,
     unique: true,
   })
   @cloudId
   arn?: string;
 
+  /**
+   * @public
+   * Internal ID for the certificate
```
aguillenv, line 82: Maybe:
```suggestion
   * AWS generated ID for the certificate
```

src/modules/aws_dynamo/entity/index.ts
```diff
@@ -3,59 +3,118 @@ import { Column, Entity, JoinColumn, ManyToOne, PrimaryGeneratedColumn } from 't
...
     nullable: false,
     enum: TableClass,
   })
   tableClass: TableClass;
 
-  // TODO: How to constrain this more appropriately in the database?
+  /**
+   * @public
+   * Complex type to represent the provisioned throughput settings for the table
+   * TODO: How to constrain this more appropriately in the database?
```
aguillenv, line 69: so this TODO shows up in the docs?

src/modules/iasql_functions/rpcs/iasql_modules_list.ts
```diff
@@ -2,8 +2,32 @@ import { IasqlFunctions } from '..';
...
+ *
+ * Returns following columns:
+ * - module_name: Name of the module that was installed
+ * - module_version: Version of the modules that was installed
+ * - dependencies: complex type representing the dependencies for the module
+ *
+ *
+ * @example
+ * ```sql
+ * SELECT * FROM iasql_preview_apply();
```
aguillenv, line 16:
```suggestion
 * SELECT * FROM iasql_modules_list();
```

src/modules/aws_dynamo/entity/index.ts
yrobla: ok i found that there is a "privateRemarks" annotation that we can use to add internal comments. I included all the TODOS behind that annotation

PR #1625, opened 2022-12-02
Push to Github Container Registry instead of using artifacts

mtp1376:
This is going to make CI faster because instead of using `upload-artifact` and `download-artifact` (which treat the image as a BLOB), we'll be uploading the image to the Github container registry which does not upload the layers that already exist and therefore the time of uploading the artifact will be significantly reduced.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -152,6 +152,7 @@ jobs:
     needs:
       - setup-module-test
     steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


dfellis (approved):
Presuming the test failure passes on a rerun.

aguillenv (approved):
Should we add a job to clean the container registry?

mtp1376:
> Should we add a job to clean the container registry periodically?

İ think it's a good idea. We can do it as a follow up task.

PR #1636, opened 2022-12-05
Split the common tests

dfellis:
The common tests now take ~20 minutes to run, when successful, so splitting these tests into parallel runs should buy us test run wins.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
         run: yarn --prefer-offline
 
       - name: Run common tests
-        run: yarn coverage $(echo $COMMONS | jq -r '."${{matrix.module}}"')
+        run: yarn coverage $(echo $COMMONS | jq -r '."${{matrix.common}}"')
         env:
           IASQL_ENV: test
           COMMONS: ${{ needs.setup-common-test.outputs['common-tests'] }}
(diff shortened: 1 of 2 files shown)


depombo (approved):

.github/workflows/ci.yml
```diff
@@ -58,9 +58,32 @@ jobs:
       - name: Run code style check
         run: yarn style
 
+  setup-common-test:
+    runs-on: ubuntu-latest
+    outputs:
+      common-tests: ${{ steps['set-common-tests'].outputs['common-tests'] }}
+      set-common-test-names: ${{ steps['set-common-test-names'].outputs['common-test-names'] }}
+      registry-image-tag: ${{ steps['set-registry-image-tag'].outputs['registry-image-tag'] }}
```
mtp1376, line 66: Is this going to work? I think there's no `registry-image-tag` step in this job. Or maybe it's failing silently since we're not using that variable anywhere.

aguillenv: I think this is not used here

aguillenv (approved):

.github/workflows/ci.yml
dfellis: Oh, crap, now I see what you're referring to. Oops. :(

PR #1637, opened 2022-12-05
event tracking for 2-way mode

depombo:


src/modules/iasql_functions/iasql.ts
```diff
@@ -1085,9 +1114,27 @@ async function commitApply(
   } while (ranFullUpdate);
   const t7 = Date.now();
   logger.info(`${dbId} applied and synced, total time: ${t7 - t1}ms`);
+  const out = iasqlPlanV3(toCreate, toUpdate, toReplace, toDelete);
+  if (recordsApplied > 0) {
+    const totalRecordsApplied = await MetadataRepo.incrementRecordsApplied(dbId, recordsApplied);
+    if (totalRecordsApplied) {
```
dfellis, line 1120: What do we do if we hit the 64-bit overflow? :zany_face:

src/modules/iasql_functions/iasql.ts
```diff
@@ -1085,9 +1114,27 @@ async function commitApply(
...
+          output: out.rows.toString()
+        })
+      }
+    }
+
+  }
   await context.orm.dropConn();
   context.orm = oldOrm;
-  return iasqlPlanV3(toCreate, toUpdate, toReplace, toDelete);
+  return out;
```
dfellis, line 1137: I think I'd rather this whole block live in an IIFE that's an async function so it doesn't block the main flow.

src/modules/iasql_functions/iasql.ts
```diff
@@ -1288,9 +1339,26 @@ async function commitSync(
...
```
dfellis, line 1361: Same here. Pull the telemetry `await`s out of the normal flow so it doesn't block the user.

aguillenv (commented):
LGTM

dfellis (approved):

package.json
```diff
@@ -19,9 +19,9 @@
     "gen-sql": "IASQL_ENV=ci ts-node ./node_modules/.bin/typeorm migration:generate -n",
     "migrate": "IASQL_ENV=ci ts-node ./node_modules/.bin/typeorm migration:run",
     "gen-module": "IASQL_ENV=ci ./.module_migration.sh",
-    "docker-build": "docker build -t iasql:latest .",
-    "docker-run": "docker run -e IASQL_ENV=local -p 5432:5432 --name iasql iasql:latest",
-    "docker-run-vol": "docker run -e IASQL_ENV=local -p 5432:5432 -v $(dirname $PWD)/db:/var/lib/postgresql/14/main --name iasql iasql:latest",
+    "docker-build": "DOCKER_DEFAULT_PLATFORM=linux/amd64 docker build -t iasql:latest .",
+    "docker-run": "DOCKER_DEFAULT_PLATFORM=linux/amd64 docker run -e IASQL_ENV=local -p 5432:5432 -p 8088:8088 --name iasql iasql:latest",
+    "docker-run-vol": "DOCKER_DEFAULT_PLATFORM=linux/amd64 docker run -e IASQL_ENV=local -p 5432:5432 -p 8088:8088 -v $(dirname $PWD)/db:/var/lib/postgresql/14/main --name iasql iasql:latest",
```
aguillenv, line 24: Oh we shouldn't be re-exposing the 8088 port

dfellis (approved):

package.json
```diff
@@ -19,9 +19,9 @@
     "gen-sql": "IASQL_ENV=ci ts-node ./node_modules/.bin/typeorm migration:generate -n",
     "migrate": "IASQL_ENV=ci ts-node ./node_modules/.bin/typeorm migration:run",
     "gen-module": "IASQL_ENV=ci ./.module_migration.sh",
-    "docker-build": "docker build -t iasql:latest .",
-    "docker-run": "docker run -e IASQL_ENV=local -p 5432:5432 --name iasql iasql:latest",
-    "docker-run-vol": "docker run -e IASQL_ENV=local -p 5432:5432 -v $(dirname $PWD)/db:/var/lib/postgresql/14/main --name iasql iasql:latest",
+    "docker-build": "DOCKER_DEFAULT_PLATFORM=linux/amd64 docker build -t iasql:latest .",
+    "docker-run": "DOCKER_DEFAULT_PLATFORM=linux/amd64 docker run -e IASQL_ENV=local -p 5432:5432 --name iasql iasql:latest",
+    "docker-run-vol": "DOCKER_DEFAULT_PLATFORM=linux/amd64 docker run -e IASQL_ENV=local -p 5432:5432 -v $(dirname $PWD)/db:/var/lib/postgresql/14/main --name iasql iasql:latest",
```
mtp1376, line 24: Are all the systems running on amd64? Like our servers. If not, this is going to cause the build/run to be slower.

src/modules/iasql_functions/iasql.ts
```diff
@@ -322,6 +323,22 @@ ${Object.keys(tableCollisions)
         dbId,
       );
     }
+    (async () => {
+      const user = await MetadataRepo.getUserFromDbId(dbId);
+      if (user && (moduleNames.length > 0 || moduleNames[0] !== 'aws_account')) {
```
mtp1376, line 328: what about the rest of list? if it's a `iasql_install('aws_account', 'aws_s3')`. Are we relying on the fact that after installing account, we'll need to sync the regions?

src/modules/iasql_functions/iasql.ts
```diff
@@ -322,6 +323,22 @@ ${Object.keys(tableCollisions)
...
+          {
+            dbId,
+          },
+          {
+            params: moduleNames,
+            output: 'Done!',
+          },
+        );
+      }
+    })();
```
mtp1376, line 341: BTW this being non-blocking and run asynchronously is ok, right?

mtp1376 (approved):

src/modules/iasql_functions/iasql.ts
dfellis: I explicitly asked for this. Metrics shouldn't ever block the actual usage for the users. We don't care if they're a fraction of a second late or whatever.

src/modules/iasql_functions/iasql.ts
depombo: oh good point

depombo: this is to ignore the dashboard setup of a database where it calls `aws_account` by itself

CONTRIBUTING.md
```diff
@@ -211,4 +211,8 @@ This section will change much more frequently compared to the other sections.
...
+## How IaSQL PostgreSQL functions work
+
+IaSQL [functions or operations](https://iasql.com/docs/function) like `iasql_install(..)`, `iasql_sync()`, `iasql_apply()` are required to provide a synchronous user experience within a PostgreSQL connection such that the user can't accidentally alter data in tables while an `apply` is taking place whilst still implementing the majority of the logic behind the operations in Node.js. It is worth pointing out that fulfilling this requirement doesn't stop someone from opening two simultaneous connections to the same database or creating two databases pointing to the same cloud account.
+
+IaSQL functions use an RPC mechanism where IaSQL spins up a Node.js HTTP server within the Docker container that is not exposed outside of the container and PGSQL invoke endpoints in the Node.js server using the `pg-http` postgres extension. At startup, the Node.js engine looks at its [metadata repository](https://github.com/iasql/iasql-engine/blob/main/src/entity/index.ts) which stores information about all the IaSQL databases the engine is managing. New databases created via the `/connect` route get a new metadata repo entry.
```
aguillenv, line 218: This has not been updated with the latest CONTRIBUTING changes, right?

src/services/telemetry.ts
```diff
@@ -45,15 +46,17 @@ export async function logEvent(uid: string, event: string, dbProps: DbProps, eve
             event !== DISCONNECT
               ? {
                   [`record_count__${dbProps.dbAlias}`]: dbProps.recordCount,
-                  [`rpc_count__${dbProps.dbAlias}`]: dbProps.rpcCount,
+                  [`records_applied__${dbProps.dbAlias}`]: dbProps.recordCount,
+                  [`records_synced__${dbProps.dbAlias}`]: dbProps.recordCount,
```
aguillenv, line 50: why the same prop here?

CONTRIBUTING.md
depombo: It has, I am adding this TODO here

src/services/telemetry.ts
depombo: good catch

CONTRIBUTING.md
aguillenv: Oh, I was commenting on it because it is referring to legacy `iasql_sync` and `iasql_apply`. Also '/connect' is not an exposed endpoint anymore

depombo: oh I see. I left the original TODO for now since it out of the scope of this PR now

src/migration/1670956536231-new_records.ts
```diff
@@ -0,0 +1,117 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "iasql_database" DROP COLUMN "operation_count"`);
+    await queryRunner.query(`ALTER TABLE "iasql_database" DROP COLUMN "rpc_count"`);
+    await queryRunner.query(
+      `ALTER TABLE "iasql_database" ADD "records_applied" integer NOT NULL DEFAULT '0'`,
+    );
+    await queryRunner.query(`ALTER TABLE "iasql_database" ADD "records_synced" integer NOT NULL DEFAULT '0'`);
+    // iasql_db_list
+    await queryRunner.query(`
+      drop function iasql_db_list;
```
dfellis, line 17: Technically we don't need these `drop function` calls when there's a `create or replace function` call afterwards, but this is probably a bit safer in case Postgres doesn't consider them to have the same function signature.

dfellis (approved):

src/migration/1670956536231-new_records.ts
depombo: I mostly followed convention done from other migration files

aguillenv (approved):

src/migration/1670956536231-new_records.ts
aguillenv: in the case of `iasql_db_list` we have to because the return table is changing and Postgres does not consider them the same to be replaced.

PR #1648, opened 2022-12-10
Restore staging

dfellis:
This branch has the last small changes needed to get staging deploying again. During this process, naming the newly merged engine+DB as `pg-staging.iasql.com`. This may be reverted to `db-staging.iasql.com` but not sure.


aguillenv (approved):

src/services/db-manager.ts
```diff
@@ -39,7 +39,8 @@ function randomHexValue() {
 }
 
 export function genDbId(dbAlias: string) {
-  return config.auth ? `_${randomHexValue()}` : dbAlias;
+  // TODO: Find a better way to distinguish multi-user vs single-user situations
```
depombo, line 42: maybe just check the configuration name directly?

dfellis: Not a good idea because if we want to add another operating environment in the future that should be multi-user the code wouldn't catch that.

But I don't know why I didn't realize that I should just *create* a multi-user flag rather than relying on the fact that currently any usage of Sentry is a multi-user environment, so I'm fixing that.

aguillenv (approved):

depombo (approved):

Issue #1604, opened 2022-11-30
Error deleting lambdas


aguillenv:

depombo:
Needed for https://github.com/iasql/meta/issues/309

PR #1656, opened 2022-12-12
Try to fix CI fails in lambda delete

mtp1376:
Example failure:
https://github.com/iasql/iasql-engine/actions/runs/3666393565/jobs/6198087937#step:7:13301

Happens in LambdaFunction `cloud.delete` when the role does not exist in the database. I initially added `e.role?.arn` and then tried to use the cloud entity value instead.


yrobla (approved):

aguillenv:
lets link this PR to https://github.com/iasql/iasql-engine/issues/1604

depombo (approved):

PR #1657, opened 2022-12-12
Fix the retry bug in Jest tests

mtp1376:
Jest [has issues](https://github.com/facebook/jest/issues/11808) with `instance of` calls. And the AWS SDK is using `instanceof Error` and `instanceof Object` to check the error object in order to decide whether it should do a retry on API error or not.
As a result we see errors in CI tests [not being retried](https://github.com/iasql/iasql-engine/actions/runs/3675331590/jobs/6215623047#step:9:423) and causes jobs to fail.

https://backend.cafe/should-you-use-jest-as-a-testing-library

https://github.com/aws/aws-sdk-js-v3/issues/4273

### DIFF
--- a/src/services/aws_macros.ts
+++ b/src/services/aws_macros.ts
@@ -88,7 +88,7 @@ export class AWS {
         if (!error) {
           return false;
         }
-        if (error.name !== 'Error')
+        if (error.message.includes('AWS SDK error wrapper'))
           // default behavior when running outside Jest
           return defaultRetryDecider(error);
         else return true; // jest has messed the error object
--- a/src/services/aws_macros.ts
+++ b/src/services/aws_macros.ts
@@ -81,22 +81,28 @@ export class AWS {
   memoryDBClient: MemoryDB;
   slowRetryStrategy: StandardRetryStrategy;
   codeBuildRetryStrategy: StandardRetryStrategy;
+  defaultRetryStrategy: StandardRetryStrategy;
 
   constructor(config: AWSConfig) {
+    const jestTunedRetryDecider = (error: any) => {
+      // copied the default behavior from aws-sdk to fix the problem caused by Jest on isRetryableByTrait
+      if (!error) {
+        return false;
+      }
+      if (IASQL_ENV === 'test' && error.message.includes('AWS SDK error wrapper')) {
+        // jest has messed the error object
+        return true;
+      }
+      // default behavior when running outside Jest
+      return defaultRetryDecider(error);
+    };
     // declare a specific slow retry strategy, to reuse in slow apis
+    this.defaultRetryStrategy = new StandardRetryStrategy(async () => 3, {
+      retryDecider: jestTunedRetryDecider,
+    });
+
     this.slowRetryStrategy = new StandardRetryStrategy(async () => SLOW_STRATEGY_RETRIES, {
-      retryDecider: error => {
-        // copied the default behavior from aws-sdk to fix the problem caused by Jest on isRetryableByTrait
-        if (!error) {
-          return false;
-        }
-        if (IASQL_ENV === 'test' && error.message.includes('AWS SDK error wrapper')) {
-          // jest has messed the error object
-          return true;
-        }
-        // default behavior when running outside Jest
-        return defaultRetryDecider(error);
-      },
+      retryDecider: jestTunedRetryDecider,
       delayDecider: (_, attempts) =>
         Math.floor(
           Math.min(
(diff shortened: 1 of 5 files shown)


src/services/aws_macros.ts
```diff
@@ -81,6 +81,16 @@ export class AWS {
   constructor(config: AWSConfig) {
     // declare an specific slow retry strategy, to reuse in slow apis
     this.slowRetryStrategy = new StandardRetryStrategy(async () => SLOW_STRATEGY_RETRIES, {
+      retryDecider: error => {
+        // copied the default behavior from aws-sdk to fix the problem caused by Jest on isRetryableByTrait
+        if (!error) {
+          return false;
+        }
+        if (error.name !== 'Error')
```
aguillenv, line 89: but isn't jest overriding the `error` and not having the `name` property? If no `error.name` it will still go through this path?

mtp1376: Jest is making the error not an instance of `Error`. Therefore the AWS SDK replaces it with a `new Error(...)` which has `name = 'Error'`.

aguillenv: ohh got it!

mtp1376:
As seen in `@aws-sdk/middleware-retry` code, the `asSdkError` function is checking whether the error is an instance of `Error`, if so returns it but if it's not (in Jest's case the last statement is executed) it's wrapped using a new `Error` object.
https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-retry/src/StandardRetryStrategy.ts#L142-L147
The new `Error` object lacks the needed properties for and `shouldRetry` call therefore returns `false`:
https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-retry/src/StandardRetryStrategy.ts#L96
https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-retry/src/StandardRetryStrategy.ts#L42
Because the `defaultRetryDecider` is using functions that rely on `error.name` which is replaced by `Error` when a new object is created and therefore they don't work as expected:
https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-retry/src/retryDecider.ts#L9-L15
For example `isTransientError` is using `error.name` as follows:
https://github.com/aws/aws-sdk-js-v3/blob/e7b1535fb6707bb84241d0bd21ca339037094246/packages/service-error-classification/src/index.ts#L26-L29

mtp1376:
@dfellis I think this is a good improvement for the `socket hang up` errors. But it has caused some of the tests to take longer than expected. Because it's retrying blindly in the test suites currently. We need to somehow replicate the original behavior.

test/functional/aws-error-jest.ts
```diff
@@ -0,0 +1,19 @@
+import { AWS } from '../../src/services/aws_macros';
```
dfellis, line 1: These tests in the `functional` directory are not run by the test suite.

dfellis (approved):

src/services/aws_macros.ts
```diff
@@ -81,6 +81,16 @@ export class AWS {
...
+      retryDecider: error => {
+        // copied the default behavior from aws-sdk to fix the problem caused by Jest on isRetryableByTrait
+        if (!error) {
+          return false;
+        }
+        if (!error.message.includes('AWS SDK error wrapper'))
+          // default behavior when running outside Jest
+          return defaultRetryDecider(error);
+        else return true; // jest has messed the error object
+      },
```
aguillenv, line 93: wouldn't be better to handle this using the `IASQL_ENV` variable to know if we are running with jest? I wonder if it is possible that any other error that does not include the `AWS SDK error wrapper` message should go through the `defaultRetryDecider` and we are changing that behaviour

depombo: I agree with this. Furthermore, there are other retryDeciders in this file. Not just this one.

aguillenv (approved):

src/services/aws_macros.ts
```diff
@@ -79,17 +81,19 @@ export class AWS {
...
       retryDecider: error => {
         // copied the default behavior from aws-sdk to fix the problem caused by Jest on isRetryableByTrait
         if (!error) {
           return false;
         }
-        if (!error.message.includes('AWS SDK error wrapper'))
-          // default behavior when running outside Jest
-          return defaultRetryDecider(error);
-        else return true; // jest has messed the error object
+        if (IASQL_ENV === 'test' && error.message.includes('AWS SDK error wrapper')) {
```
dfellis, line 91: No, please don't hook on the `IASQL_ENV` value directly. Create a new `overrideAwsRetryDecider` (or something like that) boolean that you set in `src/config/test.ts` to `true` and `false` elsewhere. That way if there is another environment in the future that needs this behavior, it can also set it.

aguillenv: good solution

aguillenv (approved):

dfellis (approved):

Issue #1605, opened 2022-11-30
Error deleting IamUser


aguillenv:

depombo:
Might entail removing AWS access keys belonging to IAM users. A follow-up to this is an rpc function that lets you create the an access key.

PR #1665, opened 2022-12-13
fix: allow to delete users with keys or profiles

yrobla:
Closes: #1605


depombo:
I guess this is hard to test until we have the rpc that creates it, but something to add once we have that later on

depombo (approved):

aguillenv (approved):
oh, this definitely works, but I thought we were going to add actual entities with at least read/delete support for access keys (and login profiles). Could be done later

yrobla:
I actually wanted to do it in a follow up, after tasks with less priority. I just wanted to fix the bug. But I will create an issue for it. It is not only keys but also user profile

PR #1666, opened 2022-12-13
Remove the redundant delete because it's an impossible case

mtp1376:
```ts
export class RepositoryPolicy {
...
  @OneToOne(() => Repository, { nullable: false, eager: true })
  @JoinColumn([
    {
      name: 'repository_id',
      referencedColumnName: 'id',
    },
    // we defined this one to make sure we are using the right region
    {
      name: 'region',
      referencedColumnName: 'region',
    },
  ])
  repository: Repository;
...
}
```
It has `nullable: false`. So all `repositoryPolicy` objects related to a repository should be first deleted in order for it to be deleted from the database. Therefore that additional read/delete is redundant and can be deleted.

### DIFF
--- a/src/modules/aws_ecr/index.ts
+++ b/src/modules/aws_ecr/index.ts
@@ -626,7 +626,7 @@ class RepositoryPolicyMapper extends MapperBase<RepositoryPolicy> {
       const opts = repositoryName
         ? {
             where: {
-              repositoryName,
+              repository: { repositoryName },
               region,
             },
           }
--- a/src/modules/aws_ecr/index.ts
+++ b/src/modules/aws_ecr/index.ts
@@ -626,7 +626,7 @@ class RepositoryPolicyMapper extends MapperBase<RepositoryPolicy> {
       const opts = repositoryId
         ? {
             where: {
-              repositoryId,
+              repository: { id: repositoryId },
               region,
             },
           }
--- a/src/modules/aws_ecr/index.ts
+++ b/src/modules/aws_ecr/index.ts
@@ -543,10 +543,6 @@ class RepositoryMapper extends MapperBase<Repository> {
       for (const e of es) {
         const client = (await ctx.getAwsClient(e.region)) as AWS;
         await this.deleteECRRepository(client.ecrClient, e.repositoryName!);
-        // Also need to delete the repository policy associated with this repository,
-        // if any
-        const policy = await this.module.repositoryPolicy.db.read(ctx, this.entityId(e));
-        await this.module.repositoryPolicy.db.delete(policy, ctx);
       }
     },
   });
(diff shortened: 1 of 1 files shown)


src/modules/aws_ecr/index.ts
```diff
@@ -626,11 +622,11 @@ class RepositoryPolicyMapper extends MapperBase<RepositoryPolicy> {
...
     delete: (es: RepositoryPolicy[], ctx: Context) => ctx.orm.remove(RepositoryPolicy, es),
     read: async (ctx: Context, id?: string) => {
-      const [repositoryName, region] = id?.split('|') ?? [undefined, undefined];
-      const opts = repositoryName
+      const [repositoryId, region] = id?.split('|') ?? [undefined, undefined];
+      const opts = repositoryId
         ? {
             where: {
-              repositoryName,
+              repositoryId,
```
aguillenv, line 629: If I'm not wrong this won't work, because the `RepositoryPolicy` entity does not have a `repositoryId` property. I think it would be something like this:
```suggestion
              repository: { id: repositoryId },
```

src/modules/aws_ecr/index.ts
```diff
@@ -626,11 +622,11 @@ class RepositoryPolicyMapper extends MapperBase<RepositoryPolicy> {
     update: (es: RepositoryPolicy[], ctx: Context) => ctx.orm.save(RepositoryPolicy, es),
     delete: (es: RepositoryPolicy[], ctx: Context) => ctx.orm.remove(RepositoryPolicy, es),
     read: async (ctx: Context, id?: string) => {
-      const [repositoryName, region] = id?.split('|') ?? [undefined, undefined];
-      const opts = repositoryName
+      const [repositoryId, region] = id?.split('|') ?? [undefined, undefined];
```
aguillenv, line 625: This is not correct. When `RepositoryPolicy` `db.read` is called it will be passing the `RepositoryPolicy.generateId()` method as id. The `generateId`, `entityId` and `idFields` methods are handling the cloud identification of the policy based on the `repositoryName`. So here we should keep `repositoryName` as the value retrieved from the `id` and the fix should be in the where clause using `repository: { repositoryName }` instead of `repository: { id: repositoryId }`

mtp1376: right

src/modules/aws_ecr/index.ts
```diff
@@ -622,11 +622,11 @@ class RepositoryPolicyMapper extends MapperBase<RepositoryPolicy> {
...
     delete: (es: RepositoryPolicy[], ctx: Context) => ctx.orm.remove(RepositoryPolicy, es),
     read: async (ctx: Context, id?: string) => {
-      const [repositoryId, region] = id?.split('|') ?? [undefined, undefined];
-      const opts = repositoryId
+      const [repositoryName, region] = id?.split('|') ?? [undefined, undefined];
+      const opts = repositoryName
         ? {
             where: {
-              repository: { id: repositoryId },
+              repositoryName,
```
aguillenv, line 629: Even though this path is clearly not being exercised, we should change this too with `repository: { repositoryName },`

aguillenv (approved):

depombo (approved):

PR #1670, opened 2022-12-14
feat: remove codebuild_import table and replace with an rpc

yrobla:
Closes: #1635


aguillenv (approved):

src/modules/aws_codebuild/rpcs/start_build.ts
```diff
@@ -0,0 +1,146 @@
...
+import { Context, RpcBase, RpcResponseObject } from '../../interfaces';
+
+/**
+ * Method for triggering the build of a project
+ *
+ * Returns following columns:
+ *
+ * - name: the name of the project that was built
+ *
+ * - status: OK if the certificate was imported successfully
```
mtp1376, line 14:
```suggestion
 * - status: OK if the build was started successfully
```

src/modules/aws_codebuild/rpcs/start_build.ts
```diff
@@ -0,0 +1,146 @@
...
+        return [
+          {
+            name,
+            status: 'KO',
+            message: 'Error getting status of build',
+          },
+        ];
+      }
+    } else {
+      return [
```
mtp1376, line 131: I suggest you add early returns instead of nested conditions. I mean `if (! projectObj) return [...]` and just have the normal behavior in case no returns triggered.

src/modules/aws_codebuild/rpcs/start_build.ts
```diff
@@ -0,0 +1,146 @@
...
+        const dbBuild = await this.module.buildList.buildListMapper(cloudBuild, ctx, projectObj.region);
+        if (!dbBuild) {
+          return [
+            {
+              name,
+              status: 'KO',
+              message: 'Error starting build',
+            },
+          ];
+        } else {
```
mtp1376, line 109: This else is not needed

src/modules/aws_codebuild/rpcs/start_build.ts
```diff
@@ -0,0 +1,146 @@
...
+
+          return [
+            {
+              name,
+              status: 'OK',
+              message: '',
+            },
+          ];
+        }
+      } else {
```
mtp1376, line 121: Same early return suggestion here

test/modules/aws-codebuild-integration.ts
```diff
@@ -306,31 +306,16 @@ phases:
...
-
-  it('apply build start', commit());
-
-  it(
-    'check build imports is empty',
-    query(
-      `
-    SELECT * FROM codebuild_build_import
-    WHERE project_name = '${dbAlias}';
+      SELECT * FROM start_build('${dbAlias}', '${region}');
```
mtp1376, line 313: Don't we need an `iasql_begin` here? Wondering what will happen in case cron tick happens

yrobla: that's an rpc, so was not enclosed in begin/commit, follows the same logic as others

mtp1376 (approved):

depombo (approved):

mtp1376:
BTW here we kept the `CodebuildBuildList` having `List` suffix. I think it'd be better to trim that suffix as a follow-up.

PR #1671, opened 2022-12-14
Revive code_* modules. Add a test for codebuild and push to ECR

mtp1376:


src/modules/iasql_functions/iasql.ts
```diff
@@ -77,9 +77,7 @@ export async function modules(all: boolean, installed: boolean, dbId: string) {
     }));
   if (all) {
     // TODO: Remove this filter once two-way mode is standard
```
depombo, line 79: remove the TODO now that you did it :)

test/modules/aws-codebuild-integration.ts
```diff
@@ -361,13 +361,87 @@ phases:
   );
 
   it('starts a transaction', begin());
+  it(
+    'creates a project that pushes to ecr',
+    query(
```
depombo, line 366: you should probably just use the `generate_put_ecr_image_build_spec` function here?

dfellis: If that is used elsewhere in the test suite, then I would say it's fine. Better to exercise more possible paths.

mtp1376: It's "building" a specific folder. In this case I've just pulled an image, retagged and pushed (to reduce the rate of errors). For that case we'll need to add Github credentials which will test multiple things.
If that's not a blocker, I'd suggest we go with the current version.

depombo: that makes sense 👍

depombo (approved):

aguillenv (approved):

PR #1675, opened 2022-12-14
Add db user to logger

### LINKED ISSUE #1554: Add user id to mezmo
this will be necessary after the re-architecture when more than one apply loop will be running

aguillenv:
Closes #1554 

image


src/modules/iasql_functions/iasql.ts
```diff
@@ -102,6 +102,7 @@ export async function install(
   syncContext?: Context,
   ormOpt?: TypeormWrapper,
 ) {
+  const loggerWithUser = logger.scope({ dbUser });
```
depombo, line 105: hmm I was hoping this would be per user using the auth0 id like telemetry. I originally did telemetry per db, but the drawback with this approach is that you cannot see all the interactions per user since disconnects and connects won't show up and form different users entirely so I won't be able to ask of mezmo to truly show me all logs for a specific user

src/services/logger.ts
```diff
@@ -21,21 +22,33 @@ const logFactory: LogFunctionFactory<Logger> = scope => {
       console.log('Fatal error in LogDNA', event);
     });
     if (config.logger.forceLocal) {
-      return (level, message, meta) => {
-        console.log(`${level}: ${message}`, meta);
+      return async (level, message, meta) => {
+        const { dbId } = scope;
+        let userId: string | undefined;
+        if (dbId) {
+          const user = await MetadataRepo.getUserFromDbId(dbId);
```
depombo, line 29: does this happen in the execution of the main path?

aguillenv: No, because we never `await` the `logger.` calls, so this occur asynchronously.

depombo (approved):

src/services/logger.ts
```diff
@@ -21,21 +22,33 @@ const logFactory: LogFunctionFactory<Logger> = scope => {
...
+          userId = user?.id;
+        }
+        console.log(`${level}: ${message}`, { ...meta, userId });
         logfn.log(message, {
           level: level === 'warning' ? 'warn' : level, // Graphile Logger vs LogDNA levels fix
-          meta,
+          meta: { ...meta, userId },
```
dfellis, line 35: Why not add the `dbId`, too?

aguillenv: makes sense

dfellis (approved):

PR #1676, opened 2022-12-15
RPC transaction wrapper

### LINKED ISSUE #1674: Wrap all rpc calls in transactions automatically
We lost serialization from graphile worker with the new architecture.

Potential race condition between rpc call and cron job.

aguillenv:
Closes #1674 

It introduces pre- and post- transaction checks for RPCs. This will let the RPC handle the transaction accordingly if needed.

Before executing the RPC, pre-transaction checks will let us know if a transaction is necessary. An RPC can have one of the following pre-transaction checks:
- `NO_CHECK`: Some RPCs do not need a transaction to be open since they may not modify the database schema, this could be the case of the tail logs RPC.
- `WAIT_FOR_LOCK`: RPCs need to open a transaction (if possible) before executing the RPC.
- `FAIL_IF_NOT_LOCKED`: This will be the case of special RPCs like `iasql_commit` or `iasql_rollback`. This kind of RPC  needs that a transaction has been explicitly opened before it can run.
- If not defined, the default is to automatically call `iasql_begin()` before executing.

The post-transaction checks are defined to know what to do after the RPC call was executed:
- `NO_CHECK`: Some RPCs do not need to do anything after they execute.
- `UNLOCK_IF_SUCCEED`: Some RPCs only need to close the transaction if they succeed, for example, `iasql_commit`.
- `UNLOCK_ALWAYS`: Try to close the transaction no matter the outcome of the RPC call.
- If not defined, the default is to automatically call `iasql_commit()` after executing.


src/router/db.ts
```diff
@@ -229,6 +238,72 @@ db.post('/rpc', async (req: Request, res: Response) => {
...
+        const openTransaction = await iasqlFunctions.isOpenTransaction(conn);
+        if (!openTransaction) {
+          throw new Error('Cannot execute without calling iasql_begin() first');
+        }
+        break;
+      case PreTransactionCheck.WAIT_FOR_LOCK:
+        await iasqlFunctions.maybeOpenTransaction(conn);
+        break;
+      default:
+        await conn.query(`select * from iasql_begin();`);
```
mtp1376, line 269: What is this for? Is there a case we'll need it?

src/router/db.ts
```diff
@@ -229,6 +238,72 @@ db.post('/rpc', async (req: Request, res: Response) => {
...
+        if (!errorOcurred) {
+          await iasqlFunctions.closeTransaction(conn);
+        }
+        break;
+      case PostTransactionCheck.UNLOCK_ALWAYS:
+        await iasqlFunctions.closeTransaction(conn);
+        break;
+      default:
+        try {
+          await conn.query(`select * from iasql_commit();`);
```
mtp1376, line 295: Same question for here.

mtp1376 (approved):
Looks great!

src/router/db.ts
aguillenv: It is the default behaviour for cases like the one Yolanda has where the rpc is not doing the whole heavy load of syncing the DB. With the checks we only open and close transactions but never commit at the end.

mtp1376: Oh got it. So this is just for the transactions and not for `commit`/`sync`, right?

aguillenv: not sure I follow your question

depombo (approved):

PR #1677, opened 2022-12-15
feat: replace codedeploy deployment create with an RPC 

yrobla:


depombo:
how did code-deploy create work before this rpc? I'm not sure I follow

src/modules/aws_codedeploy/mappers/deployment.ts
```diff
@@ -174,20 +128,23 @@ export class CodedeployDeploymentMapper extends MapperBase<CodedeployDeployment>
...
+          const cloudRecord = ctx?.memo?.cloud?.Deployment?.[this.generateId({ depId, region })];
+          cloudRecord.id = e.id;
+          await this.module.deployment.db.update(cloudRecord, ctx);
+          out.push(cloudRecord);
+        }
       }
       return out;
     },
     delete: async (deployments: CodedeployDeployment[], ctx: Context) => {
+      // deployments can't actually be deleted, we just need to recreate
```
aguillenv, line 147: do they have an expiration date or AWS deleted them at some point?

aguillenv (approved):

src/modules/aws_codedeploy/mappers/deployment.ts
yrobla: they are deleted when you delete the related application. That's what i've seen from the user interface. You cannot delete deployments per se, but they get removed when you delete the main app.

yrobla:
> how did code-deploy create work before this rpc? I'm not sure I follow

we actually didn't create. They were just used by codepipeline, so they were created when a pipeline was deployed

src/modules/aws_codedeploy/rpcs/start_deploy.ts
```diff
@@ -0,0 +1,218 @@
...
+    region: string,
+  ): Promise<RpcResponseObject<typeof this.outputTable>[]> => {
+    if (!applicationName) {
+      return [
+        {
+          id: '',
+          status: 'KO',
+          message: 'Please provide the name of the CodeDeploy application to deploy',
+        },
+      ];
```
mtp1376, line 84: I think you can refactor these into a `makeError` and `makeSuccess` private helper functions

src/modules/aws_codedeploy/rpcs/start_deploy.ts
```diff
@@ -0,0 +1,218 @@
...
+        {
+          id: deploymentId,
+          status: 'KO',
+          message: 'Error getting updated deployment',
+        },
+      ];
+    }
+
+    // create the deploy in the database
+    await this.module.deployment.db.create(dbDeploy, ctx);
```
mtp1376, line 202: You don't need this since there will be a `iasql_commit` after the RPC is finished. Also, this will prevent direct calls from other RPCs since all the FKs should be present.

mtp1376 (approved):

aguillenv (approved):

PR #1679, opened 2022-12-15
feat: modify sidebars to do not autogenerate sql

yrobla:


yrobla:
image

aguillenv (approved):

site/sidebars.js
```diff
@@ -11,16 +11,76 @@
...
+        {
+          type: 'category',
+          label: 'SQL reference',
+          collapsible: true,
+          collapsed: true,
+          items: [
+            {
+              type: 'doc',
+              label: 'SQL reference',
+              id: 'sql/sql_reference',
```
depombo, line 79: why not just `reference`?

yrobla: I saw that reference generates the link but does not add it to navigation. Didn't test however, it may have worked as well.

PR #1681, opened 2022-12-15
RFC 006 - Replicate changes between staging and prod

depombo:
RFC 006 - Sync changes across databases. We need a good story to import or replicate changes from one database to the other since most people manage staging and production as separate cloud accounts


rfcs/006 - Sync changes across databases.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Alternatives Considered
+
+There are several types of solutions to replicate staging infrastructure changes in production.
+
+### Dump a database and import the dump into another database
+
+An `export` route in the engine implemented using `pgdump` with logic to attempt to make it generic like excluding `aws_account` from the dump and so on. The dump is then intended to be passed into an `import` route for another database. `import` was deprecated for the time being due to the sheer amount of special-casing, which compounded per module, that had to be built within the functionality of export and import.
+
+### "Sync" a database to a different cloud account than the one it is connected to
```
aguillenv, line 47: The issue I see with this approach is that it might be easy for the sync job to fail due to duplicated FKs. Some resources need a unique name as an identifier and if they do not have a good name convention they will collide. Also, resources like security groups would not be easy to handle.

rfcs/006 - Sync changes across databases.md
```diff
@@ -0,0 +1,64 @@
...
+
+TBD
+
+## Alternatives Considered
+
+There are several types of solutions to replicate staging infrastructure changes in production.
+
+### Dump a database and import the dump into another database
+
+An `export` route in the engine implemented using `pgdump` with logic to attempt to make it generic like excluding `aws_account` from the dump and so on. The dump is then intended to be passed into an `import` route for another database. `import` was deprecated for the time being due to the sheer amount of special-casing, which compounded per module, that had to be built within the functionality of export and import.
```
aguillenv, line 45: I remember the troubles we had with import... It is hard to keep the relations, more now that we had to go back to joining using the autoincremented ids.

rfcs/006 - Sync changes across databases.md
```diff
@@ -0,0 +1,64 @@
...
+
+### "Sync" a database to a different cloud account than the one it is connected to
+
+An IaSQL function that takes the alias of another dataabse owned by the same user as input. `iasql_import(‘staging’)` will return the same data structure as `iasql_commit` or `iasql_preview`. This function can reuse parts of the mappers and existing IaSQL functions to run diffs across two different databases or accounts. There could have an optional `dry_run` parameter or another function to preview the changes: `iasql_preview_import(‘staging’)` or `iasql_import(‘staging’, true)`.
```
aguillenv, line 49: we could have a new column named `alias` and n rows in the `aws_credentials` table. The first one is the default  and and the other rows point to other accounts

rfcs/006 - Sync changes across databases.md
```diff
@@ -0,0 +1,64 @@
...
+
+We have leaned deeper into the infrastructure as data workflow by dropping migrations from our deployments and favoring two-way mode with transactions over IaC’s apply. This plays to our strengths and provides a compelling workflow to manage infrastructure. However, we need a good story to import or replicate changes from one database to the other since most people manage staging and production as separate cloud accounts.
+
+One important requirement to keep in mind is being able to exclude certain resources from this special sync. In other words, there might be resources in staging that people don't want to replicate in production and vice versa. This requirement can be sidestepped initially, but whatever design we pick should be able to support this down the line.
+
+## Proposal
+
+TBD
+
+## Alternatives Considered
```
aguillenv, line 39: I do not have a solution, but I think it should contemplate the following facts:
- vpcs and subnets are different in all databases/accounts
- every account has different security groups that are linked to the vpc/subnets
- most of our joins are done using auto-incremented ids.

aguillenv (commented):
I like the alternative to sync from another account so far but I forsee some issues that I'm not sure how to handle.

depombo:
> I like the alternative to sync from another account so far but I forsee some issues that I'm not sure how to handle.

Yes, you bring up a lot of great points. David came up with a few options and I will update the rfc

rfcs/006 - Sync changes across databases.md
```diff
@@ -0,0 +1,84 @@
...
+
+The IaSQL audit log table is timestamped and looking at the records from a given point in time for a staging database will give users most of the information they need to be able to replicate the changes in production. This proposal is grounded on that fact and seeks to make that workflow as smooth as possible by creating a feature that generates SQL from the audit log from a given point in time. The given point in time for this use case would be the last time staging and production were synced. The generated SQL will then be massaged and manually run by the user in production via a PostgreSQL client or a migration within a plain SQL migration system for teams that want to code review infrastructure changes. The IaSQL audit log is ordered chronologically and the SQL generated by a `iasql_get_sql_since()` function should not cause duplicate records or FK relation problems even if the expectation is that the generated SQL should be massaged.
```
aguillenv, line 41: But if someone has never used IaSQL and already has a staging env and they want to replicate that in production, they will not be able to do it. This approach will work for someone with some IaSQL usage, right?

aguillenv: And I'm not sure if I get the last sentence. The result from `iasql_get_sql_since()` won't be pretty similar to an export dump?

depombo: > And I'm not sure if I get the last sentence. The result from `iasql_get_sql_since()` won't be pretty similar to an export dump?

an export dump generates the SQL for the entire database while the proposed function takes in a timestamp and generates SQL to replicate the changes from the provided timestamp until `now()`

depombo: > But if someone has never used IaSQL and already has a staging env and they want to replicate that in production, they will not be able to do it. This approach will work for someone with some IaSQL usage, right?

This will work for someone that has two IaSQL databases: a staging and a production one

dfellis: > But if someone has never used IaSQL and already has a staging env and they want to replicate that in production, they will not be able to do it. This approach will work for someone with some IaSQL usage, right?

That's not really true? When you first install the modules the audit log still shows all of the already existing records as inserts, so if they really wanted to set the timestamp to before they created the database they can.

But the more common would be connecting staging and production to two databases, then making a change in staging, then getting the SQL that represents that change and minimally editing it to run in production.

dfellis: Oh, one thing we should probably do: for INSERT statements that are referencing an auto-generated ID, we should generate the SQL with a sub-query where we select on the `@cloudId` column's value, if available. It should be immediately clear to the user that they need to substitute that string with the production equivalent ID, and in some situations where the cloud ID is a user-provided string, it may not even need any changes.

yrobla: One problem i see about using audit logs is that the target account may be in a different state than the source one. An user may have done only 2 or 3 changes in _account 1_, but for some reason may have done other manual operations on _account 2_. If we base our sync in logs from certain changes, we are not going to properly reflect the status of account 1 into account 2.

rfcs/006 - Sync changes across databases.md
```diff
@@ -0,0 +1,84 @@
...
+
+### 3. Interactive `git rebase` style sync
+
+Similar to option 2, but when conflict arises prompt the user to handle the merge conflicts between the two cloud accounts or databases interactively. The PostgreSQL REPL doesn't allow interactive input so that would have to happen using a two-phase process via an interim table or the dashboard. This option requires a lot of resources to implement out of the gate. The current proposal and if this workflow proves to be primordial to users we can eventually prioritize it.
+
+### 4. Use existing PostgreSQL data synchronization tools
+
+There is a breadth of tools to keep two databases in sync. However, they are not enough to satisfy our requirements for this problem because the data should not be copied over exactly as-is. We have cloud-generated IDs as columns in a lot of tables that are specific to a resource in each account like an EC2 instance id.
+
+### 5. Emulate a declaration via ORMs with introspection + Upserts
```
yrobla, line 70: I was thinking that we could just do a similar procedure as we do now with the engine with sync and commit, but using two accounts:
1. Consider database on _account 1_ as the source of truth
2. Consider cloud on _account 2_ as the cloud that needs to be updated
3. With the special consideration about different aws_accounts, regions, etc... let the engine run its logic, and perform changes to reflect status of _account 1_ into _account 2_.

yrobla (commented):
added some comments

mtp1376 (commented):
I read the RFC and the proposed option seems better than the alternatives for this purpose. But TBH I think we can defer implementation of this functionality unless we need it in the near future.

rfcs/006 - Sync changes across databases.md
depombo: I am not sure if this comment was from after or before the sprint planning meeting. @dfellis point to this was that every module will have different edge cases outside of just the different `@cloudId`s in _account 1_ and _account 2_ respectively. If you also factor likely differences in the capacity of resources and values of secrets across staging and production, since different teams setup staging and prod differently, it would make this process extremely error-prone.

depombo: > I was thinking that we could just do a similar procedure as we do now with the engine with sync and commit, but using two accounts:
> 
> 1. Consider database on _account 1_ as the source of truth
> 2. Consider cloud on _account 2_ as the cloud that needs to be updated
> 3. With the special consideration about different aws_accounts, regions, etc... let the engine run its logic, and perform changes to reflect status of _account 1_ into _account 2_.

btw this alternative is basically #2 in the RFC, but lmk if I am misinterpreting something and I can list it as a separate alternative

depombo:
> I read the RFC and the proposed option seems better than the alternatives for this purpose. But TBH I think we can defer implementation of this functionality unless we need it in the near future.

Per the sprint planning and proposal, having proposed workflow (even if it is manual) for porting staging changes to production is something we need sooner rather than later

aguillenv (approved):

rfcs/006 - Replicate changes between staging and prod RFC.md
```diff
@@ -0,0 +1,87 @@
...
+The steps to replicate recent database/account A changes in database/account B when calling `iasql_import('A')` from within database B:
+1) Regular sync database A to cloud account A
+2) Regular sync database B to cloud account B
+3) New import sync functionality where database B is pointed to cloud account A directly
+
+The new import sync functionality described in #3 would expand the mapper definitions to include a new `import` function and laxer equality function for each module.
+
+### 3. Interactive `git rebase` style sync
+
+Similar to option 2, but when conflict arises prompt the user to handle the merge conflicts between the two cloud accounts or databases interactively. The PostgreSQL REPL doesn't allow interactive input so that would have to happen using a two-phase process via an interim table or the dashboard. This option requires a lot of resources to implement out of the gate. The current proposal and if this workflow proves to be primordial to users we can eventually prioritize it.
```
dfellis, line 67:
```suggestion
Similar to option 2, but when conflict arises prompt the user to handle the merge conflicts between the two cloud accounts or databases interactively. The PostgreSQL REPL doesn't allow interactive input so that would have to happen using a two-phase process via an interim table or the dashboard. This option requires a lot of resources to implement out of the gate. The current proposal and this workflow are compatible, so if this proves to be important to users we can eventually prioritize it.
```

rfcs/006 - Replicate changes between staging and prod RFC.md
```diff
@@ -0,0 +1,87 @@
...
+### 2. Sync a database to a different cloud account than the one it is connected to
+
+An IaSQL function that takes the alias of another dataabse owned by the same user as input. `iasql_import(‘staging’)` will return the same data structure as `iasql_commit` or `iasql_preview`. This function can reuse parts of the mappers and existing IaSQL functions to run diffs across two different databases or accounts. There could have an optional `dry_run` parameter or another function to preview the changes: `iasql_preview_import(‘staging’)` or `iasql_import(‘staging’, true)`.
+
+The steps to replicate recent database/account A changes in database/account B when calling `iasql_import('A')` from within database B:
+1) Regular sync database A to cloud account A
+2) Regular sync database B to cloud account B
+3) New import sync functionality where database B is pointed to cloud account A directly
+
+The new import sync functionality described in #3 would expand the mapper definitions to include a new `import` function and laxer equality function for each module.
```
dfellis, line 63: It would be good to mention in here that the AWS-provided cloud IDs would make identifying "equivalent" records from one cloud to fill in for the other very difficult, if not impossible.

Also mentioning that differences in the way deployment is done in production vs staging to deal with real-world concerns like not overspending on a staging environment but being able to handle the workload of production would cause further differences that would be impossible to identify as "valid" differences between the two.

dfellis (approved):

rfcs/006 - Replicate changes between staging and prod RFC.md
```diff
@@ -0,0 +1,87 @@
...
+
+### 3. Interactive `git rebase` style sync
+
+Similar to option 2, but when conflict arises prompt the user to handle the merge conflicts between the two cloud accounts or databases interactively. The PostgreSQL REPL doesn't allow interactive input so that would have to happen using a two-phase process via an interim table or the dashboard. This option requires a lot of resources to implement out of the gate. The current proposal and this workflow are compatible, so if this proves to be important to users we can eventually prioritize it.
+
+### 4. Use existing PostgreSQL data synchronization tools
+
+There is a breadth of tools to keep two databases in sync. However, they are not enough to satisfy our requirements for this problem because the data should not be copied over exactly as-is. We have cloud-generated IDs as columns in a lot of tables that are specific to a resource in each account like an EC2 instance id.
```
mtp1376, line 71: I some kind of "cross-database equality check for entities" would help us. I don't know of any yet, but we can think about it.

mtp1376 (approved):
One last question: are we just including rows that are directly created by the user through the DB, or we're also going to add the `sync`ed records?

rfcs/006 - Replicate changes between staging and prod RFC.md
depombo: I'm not sure I understand what you mean here or how it would help use sidestep what is highlighted

depombo:
> One last question: are we just including rows that are directly created by the user through the DB, or we're also going to add the `sync`ed records?

I believe it will be both since they go in the audit log too

dfellis:
> > One last question: are we just including rows that are directly created by the user through the DB, or we're also going to add the `sync`ed records?
> 
> I believe it will be both since they go in the audit log too. @dfellis or @aguillenv to confirm

It would have to be both or a legitimate use case where they created the infra by hand with the AWS console and then want to rapidly apply it in another environment wouldn't work.

rfcs/006 - Replicate changes between staging and prod RFC.md
mtp1376: I mean we need a way to check if two entities are the same thing, but in two different environments (in staging vs in production).
Currently we use `@cloudId` which won't work across accounts, but maybe there's a way? Like tracking the resources and self-assigning them IDs. Anyway this needs more thought.

dfellis: I'm pretty sure it's impossible to do that for some of the reasons laid out in here. Not only are some of them ambiguous (how to distinguish multiple EC2 instances with no `Name` tag) but some may be conceptually identical for the user and how they're using it, but actually different between environments for cost reasons.

yrobla (approved):

PR #1682, opened 2022-12-15
Add `import_source_credential` RPC - use CodeBuild in `ecr_build` RPC

mtp1376:


src/modules/aws_codebuild/rpcs/import_source_credential.ts
```diff
@@ -0,0 +1,82 @@
...
+  } as const;
+
+  /** @internal */
+  importSourceCredentials = crudBuilderFormat<CodeBuild, 'importSourceCredentials', string | undefined>(
+    'importSourceCredentials',
+    input => input,
+    res => res?.arn,
+  );
+
+  validServerTypes = ['BITBUCKET', 'GITHUB', 'GITHUB_ENTERPRISE'];
```
yrobla, line 26: this could be an enum

src/modules/aws_codebuild/rpcs/import_source_credential.ts
```diff
@@ -0,0 +1,82 @@
...
+  validAuthTypes = ['BASIC_AUTH', 'OAUTH', 'PERSONAL_ACCESS_TOKEN'];
+
+  private makeError(message: string) {
+    return [
+      {
+        arn: '',
+        status: 'ERROR',
```
yrobla, line 33: we should unify at some point. I return OK/KO as there were other rpcs doing that. At some point we need to have a common standard for the returns of those functions

src/modules/aws_codebuild/rpcs/import_source_credential.ts
```diff
@@ -0,0 +1,82 @@
+import { Build, CodeBuild, ImportSourceCredentialsInput } from '@aws-sdk/client-codebuild';
+import { ServerType } from '@aws-sdk/client-codebuild/dist-types/models/models_0';
+
+import { AWS, crudBuilderFormat } from '../../../services/aws_macros';
+import { Context, RpcBase, RpcResponseObject } from '../../interfaces';
+import { AwsCodebuildModule } from '../index';
+
+export class ImportSourceCredentialRpc extends RpcBase {
```
yrobla, line 8: can you add typedoc documentation as i did for other rpcs? so we keep the documentation up to date

src/modules/aws_codebuild/rpcs/import_source_credential.ts
mtp1376: Agreed. Let's raise it in the channel to discuss furtuer.

examples/ecs-fargate/django/app/infra/migrations/0001_install.py
```diff
@@ -14,12 +14,12 @@ class Migration(migrations.Migration):
         migrations.RunSQL(
             """
             SELECT * FROM iasql_install(
-                'aws_ecs_simplified'
+                'aws_ecs_simplified', 'aws_codebuild'
```
aguillenv, line 17: why is this added if it is not in use?

src/modules/aws_ecr/rpcs/build.ts
```diff
@@ -164,31 +89,44 @@ export class EcrBuildRpc extends RpcBase {
...
-    awsIamModule.role.cloud.delete(role, ctx);
+    await awsIamModule.role.cloud.delete(role, ctx);
 
     // delete credentials
-    if (githubPersonalAccessToken) await this.deleteGithubCredentials(credentialsArn, client);
+    if (githubPersonalAccessToken)
+      await awsCodebuildModule.sourceCredentialsList.cloud.delete(
+        { arn: credentialsArn } as SourceCredentialsList, // hacky but works
+        ctx,
+      );
```
aguillenv, line 126: can we add the braces here to avoid confusion?

src/modules/aws_ecr/rpcs/build.ts
```diff
@@ -140,18 +56,27 @@ export class EcrBuildRpc extends RpcBase {
     githubRef: string,
     githubPersonalAccessToken: string,
   ): Promise<RpcResponseObject<typeof this.outputTable>[]> => {
-    await this.ensureAwsIamModule(_dbId);
+    await this.ensureModules(_dbId);
```
aguillenv, line 59: can we use `dbId` instead of `_dbId`? Let's follow the convention that when the parameter is prefixed with `_` is because it is not in use.

src/router/db.ts
```diff
@@ -249,57 +249,56 @@ async function rpcCall(
   postTransaction?: PostTransactionCheck,
 ): Promise<any[] | undefined> {
   let rpcRes: any[] | undefined;
-  let errorOcurred = false;
-  let finallyErr;
+  let rpcErr;
```
aguillenv, line 252: well, changes in this file should not be here since they have not been approved/merged. I need to fix a new bug that David spot deploying to staging so that will delay the changes.

src/modules/aws_ecr/rpcs/build.ts
mtp1376: Can you clarify this a bit?

src/modules/aws_ecr/rpcs/build.ts
mtp1376: Sure.

src/modules/aws_ecr/rpcs/build.ts
mtp1376: If you mean parentheses, the linter will remove them once added.

examples/ecs-fargate/django/app/infra/migrations/0001_install.py
mtp1376: https://github.com/iasql/iasql-engine/pull/1682/files#diff-8818acbe0b7c48eb9a08c6d295a3de62882605533c5ebab4ed6b23bab3315f4aR149
It is used in `ecr_build` therefore we should install `aws_codebuild` to use it.

src/router/db.ts
mtp1376: I checked that file out from `main` branch.

src/modules/aws_codebuild/rpcs/import_source_credential.ts
mtp1376: Added.

dfellis (approved):

examples/ecs-fargate/django/app/infra/migrations/0001_install.py
aguillenv: oh, I see. Any reason to go with this pattern and not a hard dependency on the module?

aguillenv (approved):

examples/ecs-fargate/django/app/infra/migrations/0001_install.py
mtp1376: The module works without those modules installed (read/update/create/delete), so I thought we don't need to add the overhead to sync/apply by adding another module. Just the RPC depends on it so added the soft-dependency.

PR #1686, opened 2022-12-16
Better rpc error handling

### LINKED ISSUE #1699: Better RPC error handling

aguillenv:
Resolves #1699


src/router/db.ts
```diff
@@ -249,57 +249,56 @@ async function rpcCall(
...
+        await conn.query(`select * from iasql_commit();`);
+      } catch (e) {
+        await iasqlFunctions.closeTransaction(conn);
+        commitErr = e;
+      }
+      break;
   }
-  if (finallyErr) throw finallyErr;
+  if (rpcErr) throw rpcErr;
+  if (commitErr) throw commitErr;
```
dfellis, line 301: Do we want a combined error if both are defined, or is that not really possible?

aguillenv: I thought about it but not sure. It is possible because it will depend on what is defined as a post-transaction check. But it is pretty probable that if both occurred, the root cause of the `commitErr` would be the RPC failure

dfellis: Got it

dfellis (approved):

mtp1376 (approved):

src/router/db.ts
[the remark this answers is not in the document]
aguillenv: Since both reviews raised the same concern I'll combine error messages @dfellis @mtp1376

dfellis (approved):

PR #1687, opened 2022-12-17
RFC 007 - Peer review infra changes

depombo:


depombo:
actually I need to dive deeper. dbmate might have everything we need https://github.com/amacneil/dbmate#alternatives

depombo:
> actually I need to dive deeper. dbmate might have everything we need https://github.com/amacneil/dbmate#alternatives

They don't support baselining, but the project seems active. I created an issue, let's see what they say! https://github.com/amacneil/dbmate/issues/348

rfcs/007 - Code review infra changes.md
```diff
@@ -0,0 +1,60 @@
...
+## Proposal
+
+The proposal is to implement a super lightweight, open-source plain SQL migration system in its own public repository, such that it could be used separately without IaSQL. The feature set for the initial MVP version:
+- A downloadable CLI with a single `migrate` command that works with Linux, so it is compatible with CI/CD
+- Support PostgreSQL 14
+- Support `up` versioned migrations with the following format `migrations/V0__********.sql`, `migrations/V1__********.sql`
+- Be able to baseline migrations from a given version so running `migrate` ignores migrations before a given version
+- Three config values are provided to `migrate` through environment variables: baseline version (defaults to 0), migration folder location (defaults to `migrations/` in the current directory), and a required PostgreSQL connection string
+
+## Alternatives Considered
```
dfellis, line 46: I am starting to get an idea for an alternative in my mind that may supplement or supplant the main proposal. Stay tuned for a big comment in a couple of hours :)

dfellis: So one of the advantages of IaSQL is that when you write statements `INSERT`, `UPDATE`, or `DELETE` something, the database can error back at you telling you you're missing some required column, or that you're trying to set a column to an invalid value, or you're trying to remove something that still has something dependent on it.

It's a level even above what Pulumi is capable of in typed languages because the "type safety" can extend to particular records/values actually in your cloud account that just static typing would be unable to provide.

This "super power" is completely lost in a traditional migration system because the file is static and unrelated to your database. You write SQL into a migration file and you can only confirm that it will actually work by running it (against a test database or the real thing).

It does give you the ability to let others review the change and potentially sign off on it or not, but beyond their own ability to spot typos, it is unlikely they can catch any errors in what you wrote, beyond "does this seem to implement what we're trying to accomplish in our infrastructure."

But if we put the database in the loop, particular IaSQL, we could make something even better.

Using what we're thinking of in RFC 006, where we can take a set of audit log records and turn them into SQL, users could run `SELECT iasql_begin();`, then write their proposed changes in the dashboard or their preferred SQL editor of choice, and could then use the proposed `iasql_get_sql_since()` or perhaps some other function that automatically grabs the current IaSQL transaction's changes and get the SQL statements to produce the changes they are proposing, with any mistakes they made along the way automatically filtered out because those statements would not have mutated any of the tables, finally the user can `iasql_preview()` to see that the changes do what's expected of them and `iasql_rollback()` to revert the changes and drop the transaction. The generated SQL can be copy-pasted into their migration system of choice for external review, but with much more confidence that it would actually work.

But we can do better than that! Since we are already in the IaSQL database for authoring a change that should be reviewed, there's no reason why we can't have helper functions to handle all of this.

`iasql_review_changes()` would grab the audit log records starting from the current transaction, generate the SQL, generate the execution preview, and save all of that into a new table (that would only need three columns `id: int, sql: text, expected_operations: json`), then it would `iasql_revert()` to before the transaction started and return the ID as the output.

Others can call `iasql_view_changes(id)` to get the change in question and see it displayed in their own SQL connection, and then you can either `iasql_reject_changes(id)` (which would delete the row from the table) or `iasql_approve_changes(id)` (which would start a new transaction, run the SQL, then `iasql_commit()` it).

We could potentially make a fifth function that bakes in Github support `iasql_pr_changes(id, gh_repo, gh_pat)` to auto-generate a github PR with the description being a markdown version of the `iasql_view_changes` output, and merging the PR *somehow* securely calls `iasql_approve_changes(id)`, but not sure how much extra work that would be, and there's a bunch of weirdness around that PR getting out-of-sync with the database if someone calls `iasql_reject_changes(id)` directly in the database and someone else later merges the PR, so I'm less into that one as written (but perhaps it could be refined into something better, and maybe even inclusive to Gitlab and others?)

In any case, the original 4 proposed functions would just be 4 pure SQL functions with 1 table, and could be in a new `iasql_review` module that users can optionally install if they want that sort of thing, which would only depend on the other `iasql_*` modules.

mtp1376: I also had this idea of "running migrations through IaSQL and not pure Postgres" since if we go with Postgres, there will be some complexities. And Luisfer had proposed the "baselining" requirement since if we don't baseline migrations, we might run into errors on `INSERT`s (if the object already exists - for example) which is one of the complexities.
So the idea of "running migrations through IaSQL helper functions" comes into the mind. I was thinking of a way to *somehow* know if an `INSERT` should be executed or it has already been executed. That'd be possible because we keep track of changes we push/pull to/from the cloud and we might be able to see if a change has already been pushed to the cloud. Which opens our hand to appropriately handle the situation (which if the rows are directly `INSERT`ed to the Postgres, isn't possible).

But the problem in my mind was this: having a function like `run_iasql_migration('INSERT INTO bucket () VALUES ()')` is an ugly method and providing an interface that treats SQL queries like strings is a bad idea.

David's proposed idea is a good one and eliminates the ugliness I mentioned above. Also we can programmatically handle approvals for a set of changes (and I suggest we also add multiple users for the same database - e.g. `david` and `mohammad` as two engineers maintaining an infrastructure). And changes can be approved if at least a number of users have approved them.

I don't see any downside for "running migrations through IaSQL helper functions" and not directly.

depombo:
Putting this on hold for now until we have more data on disconnect and connect rates from usage.

depombo:
Will re-open if/when the time comes

aguillenv:
Thinking out loud, maybe adding another special set of commands, similar to iasql_begin/iasql_commit, that creates a "sandbox" where you can do all the queries you want and then the close/return command gives you the generated SQL during that state so you can then submit those queries?

depombo:
> Thinking out loud, maybe adding another special set of commands, similar to iasql_begin/iasql_commit, that creates a "sandbox" where you can do all the queries you want and then the close/return command gives you the generated SQL during that state so you can then submit those queries?

Hmm I am not sure about adding more commands, but perhaps 🤔. Not for the MVP, but eventually we should also need to allow multiple WIP transactions per user and across users. What you are mentioning would solve that without changing the current workflow too

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+- works with `iasql_upgrade`
+- works with any source control solution without the need for special access to it
+- preserves the ability to `iasql_preview` changes in development and get the type-safety feedback from the dashboard editor
+
+The workflow simply suggests users run `iasql_begin`, perform all the necessary changes, and before doing `iasql_commit` they manually put the output of `iasql_preview` and the SQL generated thus far into a markdown file within a repository of their liking. Once the PR is approved by their peers and landed, they can manually run `iasql_commit`. We can require a parameter for `iasql_commit(message)` which is a string attached as metadata for the transaction that can be set to be the PR URL. Best practices are thus enforced using convention and not configuration with a complete reference of what was done via the IaSQL audit log if necessary.
+
+The required work in the IaSQL engine is associating entries in the audit log with specific transactions and adding a `get_sql_for_transaction()` that returns the SQL for the current transaction so users don't have to provide a timestamp for `get_sql_since()`.
```
aguillenv, line 50: Also, let the transactions live forever.

aguillenv (commented):
I must say I like the simplicity of the proposal. For starters convention over configuration makes sense to me and seems a clean approach, at least for now.

We would keep transactions as they are working, we do not overcomplicate with multi-users, and it will be just one transaction at a time so we do not have conflicts. The only "cons" I see are that it can create a bit of noise in the audit logs, but not an issue if `iasql_rollback` also accepts a `comment` and we can let know why a transaction was not committed.

depombo:
> The only "cons" I see are that it can create a bit of noise in the audit logs, but not an issue if `iasql_rollback` also accepts a `comment` and we can let know why a transaction was not committed.

I think this is hard to avoid even outside the context of this proposal unless we introduce a conditional that sends changes somewhere else which as we discussed offline. Once we associate changes with transactions in the audit log we can filter by transactions as opposed to "staged" changes which should make this less of a concern.

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+- Allow plain SQL or have an introspective ORM
+- Free and ideally open source
+- Baseline/ignore certain migrations
+- Be resilient to `iasql_ugprade`
+
+Data migrations require being able to baseline migrations to avoid duplicate record violations on `iasql_upgrade`. Flyway comes closest to meeting all our requirements. However, it is not possible to baseline/ignore existing migrations on their free product.
+
+We could configure `iasql_upgrade` to preserve a `migrations` table and make the name of the table configurable, but the baselining issue would remain as `iasql_upgrade` can change the schema which renders existing migrations on past IaSQL versions invalid. As a result, we need an opinionated solution to unblock peer review of infra changes.
+
+## Proposal
```
aguillenv, line 40: Thinking about it, a next step that can follow up this proposal could be adding the GitHub module, not to lock the process to a third-party vendor as GitHub but to demonstrate we have the ability to create this kind of module for third parties and can be integrated into the convention. For example, we can `get_sql_for_transaction` and then `gh_create_pr` or something and they can automate that step instead of adding the content and the preview directly from the IaSQL dashboard (without the approval process integration).

depombo: Yeah, if it is automating the convention and not tightly integrated with the transaction workflow I think that is a great idea.

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
+# 007 - Peer review infrastructure changes RFC
+
+### Proposed
+
+2022-12-16
```
dfellis, line 5: Should the proposal date be changed since this is a revival?

rfcs/006 - Replicate changes between staging and prod RFC.md
```diff
@@ -39,7 +39,7 @@ Important requirements to keep in mind are being able to:
...
+The IaSQL audit log table is timestamped and looking at the records from a given point in time for a staging database will give users most of the information they need to be able to replicate the changes in production. This proposal is grounded on that fact and seeks to make that workflow as smooth as possible by creating a feature that generates SQL from the audit log from a given point in time. The given point in time for this use case would be the last time staging and production were synced. The generated SQL will then be massaged and manually run by the user in production. The IaSQL audit log is ordered chronologically and the SQL generated by a `iasql_get_sql_since()` function should not cause duplicate records or FK relation problems even if the expectation is that the generated SQL should be massaged.
```
dfellis, line 42: I don't know if we should edit past RFCs like this even if we have changed our perspective in the meantime.

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+- Luis Fernando De Pombo <l@iasql.com>
+
+## Summary
+
+Eng teams move over from using cloud UIs to IaC because cloud infrastructure changes can lead to outages more often than business logic changes. As a result, it is considered a DevOps best practice to code review infrastructure changes using IaC tools and show the `preview` of the resulting change in pull requests to raise the quality bar and reduce the possibility of mistakes. Cloud infrastructure cannot be code reviewed with IaSQL using migration systems even if they support non-schema, or data, migrations (some ORMs like Prisma only support schema changes), and each migration is wrapped in an IaSQL transaction. Through our research and dogfood efforts, we have not found a migration system that fulfills the requirements needed to code review changes with IaSQL:
+- Allow data-only, `up` migrations
+- Allow plain SQL or have an introspective ORM
+- Free and ideally open source
+- Baseline/ignore certain migrations
+- Be resilient to `iasql_ugprade`
```
dfellis, line 34: This should be changed to not reference the `iasql_upgrade` command but just about upgrades to the IaSQL engine and schema.

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+
+Data migrations require being able to baseline migrations to avoid duplicate record violations on `iasql_upgrade`. Flyway comes closest to meeting all our requirements. However, it is not possible to baseline/ignore existing migrations on their free product.
```
dfellis, line 36: Same comment here wrt `iasql_upgrade`, just note that when IaSQL upgrades its schema it reconstructs the entire database so migration data will get dropped and therefore the migration system needs to be able to deal with duplicate record violations.

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+
+We could configure `iasql_upgrade` to preserve a `migrations` table and make the name of the table configurable, but the baselining issue would remain as `iasql_upgrade` can change the schema which renders existing migrations on past IaSQL versions invalid. As a result, we need an opinionated solution to unblock peer review of infra changes.
```
dfellis, line 38: Need to rephrase this. `iasql_upgrade` as a command no longer exists and the upgrade process is different from before, no longer being in-place within the database but done with construction of a new database and then atomic swapping of them by name.

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Proposal
+
+Take inspiration from [RFC 006](./006%20-%20Replicate%20changes%20between%20staging%20and%20prod%20RFC.md) and create a how-to guide in the documentation much like the guide on [how to replicate changes across infra environments](https://iasql.com/docs/envs/) along with an example repository that walks users through how to code review their changes using **convention over configuration** such that the workflow:
+- works with any cloud identity and access management structure
+- works with `iasql_upgrade`
```
dfellis, line 44: No such thing as `iasql_upgrade`

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+- works with any source control solution without the need for special access to it
```
dfellis, line 45: I don't understand this bullet point. What source control solution are we talking about here?

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+- preserves the ability to `iasql_preview` changes in development and get the type-safety feedback from the dashboard editor
+
+The workflow simply suggests users run `iasql_begin`, perform all the necessary changes, and before doing `iasql_commit` they manually put the output of `iasql_preview` and the SQL generated thus far into a markdown file within a repository of their liking. Once the PR is approved by their peers and landed, they can manually run `iasql_commit`. We can require a parameter for `iasql_commit(message)` which is a string attached as metadata for the transaction that can be set to be the PR URL. Best practices are thus enforced using convention and not configuration with a complete reference of what was done via the IaSQL audit log if necessary.
```
dfellis, line 48: Why call out "markdown file" here? It could just as well be a CSV file since the `iasql_preview` output is a table. I see below a mention of adding an `get_sql_for_transaction` function but it's not mentioned as something to include in the artifact to review. That could be a justification for markdown vs other formats since you can use Github's automatic pretty printing of that file format to your advantage during review, but it's not justified here. 

Also, this flow as described doesn't mention anything about showing the relevant SQL statements but maybe it should? In that case, it may make sense to rework this and the following paragraph.

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+
+The required work in the IaSQL engine is associating entries in the audit log with specific transactions and adding a `get_sql_for_transaction()` that returns the SQL for the current transaction so users don't have to provide a timestamp for `get_sql_since()`. Additionally, we can modify transactions so they are active past 30 minutes and do not expire at a given point in time as users are making changes to the database.
+
+Down the line, we can revisit another alternative for a longer-term implementation and/or automate different parts of the convention for this workflow when we have more data points and resources. It might be as simple as adding a job template in each CI platform that only requires a Postgres connection string and can be triggered manually within the CI/CD to open a PR with the details of the current transaction.
```
dfellis, line 52: If we are going with "convention over configuration" it would be both 1) pretty easy to do (a pure pl/pgsql function if desired) and 2) more "conventional" if there was an `iasql_create_review(title text, description text)` function they call that internally creates the context of the markdown file consisting of the title and description of the change, followed by ` ```sql ` and `<table>` formatted blocks powered by `get_sql_for_transaction()` and `iasql_preview()`

Then they just have to copy-paste this into github or whatever version control system their team uses.

The documentation can describe what it's doing under the hood with these two functions but reduce the friction involved in formatting the output for consumption.

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+
+Pros: 
+- This option does not require implementing multi-user or a GitHub provider. We also have a decent amount of experience building CLIs in Rust.
+- This option requires little user education. UX is similar to what people expect in a traditional database workflow akin to how we converged on transactions over IaC's `apply`.
+- Less surface area for things going wrong since the current transaction workflow does not change.
+
+Cons: 
+- Migration files are static so there is no data model feedback and it is not possible to look at `iasql_preview` as you develop. This can be sidestepped by suggesting users do `iasql_begin`, write SQL, `iasql_preview`, `get_sql_since`, and then `iasql_rollback` at the cost of adding noise to the IaSQL audit log.
+- The user has to remember to baseline migrations after every `iasql_ugprade`.
+- The user has to configure IaSQL to preserve a migrations table on `iasql_upgrade`, but we can set the defaults in both places accordingly.
```
dfellis, line 77: Need to rewrite these to not reference `iasql_upgrade`

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+
+The required work in the IaSQL engine is associating entries in the audit log with specific transactions and adding a `get_sql_for_transaction()` that returns the SQL for the current transaction so users don't have to provide a timestamp for `get_sql_since()`. Additionally, we can modify transactions so they are active past 30 minutes and do not expire at a given point in time as users are making changes to the database.
+
+Down the line, we can revisit another alternative for a longer-term implementation and/or automate different parts of the convention for this workflow when we have more data points and resources. It might be as simple as adding a job template in each CI platform that only requires a Postgres connection string and can be triggered manually within the CI/CD to open a PR with the details of the current transaction.
+
+## Alternatives Considered that don't require multi-user
```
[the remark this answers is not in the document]
dfellis: Eg, the convention we describe could have a variant where, after you generate the file to review, you `iasql_revert` the transaction, and approval/merging of the review PR executes your `get_sql_for_transaction` SQL statements within a transaction automatically via CI, then the "keep transaction alive forever" requirement and "just working on the honor system here" goes away.

rfcs/007 - Peer review infra changes.md
depombo: If the function has no side effect and just outputs a bunch of text that sounds great!

dfellis: It should have zero side-effects (excluding the `cloud.read` calls that `iasql_preview` already would trigger, but those should not ever change anything in the cloud), yes

rfcs/006 - Replicate changes between staging and prod RFC.md
depombo: I am just removing a detail, more than changing something fundamental about the RFC, since we do not recommend using a migration system and the RFC is reference in https://iasql.com/docs/envs. Seems inconsequential, but I can add a superseded note or something

rfcs/007 - Peer review infra changes.md
```diff
@@ -0,0 +1,125 @@
...
+
+The workflow simply suggests users run `iasql_begin`, perform all the necessary changes, and before doing `iasql_commit` they manually put the output of `iasql_preview` and the SQL generated thus far into a markdown file within a repository of their liking. Once the PR is approved by their peers and landed, they can manually run `iasql_commit`. We can require a parameter for `iasql_commit(message)` which is a string attached as metadata for the transaction that can be set to be the PR URL. Best practices are thus enforced using convention and not configuration with a complete reference of what was done via the IaSQL audit log if necessary.
+
+The required work in the IaSQL engine is associating entries in the audit log with specific transactions and adding a `get_sql_for_transaction()` that returns the SQL for the current transaction so users don't have to provide a timestamp for `get_sql_since()`. Additionally, we can modify transactions so they are active past 30 minutes and do not expire at a given point in time as users are making changes to the database.
```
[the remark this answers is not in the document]
depombo: I added it per @aguillenv's request. I see your point. I will remove it from this RFC since to me this does not seem like a requirement for it and seems even somewhat orthogonal.

rfcs/007 - Peer review infra changes.md
depombo: sure

rfcs/007 - Peer review infra changes.md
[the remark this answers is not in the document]
dfellis: There is something to be said about looking at this more closely, though, and how the user-defined transactions and the cron's internal transactions ought to have differing behaviors.

1. If a cron gets "stuck" (which could be due to a crash or a restart) we *don't* want to roll back the state it was operating on, while if a manual `iasql_commit` blows up we probably do.
2. If a transaction is simply "left open" for too long without calling commit or rollback, such as by a user taking a while to figure out how to write the SQL they need to create the infra they want, this is not something that a cron should *ever* do as it's essentially "just" `SELECT iasql_begin(); SELECT iasql_commit()`, so *perhaps* what we need is a `commit_started` audit log record we execute the timeout on, not against the `transaction_started` audit log record?

However if we make the convention be one of "when you're doing a transaction to generate a review, you should roll back your changes when done" then we side-step the 30 minute timeout issue for this PR for all but the "taking too long to write the IaSQL to be reviewed" issue, which may still be enough to need to consider it in this PR, but we could reduce scope by ignoring that for the time being?

rfcs/006 - Replicate changes between staging and prod RFC.md
depombo: discussed offline. we will remove the mention in the docs of the RFC and keep RFC 006 as is keeping the RFC as historical record of our thought processes behind things

rfcs/007 - Peer review infra changes.md
```diff
@@ -31,23 +31,23 @@ Eng teams move over from using cloud UIs to IaC because cloud infrastructure cha
 - Allow plain SQL or have an introspective ORM
 - Free and ideally open source
 - Baseline/ignore certain migrations
-- Be resilient to `iasql_ugprade`
+- Be resilient to IaSQL schema upgrades which involve tables getting dropped and recreated
```
dfellis, line 34: We actually don't drop tables anymore during upgrade, we create a new database and then swap to it.

rfcs/007 - Peer review infra changes.md
```diff
@@ -31,23 +31,23 @@ Eng teams move over from using cloud UIs to IaC because cloud infrastructure cha
...
+The required work in the IaSQL engine is associating entries in the audit log with specific transactions and adding a `get_sql_for_transaction()` that returns the SQL for the current transaction so users don't have to provide a timestamp for `get_sql_since()`. To formalize the convention, it would be pretty easy to do a pure pl/pgsql function `iasql_create_review(title text, description text)` function that internally creates the context of the markdown file consisting of the title and description of the change, followed by `sql` code and <table> formatted blocks powered by get_sql_for_transaction() and iasql_preview() respectively. Then they just have to copy-paste this into GitHub or whatever version control system their team uses. The documentation can describe what it's doing under the hood with these two functions but reduce the friction involved in formatting the output for consumption.
```
dfellis, line 50:
```suggestion
The required work in the IaSQL engine is associating entries in the audit log with specific transactions and adding a `get_sql_for_transaction()` that returns the SQL for the current transaction so users don't have to provide a timestamp for `get_sql_since()`. To formalize the convention, it would be pretty easy to do a pure pl/pgsql function `iasql_create_review(title text, description text)` function that internally creates the context of the markdown file consisting of the title and description of the change, followed by ` ```sql` code and `<table>` formatted blocks powered by get_sql_for_transaction() and iasql_preview() respectively. Then they just have to copy-paste this into GitHub or whatever version control system their team uses. The documentation can describe what it's doing under the hood with these two functions but reduce the friction involved in formatting the output for consumption.
```

rfcs/007 - Peer review infra changes.md
```diff
@@ -74,7 +74,7 @@ Pros:
 Cons: 
 - Migration files are static so there is no data model feedback and it is not possible to look at `iasql_preview` as you develop. This can be sidestepped by suggesting users do `iasql_begin`, write SQL, `iasql_preview`, `get_sql_since`, and then `iasql_rollback` at the cost of adding noise to the IaSQL audit log.
 - The user has to remember to baseline migrations after every `iasql_ugprade`.
```
dfellis, line 76: Still referencing `iasql_upgrade` here

rfcs/007 - Peer review infra changes.md
depombo: the markdown file itself did not like the triple quotes but will wrap `<table>`

rfcs/007 - Peer review infra changes.md
depombo: I didn't catch it because of the typo

rfcs/007 - Peer review infra changes.md
dfellis: But I left my comment that exact way? You can wrap backticks with backticks using a very precise spacing mechanism

dfellis: Updated the suggestion

rfcs/007 - Peer review infra changes.md
```diff
@@ -31,7 +31,7 @@ Eng teams move over from using cloud UIs to IaC because cloud infrastructure cha
 - Allow plain SQL or have an introspective ORM
 - Free and ideally open source
 - Baseline/ignore certain migrations
-- Be resilient to IaSQL schema upgrades which involve tables getting dropped and recreated
+- Be resilient to IaSQL schema upgrades which involve creating a new database and swamping it
```
dfellis, line 34:
```suggestion
- Be resilient to IaSQL schema upgrades which involve creating a new database and swapping it
```

rfcs/007 - Peer review infra changes.md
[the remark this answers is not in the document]
aguillenv: First, for context, it is not a "timeout" per se, what we do is that to *open* a new transaction we check if in the last 30 minutes, there's another *open* transaction or there's a `commit` running (checking a `START_COMMIT` event). If we find one we error saying there's another transaction open, if we don't we insert the new *open* transaction log. The code can be found here: https://github.com/iasql/iasql/blob/main/src/modules/iasql_functions/iasql.ts#L1854L1887

Now, I've been thinking about this. Help me walk through it. What are the possible scenarios if we remove the "timeout" and the engine crashes and/or restarts? I see 2 main scenarios:

1. Container without persistence
In this scenario, it does not matter that much. when the container restarts the data will be gone, with no audit logs and will work as a fresh container, right?

2. Container with persistent data
From this scenario I see 2 possible issues depending on which was the *latest relevant* log:

- `OPEN_TRANSACTION`: In this case, it will be good, there will be an open transaction, and the users can execute `commit` or `rollback` since these 2 methods need an open transaction to work.
- `START_COMMIT`: This case will mean some change might have gone through, but some not. Hard to know. If we try a `commit` or a `rollback` it will try to look if there's an open transaction, which will be there. The "problem" will be that previous changes, applied or not will be ignored since when a `commit` tries to go through again, it will look for changes since the previous `START_COMMIT` which will be the one that got halfway and it may not be new changes, so it will just `sync` and unlock.

Am I missing a case related to the restart/crash I'm not taking into account it could leave things in a locked state? If you have in mind a possible scenario let me know and we can walk through it.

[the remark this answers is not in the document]
aguillenv: First, for context, it is not a "timeout" per se, what we do is that to *open* a new transaction we check if in the last 30 minutes, there's another *open* transaction or there's a `commit` running (checking a `START_COMMIT` event). If we find one we error saying there's another transaction open, if we don't we insert the new *open* transaction log. The code can be found here: https://github.com/iasql/iasql/blob/main/src/modules/iasql_functions/iasql.ts#L1854L1887

Now, I've been thinking about this. Help me walk through it. What are the possible scenarios if we remove the "timeout" and the engine crashes and/or restarts? I see 2 main scenarios:

1. Container without persistence
In this scenario, it does not matter that much. when the container restarts the data will be gone, with no audit logs and will work as a fresh container, right?

2. Container with persistent data
From this scenario I see 2 possible issues depending on which was the *latest relevant* log:

- `OPEN_TRANSACTION`: In this case, it will be good, there will be an open transaction, and the users can execute `commit` or `rollback` since these 2 methods need an open transaction to work.
- `START_COMMIT`: This case will mean some change might have gone through, but some not. Hard to know. If we try a `commit` or a `rollback` it will try to look if there's an open transaction, which will be there. Currently, we are checking in the `commit` function if there's another commit running. In this case that will fail since that commit will never end. If we add a "timeout" or bypass that one, then another "problem" will be that previous changes, applied or not will be ignored since when a `commit` tries to go through again, it will look for changes since the previous `START_COMMIT` which will be the one that got halfway and it may not be new changes, so it will just `sync` and unlock.

Am I missing a case related to the restart/crash I'm not taking into account it could leave things in a locked state? If you have in mind a possible scenario let me know and we can walk through it.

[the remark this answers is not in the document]
dfellis: So this issue can affect even containers without persistence in the following way:

1. `iasql_commit` is called that has triggered a lot of work.
2. That RPC call is an HTTP req under the hood to the engine, so if it was triggered by the user it apparently has a 15 minute limit from the browser, and a [1 hour limit](https://github.com/iasql/iasql/blob/main/src/modules/interfaces.ts#L554) between Postgres and the engine.
3. If the commit takes longer than an hour *and then fails* in a semi-safe way (not doing what we expected, but not crashing the process), returning an error message back up, there is nothing monitoring that error message anymore. 
4. Now the `START_COMMIT` has no `END_COMMIT` and never will, so if we just wait forever on this point, no cron job will ever run again, and the current behavior of `iasql_commit` and `iasql_rollback` may or may not function in this case, so it may not even be possible to get out of this situation.

Because the cron job uses the same audit log event types as the user, we can't be sure if we should automatically try again (if it was from the cron) or if we should wait for user input (if it was from the user). The *reason* why we shouldn't always rely on input from the user to get things un-stuck is if the user does not use transactions at all and is using it solely as an eventually-consistent database. The user may not even know the transaction option is possible or how it works under the hood, so expecting the user to get things un-stuck seems unreasonable to me.

Now, on your other point with `OPEN_TRANSACTION` but no `START_COMMIT`, this piece is *clearly* only possible if it was initiated by the user, so blocking forever in *this* state seems perfectly fine to me.

The conundrum that I see, though, is if it gets stuck for too long in `START_COMMIT` state; if it was via `cron` the proper solution is to just try again and not try to roll back the user's changes, but if it was a user-initiated transaction, catching this and automatically initiating a rollback seems more reasonable.

If we can't figure out how to distinguish between these two states, or if we don't want to (which could also be reasonable), then "compromising" by always triggering a rollback in this sort of stuck state seems less bad; the eventually consistent user would see that their change never went through, try again, and hopefully it then works, while the user that uses transactions would get the error back that the commit failed and that it rolled back for that reason and then they can decide if they want to try again or not.

But do we have these two different states right now? I thought we only had the beginning and end of the transaction, and no event on when a commit or rollback has started.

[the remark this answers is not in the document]
dfellis: Ideally we can use the same solution in all situations and I was imagining a full revert, but simply getting out of the broken state with a sync could also be fine -- perhaps if there's a `START_COMMIT` without an `END_COMMIT` after some timeout that triggers the cron job to execute a `sync` instead of a `commit`?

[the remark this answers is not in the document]
aguillenv: yes, I can see that, like if the `START_COMMIT` has been open for I don't know, 5hrs, call `sync` instead? and then we can remove the logic for the open transaction. The conditional would be if it is open and how long have been "commiting".

mtp1376 (approved):
I think our current setup is overfitted on one-active-trasaction-at-a-time design. I was thinking if we could have multiple transactions (distinguished with their names), then it'd be possible to keep the current transaction that's going to be reviewed in-progress while doing other cloud modifications as well. But not sure how that'd fit @aguillenv @dfellis

aguillenv (approved):

aguillenv:
> I think our current setup is overfitted on one-active-trasaction-at-a-time design. I was thinking if we could have multiple transactions (distinguished with their names), then it'd be possible to keep the current transaction that's going to be reviewed in-progress while doing other cloud modifications as well. But not sure how that'd fit @aguillenv @dfellis

I've thought about that before, but not sure if it is what we want, though. The idea behind transactions is to have explicit control of what you are doing since the changes in cloud infra can be quite critical. Having multiple transactions could lead to "merge conflicts" between the 2 transaction states that could be hard to manage.

dfellis:
> > I think our current setup is overfitted on one-active-trasaction-at-a-time design. I was thinking if we could have multiple transactions (distinguished with their names), then it'd be possible to keep the current transaction that's going to be reviewed in-progress while doing other cloud modifications as well. But not sure how that'd fit @aguillenv @dfellis
> 
> I've thought about that before, but not sure if it is what we want, though. The idea behind transactions is to have explicit control of what you are doing since the changes in cloud infra can be quite critical. Having multiple transactions could lead to "merge conflicts" between the 2 transaction states that could be hard to manage. But I do not discard going into that direction eventually

On top of that, we could only do that if we actually had IaSQL transactions as *actual* transactions where the state is separate between the two of them; doable, but we might have to switch to being a native Postgres extension (in the best case scenario) or an actual fork of Postgres to make that happen.

dfellis (approved):

PR #1693, opened 2022-12-19
fix: add sidebar by default on SQL reference

yrobla:

### DIFF
--- a/site/docusaurus.config.js
+++ b/site/docusaurus.config.js
@@ -66,7 +66,7 @@ const theButton = require('./src/rehype/thebutton');
           // other options are passed to posthog-js init as is
         },
       ],
-      ['docusaurus-plugin-iasql', {}],
+      ['docusaurus-plugin-iasql', { usedSidebar: 'docs' }],
     ],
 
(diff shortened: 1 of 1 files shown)


site/docusaurus.config.js
```diff
@@ -66,7 +66,7 @@ const theButton = require('./src/rehype/thebutton');
           // other options are passed to posthog-js init as is
         },
       ],
-      ['docusaurus-plugin-iasql', {}],
+      ['docusaurus-plugin-iasql', { usedSidebar: 'docs' }],
```
aguillenv, line 69: what's the result of this change?

yrobla: the problem was that when you were inside a doc, it didn't have any sidebar. Now it will display just the current sidebar as all the documents.

aguillenv: oh, I see, I've reproduced it

aguillenv (approved):

dfellis (approved):

mtp1376 (approved):

PR #1696, opened 2022-12-20
Revert some suspicious commits

dfellis:
Staging is currently busted, reverting some changes since the last time I manually deployed it to see what's going on


yrobla (approved):

aguillenv:
@dfellis the problem is within the `RPC transaction wrapper`, not the other 2. I've triggered it with them and it works.

aguillenv:
> @dfellis the problem is within the `RPC transaction wrapper`, not the other 2. I've triggered it with them and it works. Can we revert the revert for those?

aguillenv (approved):

dfellis:
We finally figured out that it was a backwards-incompatible schema change, but this won't affect production, it was only stale test databases in staging that were using a schema that will never be released that were failing.

aguillenv:
For more context, the new default behaviour when no RPC pre-transaction check is to call `iasql_begin`. The older databases' RPCs do not have checks specified so they fall into the default behaviour making `iasql_begin` call `iasql_begin` indefinitely.

Issue #1669, opened 2022-12-14
Add complete support for User Access Keys and User profiles


yrobla:
Right now, to delete an IAM user we delete the keys and the related user profiles manually. Same with user policies. We should add complete support for all that related entities

PR #1697, opened 2022-12-20
feat: add full support for aws access keys

yrobla:
Closes: #1669


src/modules/aws_iam/mappers/access_key.ts
```diff
@@ -0,0 +1,148 @@
...
+  cloud = new Crud2<AccessKey>({
+    create: async (es: AccessKey[], ctx: Context) => {
+      // Do not cloud create, just restore database
+      await this.module.accessKey.db.delete(es, ctx);
+    },
+    read: async (ctx: Context, id?: string) => {
+      const client = (await ctx.getAwsClient()) as AWS;
+
+      if (id) {
+        const { keyId, user } = this.idFields(id);
```
aguillenv, line 68: I do not see `user` decorated with `@cloudId`, how is this working?

aguillenv (commented):
the RPC index is not exporting the fn, also, can we add some tests for the RPC call?

src/modules/aws_iam/mappers/access_key.ts
yrobla: yep, an accidental removal... i will readd

src/modules/aws_iam/entity/access_key.ts
```diff
@@ -0,0 +1,90 @@
...
+    enum: accessKeyStatusEnum,
+  })
+  status?: accessKeyStatusEnum;
+
+  /**
+   * @public
+   * Reference to the user for this record
+   */
+  @ManyToOne(() => IamUser, {
+    cascade: true,
```
aguillenv, line 83: also, cascade here would not trigger the creation of the user for every access_key created?

yrobla:
> the RPC index is not exporting the fn, also, can we add some tests for the RPC call?

there are tests for the RPC call .. the access_key_request() ones

aguillenv:
> > the RPC index is not exporting the fn, also, can we add some tests for the RPC call?
> 
> there are tests for the RPC call .. the access_key_request() ones

oh sorry, I missed it!

src/modules/aws_iam/entity/access_key.ts
```diff
@@ -79,10 +79,7 @@ export class AccessKey {
    * @public
    * Reference to the user for this record
    */
-  @ManyToOne(() => IamUser, {
-    cascade: true,
-    eager: true,
```
aguillenv: oh, why did you remove the eager too?

yrobla: my bad, i thought it was not needed

aguillenv (approved):

dfellis (approved):

PR #1702, opened 2022-12-20
Fix availability zone selection

### LINKED ISSUE #1700: Flakiness in vpc route table integration
```
insert or update on table "subnet" violates foreign key constraint "FK_89d16ba5682889f8fae7927052c"
```
https://github.com/iasql/iasql-engine/actions/runs/3743011942/jobs/6354724278

The problem is that the test is assuming that every region will have an availability zone with the pattern `<region>a`

aguillenv:
Resolves #1700

### DIFF
--- a/test/modules/aws-vpc-routetable-integration.ts
+++ b/test/modules/aws-vpc-routetable-integration.ts
@@ -135,7 +135,7 @@ describe('RouteTable Integration Testing', () => {
     query(
       `
       INSERT INTO subnet (availability_zone, vpc_id, cidr_block, region)
-      SELECT '${region}a', id, '10.${randIPBlock}.1.0/24', '${region}'
+      SELECT (SELECT name FROM availability_zone WHERE region = '${region}' ORDER BY 1 DESC LIMIT 1), id, '10.${randIPBlock}.1.0/24', '${region}'
       FROM vpc
       WHERE tags ->> 'name' = '${prefix}';
   `,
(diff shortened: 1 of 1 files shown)


test/modules/aws-vpc-routetable-integration.ts
```diff
@@ -135,7 +135,7 @@ describe('RouteTable Integration Testing', () => {
     query(
       `
       INSERT INTO subnet (availability_zone, vpc_id, cidr_block, region)
-      SELECT '${region}a', id, '10.${randIPBlock}.1.0/24', '${region}'
+      SELECT (SELECT name FROM availability_zone WHERE region = '${region}' ORDER BY 1 DESC LIMIT 1), id, '10.${randIPBlock}.1.0/24', '${region}'
```
dfellis, line 138: `Order by 1`?

aguillenv: It will order by the first column that is being returned. In this case, would be equivalent to:
```suggestion
      SELECT (SELECT name FROM availability_zone WHERE region = '${region}' ORDER BY name DESC LIMIT 1), id, '10.${randIPBlock}.1.0/24', '${region}'
```

dfellis: Wtfsql 😅

dfellis (approved):

test/modules/aws-vpc-routetable-integration.ts
mtp1376: That pattern is used a lot in SQL injection because we don't know the column names :))

PR #1705, opened 2022-12-21
Fix pattern picking availability zones in tests

### LINKED ISSUE #1700: Flakiness in vpc route table integration

aguillenv:
Resolves #1700


mtp1376 (approved):
👍🏻

dfellis (approved):

PR #1706, opened 2022-12-21
Rework the release scripts

dfellis:
Now that there's only one active version, a lot of the version-tracking logic is no longer necessary, too, so dropped that, as well.


src/config/production.ts
```diff
@@ -23,7 +23,6 @@ const config: ConfigInterface = {
   sentry: {
     dsn: 'https://e257e8d6646e4657b4f556efc1de31e8@o1090662.ingest.sentry.io/6106929',
     environment: 'production',
-    release: process.env.SENTRY_RELEASE ?? throwError('No Sentry Release defined'),
```
depombo, line 26: why does this go away?

dfellis: I moved it [here](https://github.com/iasql/iasql-engine/pull/1706/files#diff-198a1431d7c5e26ea78bc6e54b34b54e6f8ab342dd48ea11013877e8466a87c5R15) because now there's only one version in the codebase and so there's no longer any ambiguity on what version to send along to Sentry.

package.json
```diff
@@ -23,7 +23,6 @@
     "docker-run": "docker run -e IASQL_ENV=local -p 5432:5432 --name iasql iasql:latest",
     "docker-run-vol": "docker run -e IASQL_ENV=local -p 5432:5432 -v $(dirname $PWD)/db:/var/lib/postgresql/14/main --name iasql iasql:latest",
     "version": "./.version.sh",
-    "drop": "./.drop.sh",
     "develop": "./.develop.sh",
```
depombo, line 26: what does `develop.sh` do? shouldn't it go away as well?

depombo (approved):

PR #1707, opened 2022-12-21
Deploy to production

dfellis:


.github/workflows/production.yml
```diff
@@ -0,0 +1,69 @@
...
+
+  deploy:
+    runs-on: ubuntu-latest
+    needs:
+      - setup-build
+      - build-and-push
+    steps:
+      - name: SSH into Production and Deploy
+        env:
+          STAGING_PEM_B64: ${{ secrets.PRODUCTION_PEM_B64 }}
```
aguillenv, line 55: should we rename the env var?

dfellis: Yep, I missed that in my copy-paste!

dfellis:
```suggestion
          PRODUCTION_PEM_B64: ${{ secrets.PRODUCTION_PEM_B64 }}
```

aguillenv (approved):

depombo (approved):

PR #1713, opened 2022-12-21
Cancel connections before disconnect

### LINKED ISSUE #1737: Intermittent error on disconnect
`QueryFailedError: database "allmodulestest" is being accessed by other users`

aguillenv:
Resolves #1737 

- Remove all connections before disconnect
- Try to hold a transaction to avoid leaving misconfigured resources in the cloud if a disconnect occurs during a commit.


dfellis (approved):

dfellis:
So this change consistently breaks the test suite [here](https://github.com/iasql/iasql-engine/actions/runs/3751983547/jobs/6375781434#step:5:7733) during disconnect, erroring out because it can't get a transaction lock. It may be better if it tries to get a transaction lock and if that fails to still continue with the disconnection?

yrobla (approved):

src/services/iasql.ts
```diff
@@ -85,21 +87,29 @@ export async function connect(dbAlias: string, uid: string, email: string, dbId
...
-      database: db.pgName,
-    });
-    try {
-      await conn2.query(`SELECT * FROM query_cron('unschedule');`);
-      await conn2.query(`SELECT * FROM query_cron('unschedule_purge');`);
-    } catch (e) {
-      /** Do nothing */
-    }
+    // Cancel all connections
+    await conn.query(`REVOKE CONNECT ON DATABASE ${db.pgName} FROM PUBLIC, ${config.db.user}, ${db.pgUser};`);
```
mtp1376, line 95: This line is doing a `REVOKE CONNECT` call on `config.db.user`. Then the code (in `unscheduleJobs`) is trying to connect to the same database using `dbMan.baseConnConfig` which also has its `username = config.db.user`. Am I missing something here or it should always fail?

aguillenv: Good catch! Yes, I've messed up the order. We should try to unscheduled first, then revoke and cancel all ongoing connections.

yrobla (approved):

dfellis (approved):

PR #1714, opened 2022-12-21
Add run detach script

aguillenv:


depombo (approved):

dfellis (approved):
Approved, but can you just `yarn docker-run -d` for the same effect? Or does docker have some weird arg ordering issues?

aguillenv:
> Approved, but can you just `yarn docker-run -d` for the same effect? Or does docker have some weird arg ordering issues?

It does!

dfellis:
Whelp, feel free to merge then, but that sucks. :(

yrobla (approved):

PR #1715, opened 2022-12-22
feat: allow to create/update/delete pass via rpc

yrobla:


dfellis (approved):
Inputs all needing to be strings hurts us here a bit, but that's not the fault of this PR.

src/modules/aws_iam/migration/1671659720979-aws_iam.ts
```diff
@@ -0,0 +1,29 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
dfellis, line 1: Btw, things are failing right now because the old migration was accidentally not deleted.

aguillenv (approved):

PR #1718, opened 2022-12-22
Add Support For `BucketWebsite`

mtp1376:


yrobla (approved):

src/modules/aws_s3/mappers/bucket_website.ts
```diff
@@ -0,0 +1,112 @@
...
+      for (const e of es) {
+        await this.saveBucketWebsite(ctx, e);
+        out.push(e);
+      }
+      return out;
+    },
+    read: async (ctx: Context, id?: string) => {
+      const out: BucketWebsite[] = [];
+      const buckets =
+        ctx.memo?.cloud?.Bucket && Object.values(ctx.memo?.cloud?.Bucket).length
```
dfellis, line 49: Nit. You don't need the `?.` operator in the second half of this because you can only reach it if `ctx.memo.cloud.Bucket` is not null-ish the whole way through.

Could be:

```suggestion
        ctx.memo?.cloud?.Bucket && Object.values(ctx.memo.cloud.Bucket).length
```

src/modules/aws_s3/mappers/bucket_website.ts
```diff
@@ -0,0 +1,112 @@
...
+          ? Object.values(ctx.memo?.cloud?.Bucket)
+          : await this.module.bucket.cloud.read(ctx);
+
+      if (!!id) {
+        const bucket = buckets.find((b: Bucket) => b.name === id);
```
dfellis, line 54: There should be a conditional if the bucket is not found, perhaps throwing an error that we can't find the expected bucket if the `bucket` variable is undefined?

depombo:
how does the performance compare to exposing the html file on a CDN?

dfellis (approved):

PR #1726, opened 2022-12-23
add repo image size and date

depombo:
Most of this PR revolves around changing the list private repository images client call to the same one used for public repository images which has a different data structure

### DIFF
--- a/src/modules/aws_ecr/entity/repository_image.ts
+++ b/src/modules/aws_ecr/entity/repository_image.ts
@@ -64,6 +64,26 @@ export class RepositoryImage {
 
   /**
    * @public
+   * Date the image was pushed into the repository
+   */
+  @Column({
+    nullable: true,
+    type: 'timestamp with time zone',
+  })
+  pushedAt: Date;
+
+  /**
+   * @public
+   * Size of the image in MB
+   */
+  @Column({
+    nullable: true,
+    type: 'int',
+  })
+  sizeInMB: number;
+
+  /**
+   * @public
    * Reference to the private repository that is containing the image
    */
   @ManyToOne(() => Repository, { nullable: true })
--- a/src/modules/aws_ecr/index.ts
+++ b/src/modules/aws_ecr/index.ts
@@ -38,10 +38,10 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
     if (!image.imageDigest) {
       throw new Error('Invalid repository image');
     }
-    const imageTag = image.imageTags && image.imageTags.length > 0 ? image.imageTags[0] : '<untagged>';
+    const imageTag = image.imageTags?.[0] ?? '<untagged>';
     out.imageDigest = image.imageDigest;
     out.imageTag = imageTag;
-    if (image.imageSizeInBytes) out.sizeInMB = Math.round(image.imageSizeInBytes * 0.000001);
+    if (image.imageSizeInBytes) out.sizeInMB = Math.round(image.imageSizeInBytes / 1024 / 1024);
     if (image.imagePushedAt) out.pushedAt = image.imagePushedAt;
     (out.privateRepository = undefined), (out.publicRepository = undefined);
     if (type === 'private') {
--- a/src/modules/aws_ecr/index.ts
+++ b/src/modules/aws_ecr/index.ts
@@ -115,9 +115,8 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
     );
     if (images && images.imageDetails) {
       for (const image of images.imageDetails) {
-        if (image.imageDigest) {
-          const imageTag = image.imageTags && image.imageTags.length > 0 ? image.imageTags[0] : '<untagged>';
-          const imageId = { imageDigest: image.imageDigest, imageTag };
+        if (image.imageDigest && image.imageTags && image.imageTags.length > 0) {
+          const imageId = { imageDigest: image.imageDigest, imageTag: image.imageTags[0] };
           await this.deleteRepositoryImage(client, [imageId], repository.repositoryName, image.registryId);
         }
       }
@@ -133,9 +132,8 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
     );
     if (images && images.imageDetails) {
       for (const image of images.imageDetails) {
-        if (image.imageDigest) {
-          const imageTag = image.imageTags && image.imageTags.length > 0 ? image.imageTags[0] : '<untagged>';
-          const imageId = { imageDigest: image.imageDigest, imageTag };
+        if (image.imageDigest && image.imageTags && image.imageTags.length > 0) {
+          const imageId = { imageDigest: image.imageDigest, imageTag: image.imageTags[0] };
           await this.deleteRepositoryImage(client, [imageId], repository.repositoryName, image.registryId);
         }
       }
@@ -161,12 +159,7 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
             const rawImage = await this.listRepositoryImages(client.ecrClient, [imageId], decoded[3]);
             if (rawImage?.imageDetails && rawImage.imageDetails[0]) {
               const imageDetail = rawImage.imageDetails[0];
-              imageDetail.imageTags = imageDetail.imageTags
-                ? imageDetail.imageTags.filter(t => t === decoded[1])
-                : [];
-              if (imageDetail.imageDigest) {
-                return await this.repositoryImageMapper(imageDetail, ctx, type, undefined);
-              }
+              return await this.repositoryImageMapper(imageDetail, ctx, type, undefined);
             }
           }
         } else {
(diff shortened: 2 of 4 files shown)


src/modules/aws_ecr/index.ts
```diff
@@ -31,16 +31,18 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
...
 
     // id is generated with imageDigest + tag
-    if (!image.imageId?.imageDigest) {
+    if (!image.imageDigest) {
       throw new Error('Invalid repository image');
     }
-    out.imageDigest = image.imageId.imageDigest;
-    out.imageTag = image.imageId.imageTag ?? '<untagged>';
-
+    const imageTag = image.imageTags && image.imageTags.length > 0 ? image.imageTags[0] : '<untagged>';
```
dfellis, line 41: You may be able to replace this with:

```suggestion
    const imageTag = image?.imageTags?.[0] ?? '<untagged>';
```

src/modules/aws_ecr/index.ts
```diff
@@ -31,16 +31,18 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
...
+    out.imageDigest = image.imageDigest;
+    out.imageTag = imageTag;
+    if (image.imageSizeInBytes) out.sizeInMB = Math.round(image.imageSizeInBytes * 0.000001);
```
dfellis, line 44: Shouldn't this be:

```suggestion
    if (image.imageSizeInBytes) out.sizeInMB = Math.round(image.imageSizeInBytes / 1024 / 1024);
```

dfellis (approved):

src/modules/aws_ecr/entity/repository_image.ts
```diff
@@ -62,6 +62,26 @@ export class RepositoryImage {
...
+
+  /**
+   * @public
+   * Size of the image in MB
+   */
+  @Column({
+    nullable: true,
+    type: 'int',
+  })
+  sizeInMB: number;
```
mtp1376, line 83: Not important for this PR, but maybe we should define a convention for units? (e.g. saving them always in MB)

src/modules/aws_ecr/index.ts
```diff
@@ -115,13 +113,13 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
...
-    if (images && images.imageIds && images.imageIds.length > 0) {
-      await this.deleteRepositoryImage(
-        client,
-        images.imageIds,
-        repository.repositoryName,
-        repository.registryId,
-      );
+    if (images && images.imageDetails) {
+      for (const image of images.imageDetails) {
+        if (image.imageDigest && image.imageTags && image.imageTags.length > 0) {
```
mtp1376, line 118: So the `image.imageTags` should always be set here? Is that a must? I see in the `repositoryImageMapper` it's being treated as possibly-undefined. But here we're relying on its existence. Just a non-blocking question.

src/modules/aws_ecr/index.ts
```diff
@@ -158,9 +156,10 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
           const idRegion = decoded[4];
           if (enabledRegions.includes(idRegion)) {
             const client = (await ctx.getAwsClient(idRegion)) as AWS;
-            const rawImage = await this.getRepositoryImage(client.ecrClient, [imageId], decoded[3]);
-            if (rawImage?.images && rawImage.images[0]) {
-              return await this.repositoryImageMapper(rawImage.images[0], ctx, type, idRegion);
+            const rawImage = await this.listRepositoryImages(client.ecrClient, [imageId], decoded[3]);
```
mtp1376, line 159: Maybe this `decoded` variable can be replaced by using `this.idFields` function? Or can be done as a follow-up refactor task.

src/modules/aws_ecr/index.ts
```diff
@@ -200,14 +190,24 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
...
+                              ctx,
+                              'private',
+                              region,
+                            ),
+                          );
+                        }
+                      } else {
+                        out.push(await this.repositoryImageMapper(imageDetail, ctx, 'private', region));
+                      }
+                    }
```
mtp1376, line 210: It's a bit hard to follow this nested code. Maybe we can wrap it in a function and early-return instead of `if-else` conditions?

test/modules/aws-ecr-integration.ts
```diff
@@ -203,7 +203,14 @@ describe('ECR Integration Testing', () => {
     `,
         (res: any[]) => {
           expect(res.length).toBe(3);
+          expect(res.map(i => i['size_in_mb']).reduce((a, b) => a + b, 0)).toBe(0);
           expect(res.filter(i => i['image_tag'] === '<untagged>').length).toBe(1);
+          const now = new Date();
+          const pushedAt = new Date(res[0]['pushed_at']);
+          expect(pushedAt.getDate()).toBe(now.getDate());
```
yrobla, line 210: isn't there a way to just do it on single line?

yrobla (commented):
added comment

test/modules/aws-ecr-integration.ts
```diff
@@ -203,7 +203,14 @@ describe('ECR Integration Testing', () => {
     `,
         (res: any[]) => {
           expect(res.length).toBe(3);
+          expect(res.map(i => i['size_in_mb']).reduce((a, b) => a + b, 0)).toBe(0);
```
aguillenv, line 206: either we are pushing something wrong to the repo or this check is not working properly, the size should not be 0, I think. If you see [here](https://gallery.ecr.aws/docker/library/hello-world) the image tags tab it says 122MB

src/modules/aws_ecr/index.ts
```diff
@@ -2,7 +2,7 @@ import {
   ECR,
   Repository as RepositoryAws,
   paginateDescribeRepositories,
-  Image as ImageAws,
+  ImageDetail,
```
aguillenv, line 5: is this a superset if `Image`?

src/modules/aws_ecr/index.ts
```diff
@@ -31,16 +31,18 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
...
+    if (!image.imageDigest) {
       throw new Error('Invalid repository image');
     }
-    out.imageDigest = image.imageId.imageDigest;
-    out.imageTag = image.imageId.imageTag ?? '<untagged>';
-
+    const imageTag = image.imageTags?.[0] ?? '<untagged>';
+    out.imageDigest = image.imageDigest;
+    out.imageTag = imageTag;
+    if (image.imageSizeInBytes) out.sizeInMB = Math.round(image.imageSizeInBytes / 1024 / 1024);
```
aguillenv, line 44: Maybe we should make sure `image.imageSizeInBytes` is a number? I think this would do the work
```suggestion
    if (image.imageSizeInBytes) out.sizeInMB = Math.round(+image.imageSizeInBytes / 1024 / 1024);
```

src/modules/aws_ecr/index.ts
```diff
@@ -115,13 +113,13 @@ class RepositoryImageMapper extends MapperBase<RepositoryImage> {
...
     );
-    if (images && images.imageIds && images.imageIds.length > 0) {
-      await this.deleteRepositoryImage(
-        client,
-        images.imageIds,
-        repository.repositoryName,
-        repository.registryId,
-      );
+    if (images && images.imageDetails) {
+      for (const image of images.imageDetails) {
```
aguillenv, line 117: nit:
```suggestion
    for (const image of images?.imageDetails ?? []) {
```

test/modules/aws-ecr-integration.ts
depombo: It is on the same date, but at a different time. The only way I can think of is to get both dates as an ISO string and truncate them to exclude the time, but that would be harder to follow even if it involves less LOC. I could check that `now` is bigger than `pushed_at` in a single line, but that wouldn't be as exhaustive.

src/modules/aws_ecr/index.ts
depombo: not a straight super set, there are some minor differences between the two

test/modules/aws-ecr-integration.ts
depombo: size is 14kB in the link you provided though I am seeing 4kB, but both of those round to 0 in MB. I think it is okay as-is or should do you want to track kB or bytes? That will be too small when the image is in GB

src/modules/aws_ecr/index.ts
depombo: it is typed into a number

src/modules/aws_ecr/index.ts
depombo: seems like more of a nit, can be a follow-up. I think it is okay as-is

src/modules/aws_ecr/entity/repository_image.ts
depombo: I believe it is MB everywhere

test/modules/aws-ecr-integration.ts
aguillenv: ohh I see, then that's ok, let's keep it as is!

aguillenv (approved):

mtp1376 (approved):

PR #1752, opened 2023-01-03
Get sql from audit logs

### LINKED ISSUE #1694: Implement RFC 006 - Generate SQL from audit table

aguillenv:
RFC 006

Resolves #1694 

Query examples:
```sql
SELECT * FROM iasql_get_sql_since();
SELECT * FROM iasql_get_sql_since('2023-01-05T12:00:00');
SELECT * FROM iasql_get_sql_since((now() - interval '5 minutes')::text);
```

Result examples:
```sql

INSERT INTO load_balancer_security_groups (load_balancer_id, security_group_id)
VALUES ((SELECT id FROM load_balancer WHERE load_balancer_arn = 'arn:aws:elasticloadbalancing:us-east-2:257682470237:loadbalancer/app/test/88a5af6043314c2b'), (SELECT id FROM security_group WHERE group_id = 'sg-06204058f6faf6f52' AND region = 'us-east-2'));

INSERT INTO load_balancer (load_balancer_name, load_balancer_arn, dns_name, canonical_hosted_zone_id, created_time, scheme, state, load_balancer_type, subnets, availability_zones, ip_address_type, region, attributes, vpc)
VALUES ('test', 'arn:aws:elasticloadbalancing:us-east-2:257682470237:loadbalancer/app/test/88a5af6043314c2b', 'test-506077814.us-east-2.elb.amazonaws.com', 'Z3AADJGX6KTTL2', '2023-01-05T16:49:16.99+00:00', 'internet-facing', 'active', 'application', '{'subnet-086d1c2cc46bdd254','subnet-0bfe61253a129d77d'}', '{'us-east-2a','us-east-2b'}', 'ipv4', 'us-east-2', '{'{"Key":"access_logs.s3.enabled","Value":"false"}','{"Key":"access_logs.s3.bucket","Value":""}','{"Key":"access_logs.s3.prefix","Value":""}','{"Key":"idle_timeout.timeout_seconds","Value":"60"}','{"Key":"deletion_protection.enabled","Value":"false"}','{"Key":"routing.http2.enabled","Value":"true"}','{"Key":"routing.http.drop_invalid_header_fields.enabled","Value":"false"}','{"Key":"routing.http.xff_client_port.enabled","Value":"false"}','{"Key":"routing.http.preserve_host_header.enabled","Value":"false"}','{"Key":"routing.http.xff_header_processing.mode","Value":"append"}','{"Key":"load_balancing.cross_zone.enabled","Value":"true"}','{"Key":"routing.http.desync_mitigation_mode","Value":"defensive"}','{"Key":"waf.fail_open.enabled","Value":"false"}','{"Key":"routing.http.x_amzn_tls_version_and_cipher_suite.enabled","Value":"false"}'}', (SELECT id FROM vpc WHERE vpc_id = 'vpc-024bfb31f9340a7cc' AND region = 'us-east-2'));

INSERT INTO target_group (target_group_name, target_type, target_group_arn, ip_address_type, protocol, port, health_check_protocol, health_check_port, health_check_enabled, health_check_interval_seconds, health_check_timeout_seconds, healthy_threshold_count, unhealthy_threshold_count, health_check_path, protocol_version, region, vpc)
VALUES ('test', 'instance', 'arn:aws:elasticloadbalancing:us-east-2:257682470237:targetgroup/test/f699ce47a413e7e1', 'ipv4', 'HTTP', 80, 'HTTP', 'traffic-port', true, 30, 5, 5, 2, '/', 'HTTP1', 'us-east-2', (SELECT id FROM vpc WHERE vpc_id = 'vpc-024bfb31f9340a7cc' AND region = 'us-east-2'));

INSERT INTO listener (listener_arn, port, protocol, action_type, load_balancer_id, target_group_id)
VALUES ('arn:aws:elasticloadbalancing:us-east-2:257682470237:listener/app/test/88a5af6043314c2b/ec1b9847eae71ce9', 80, 'HTTP', 'forward', (SELECT id FROM load_balancer WHERE load_balancer_arn = 'arn:aws:elasticloadbalancing:us-east-2:257682470237:loadbalancer/app/test/88a5af6043314c2b'), (SELECT id FROM target_group WHERE target_group_arn = 'arn:aws:elasticloadbalancing:us-east-2:257682470237:targetgroup/test/f699ce47a413e7e1'));

```


yrobla (commented):
maybe add some testing of the different options of the rpc call?

yrobla (approved):

src/modules/iasql_functions/rpcs/iasql_get_sql_since.ts
```diff
@@ -0,0 +1,290 @@
...
+  return queries;
+}
+
+/** @internal */
+async function getValue(
+  tableName: string,
+  k: string,
+  v: any,
+  mbt: { [key: string]: ModuleInterface },
+  orm: TypeormWrapper,
```
mtp1376, line 171: I think a bit of inline documentation and more clear variable naming could help here.

mtp1376 (approved):
There's a good chunk of this I can't easily understand. But the overall implementation looks good to me.

src/modules/iasql_functions/rpcs/iasql_get_sql_since.ts
```diff
@@ -0,0 +1,290 @@
...
+
+/** @internal */
+async function getChangeLogs(limitDate: string, orm: TypeormWrapper): Promise<IasqlAuditLog[]> {
+  const whereClause: any = {
+    changeType: In([AuditLogChangeType.INSERT, AuditLogChangeType.UPDATE, AuditLogChangeType.DELETE]),
+  };
+  if (limitDate) {
+    const castRes = await orm.query(`SELECT try_cast('${limitDate}', NULL::timestamp with time zone);`);
+    const castedValue = castRes?.pop()?.try_cast;
+    if (castedValue === null) throw new Error(`Cannot cast ${limitDate} to timestamp with time zone`);
```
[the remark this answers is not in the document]
aguillenv: that should not happen because try_cast should return null or a value but let me handle that just in case!

src/modules/iasql_functions/rpcs/iasql_get_sql_since.ts
aguillenv: Thank you for the feedback! I have added more comments and updated some variables' names. Let me know if it is a bit clearer now, or what else I could do to make it clearer

src/modules/iasql_functions/rpcs/iasql_get_sql_since.ts
```diff
@@ -0,0 +1,342 @@
...
+        );
+      }
+    }
+  }
+  return `${value}`;
+}
+
+/** @internal */
+function getDbString(v: any): string {
+  if (v === undefined) return `${null}`;
```
dfellis, line 252: Why this syntax to get `'null'`?

src/modules/iasql_functions/rpcs/iasql_get_sql_since.ts
```diff
@@ -0,0 +1,342 @@
...
+  if (typeof v === 'string') return `'${v}'`;
```
dfellis, line 253: Would it make sense to use `pg-format` here for these conversions, too? It's possible that we can replace the whole function with `format('$1', [v])`.

dfellis: (Also because this particular line will blow up if the string contains single quotes, itself)

src/modules/iasql_functions/rpcs/iasql_get_sql_since.ts
```diff
@@ -0,0 +1,342 @@
...
+      logger.warn(e.message ?? 'Error finding relation');
+      e = null;
+    }
+    const values = await Promise.all(
+      cloudColumns.map(
+        async (cc: string) =>
+          await jsonValueToDbString(entityMetadata.tableName, cc, e[cc], modsIndexedByTable, orm),
+      ),
+    );
+    return `(SELECT id FROM ${entityMetadata?.tableName} WHERE ${cloudColumns
```
dfellis, line 337: It's not *always* `id` as the join column name, though. Should we pull that from the entity, too?

aguillenv: why do you think it is necessary for other join column names? As I thought about it, the database `id` is the only one that is worth updating with its subquery because it is autogenerated, it will not match across databases. Other columns like `load_balancer_name` (if we were joined by it) are up to the user to update them.

src/modules/iasql_functions/rpcs/iasql_get_sql_since.ts
aguillenv: because we treat `undefined` as `null`s in the database, this should not occur, but if the value coming from the JSON is `undefined` we updated it to `null`. This function just returns the valid "typed" strings for the query.

Issue #1765, opened 2023-01-09
pipelinedeclaration does not fail inmediately when there is an error in the pipeline


yrobla:
When there is an error in the pipeline, the engine is not failing inmediately as soon as it's detecting that. Instead, it waits until the timeout period and then fails. This can be optimized by properly checking the status of the pipeline

PR #1766, opened 2023-01-09
fix: terminate pipeline execution when no pending/stopping states are…

yrobla:
… in the execution

Closes: #1765

### DIFF
--- a/src/modules/aws_codepipeline/mappers/pipeline_declaration.ts
+++ b/src/modules/aws_codepipeline/mappers/pipeline_declaration.ts
@@ -126,6 +126,7 @@ export class PipelineDeclarationMapper extends MapperBase<PipelineDeclaration> {
                   }
                 }
               }
+              if (!pipelinePending) break;
               // then we check if the stage completed successfully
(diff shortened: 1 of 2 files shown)


src/modules/aws_codepipeline/mappers/pipeline_declaration.ts
```diff
@@ -107,19 +106,34 @@ export class PipelineDeclarationMapper extends MapperBase<PipelineDeclaration> {
...
+              if (state && state.actionStates) {
+                for (const action of state.actionStates) {
+                  const latestStatus = action.latestExecution?.status;
+                  if (
+                    latestStatus &&
+                    ['Cancelled', 'Stopped', 'Superseeded', 'Failed'].includes(latestStatus)
+                  ) {
+                    // pipeline has failed, we can stop it
+                    pipelinePending = false;
+                    break;
```
aguillenv, line 125: this will break the inner loop on `state.actionStates`, but if `pipelinePending = false` do we want to also `break` the `data.stageStates` loop?

aguillenv (approved):

mtp1376 (approved):

PR #1774, opened 2023-01-11
Update `CONTRIBUTING.md` with the explanations for HTTP-based RPCs

mtp1376:


CONTRIBUTING.md
```diff
@@ -275,4 +275,22 @@ This section will change much more frequently compared to the other sections.
 
 There are probably many other footguns at the moment, feel free to update this with any you can think of!
 
-<!-- TODO: Revive Postgres function explanation for the new HTTP-based RPC approach -->
+## How IaSQL PostgreSQL functions work
+
+IaSQL container consists of two components:
```
dfellis, line 280: Nit

```suggestion
The IaSQL container consists of two components:
```

CONTRIBUTING.md
```diff
@@ -275,4 +275,22 @@ This section will change much more frequently compared to the other sections.
...
+
+The database-engine communication takes place via the Express.js server that the engine is listening on it. Whenever an IaSQL RPC command is called on the database (`SELECT iasql_install(...)`, `SELECT iasql_commit()`, or more generally `iasql_*`), an HTTP post request is sent from the Postgres database to the Express.js server. [We're using](https://github.com/iasql/iasql-engine/blob/c12d773402ad60d3c848743ced400584c08fcf8e/src/modules/interfaces.ts#L467-L482) [`pg_http` Postgres extension](https://github.com/pramsey/pgsql-http) to send HTTP requests from our Postgres database to the Express.js server.
+
+<img alt="IaSQL Container" src="assets/IaSQL-container.png" width="300"/>
+
+The Express.js server is trusting the input that it gets from the HTTP requests because they're all coming from the database and the `pg_http` calls in the database are configured in a pseudo-fixed structure that should not allow arbitrary requests.
```
dfellis, line 288: Not a comment on your PR here, but perhaps we should have a task to have the engine reject all non-localhost connections? We don't expose the port, but someone could and then accidentally break all security for themselves, right? Also weirdness with malformed TCP packets could be a vulnerability here.

CONTRIBUTING.md
```diff
@@ -275,4 +275,22 @@ This section will change much more frequently compared to the other sections.
...
+
+### How Is IaSQL Front-End Communicating The Database?
+
+If the only thing the IaSQL container exposes is the `5432` port which is a Postgres port, how does the IaSQL dashboard communicate with it?
+
+We have created a `run` service (listening on `https://run.iasql.com/`) which is taking the requests from the users and communicating with the IaSQL container's Postgres port. It's basically an "HTTP to Postgres" service with a bit of additional spices.
```
dfellis, line 294: Nit: "additional spices" is fine by me, but I would explicitly call out authentication here to address some security concerns?

```suggestion
We have created a `run` service (listening on `https://run.iasql.com/`) which is taking the requests from the users and communicating with the IaSQL container's Postgres port. It's basically an "HTTP to Postgres" service with authentication and a bit of additional spices.
```

CONTRIBUTING.md
```diff
@@ -275,4 +275,22 @@ This section will change much more frequently compared to the other sections.
...
+
+The database-engine communication takes place via the Express.js server that the engine is listening on it. Whenever an IaSQL RPC command is called on the database (`SELECT iasql_install(...)`, `SELECT iasql_commit()`, or more generally `iasql_*`), an HTTP post request is sent from the Postgres database to the Express.js server. [We're using](https://github.com/iasql/iasql-engine/blob/c12d773402ad60d3c848743ced400584c08fcf8e/src/modules/interfaces.ts#L467-L482) [`pg_http` Postgres extension](https://github.com/pramsey/pgsql-http) to send HTTP requests from our Postgres database to the Express.js server.
+
+<img alt="IaSQL Container" src="assets/IaSQL-container.png" width="300"/>
+
+The Express.js server is trusting the input that it gets from the HTTP requests because they're all coming from the database and the `pg_http` calls in the database are configured in a pseudo-fixed structure that should not allow arbitrary requests.
+
+### How Is IaSQL Front-End Communicating The Database?
```
aguillenv, line 290: is it ok to have this section here? wondering since the dashboard is private

mtp1376: This info can be gathered by anyone checking our website so it's not something private in my opinion. But it might be not related to the `iasql-engine` repository therefore we might not have it here. Do you think we should omit this section?

aguillenv: I think it would be interesting having this section in the dashboard's README.md rather than here. To me, it does not belong here, but I'm not against it. Let's see what are the other reviewers' thoughts about it.

CONTRIBUTING.md
mtp1376: Makes sense. I think that should be the default behavior for the engine's Express.js server but configurable.

aguillenv (approved):

depombo (approved):

CONTRIBUTING.md
```diff
@@ -275,4 +275,14 @@ This section will change much more frequently compared to the other sections.
...
+
+The IaSQL container consists of two components:
+- IaSQL Engine which is listening via [an Express.js server](https://github.com/iasql/iasql-engine/blob/c12d773402ad60d3c848743ced400584c08fcf8e/src/index.ts#L73-L77) on 8088 port (only accessible from within the container, therefore no direct access from outside)
+- IaSQL [Postgres database](https://github.com/iasql/iasql-engine/blob/c12d773402ad60d3c848743ced400584c08fcf8e/docker-entrypoint.sh#L47) which is listening on port 5432. This port is also exposed to the public so that the users are able to connect to their databases.
+
+The database-engine communication takes place via HTTP requests to the Express.js server. Whenever an IaSQL RPC command is called on the database (`SELECT iasql_install(...)`, `SELECT iasql_commit()`, or more generally `iasql_*`), an HTTP post request is sent from the Postgres database to the Express.js server. [Using](https://github.com/iasql/iasql-engine/blob/c12d773402ad60d3c848743ced400584c08fcf8e/src/modules/interfaces.ts#L467-L482) [`pg_http` Postgres extension](https://github.com/pramsey/pgsql-http) to send HTTP requests from the Postgres database to the Express.js server.
```
dfellis, line 284: Nit

```suggestion
The database-engine communication takes place via HTTP requests to the Express.js server. Whenever an IaSQL RPC command is called on the database (`SELECT iasql_install(...)`, `SELECT iasql_commit()`, or more generally `iasql_*`), an HTTP post request is sent from the Postgres database to the Express.js server. [Using](https://github.com/iasql/iasql-engine/blob/c12d773402ad60d3c848743ced400584c08fcf8e/src/modules/interfaces.ts#L467-L482) the [`pg_http` Postgres extension](https://github.com/pramsey/pgsql-http) to send HTTP requests from the Postgres database to the Express.js server.
```

CONTRIBUTING.md
mtp1376: https://github.com/iasql/iasql-engine/issues/1779

Issue #1731, opened 2022-12-23
Autoretry failed tests in the CI at least once


depombo:

aguillenv:
maybe something like this could help https://github.com/marketplace/actions/retry-step

PR #1776, opened 2023-01-11
feat: enable autoretry of tests once

yrobla:
This uses a third party action to allow retries. Set autoretry to 1, only if there is an error. Added that for actions that are prone to failure.

Closes: #1731

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,12 +119,12 @@ jobs:
           max_attempts: 1
           retry_on: error
           command: yarn coverage $(echo $COMMONS | jq -r '."${{matrix.common}}"')
-          env:
-            IASQL_ENV: test
-            COMMONS: ${{ needs.setup-common-test.outputs['common-tests'] }}
-            AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID_TESTING }}
-            AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY_TESTING }}
-            A0_IASQL_API_TOKEN: ${{ secrets.A0_IASQL_API_TOKEN }}
+        env:
+          IASQL_ENV: test
+          COMMONS: ${{ needs.setup-common-test.outputs['common-tests'] }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID_TESTING }}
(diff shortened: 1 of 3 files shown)


depombo (approved):

.github/workflows/ci.yml
```diff
@@ -318,10 +362,14 @@ jobs:
...
           STAGING_SECRET_ACCESS_KEY: ${{ secrets.STAGING_SECRET_ACCESS_KEY }}
           GH_PAT: ${{ secrets.GH_PAT }}
-        run: |
-          export AWS_ACCESS_KEY_ID=$(echo "${ACCESS_KEY_IDS}" | jq -r ".[${ACCOUNT_INDEX}]")
-          export AWS_SECRET_ACCESS_KEY=$(echo "${SECRET_ACCESS_KEYS}" | jq -r ".[${ACCOUNT_INDEX}]")
-          yarn test $(echo $MODULES | jq -r '."${{matrix.module}}"')
+        with:
+          max_attempts: 2
+          retry_on: error
+          timeout_minutes: 60
```
aguillenv, line 368: what's the difference between this timeout and the one in line 354?

.github/workflows/ci.yml
```diff
@@ -318,10 +362,14 @@ jobs:
           STAGING_ACCESS_KEY_ID: ${{ secrets.STAGING_ACCESS_KEY_ID }}
           STAGING_SECRET_ACCESS_KEY: ${{ secrets.STAGING_SECRET_ACCESS_KEY }}
           GH_PAT: ${{ secrets.GH_PAT }}
-        run: |
-          export AWS_ACCESS_KEY_ID=$(echo "${ACCESS_KEY_IDS}" | jq -r ".[${ACCOUNT_INDEX}]")
-          export AWS_SECRET_ACCESS_KEY=$(echo "${SECRET_ACCESS_KEYS}" | jq -r ".[${ACCOUNT_INDEX}]")
-          yarn test $(echo $MODULES | jq -r '."${{matrix.module}}"')
+        with:
+          max_attempts: 2
```
aguillenv, line 366: wondering if we should add a maximum of 3 attempts here

.github/workflows/ci.yml
yrobla: actually i will move to 90 minutes there. The retry action needs a timeout to be specified. But agree, needs to be the same

PR #1781, opened 2023-01-12
Postmortem for v0.0.23 release

### LINKED ISSUE #1769: Investigate `run` staging failure over time without usage

dfellis:
Resolves #1769

### DIFF
--- a/postmortems/002 - Postmortem for v0.0.23 release.md
+++ b/postmortems/002 - Postmortem for v0.0.23 release.md
@@ -1,0 +1,31 @@
+# 002 - Postmortem for v0.0.23 release
+
+## Level: Internal
+
+## Author(s)
+
+- David Ellis <david@iasql.com>
+
+## Summary
+
+We have been unable to release v0.0.23, a significant rearchitecting and feature-adding version of IaSQL, for approximately a month, *primarily* due to issues with the new infrastructure required in staging and production. There are five incidences we have observed on staging (none on production as production is not yet in use) and we have applied mitigations to staging and production for them all. This document will outline them.
+
+## Timeline
+
+- **2022-12-20**: Staging cannot load docker container images because it has run out of disk space. Increased boot volume to 16GB.
+- **2022-12-27**: Staging deploys fail complaining about lack of disk space, again. Dropping `stdout/err` in favor of LogDNA recommended.
+- **2023-01-03**: Staging `run` service crashed, cannot be deployed automatically. Deployment code written, better logging and auto-restart added to `run` service. Discord altering on staging/production status revived.
+- **2023-01-08**: Staging instance becomes unreachable. Somehow the DHCP daemon failed to get an IP address assigned by AWS. Manually stopping and starting the server revived it.
+- **2023-01-11**: Staging cannot deploy. Authorization to ghcr.io failed. Replaced Github Personal Access Token. Recommend monthly calendar task to replace the token until ghcr.io is replaced with dockerhub (potentially).
+
+## Detection
+
+Our test suite for the Dashboard uses staging, so tests there start failing, but with low activity over the holidays, this was delayed for the January 3rd situation. The first and last issue happened during deployment so they were noticed immediately. The lack of disk space on December 27th and the instance being unreachable on January 8th were only spotted by deployment failures (but not caused by them). They should be spotted sooner now that Discord health notifications have been revived.
+
+## Response, Cause, Prevention
+
+Combining these three sections because most of the incidents were relatively easy to figure out and resolve.
+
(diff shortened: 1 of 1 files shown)


mtp1376 (approved):
Great explanation. Thanks.

postmortems/002 - Postmortem for v0.0.23 release.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Timeline
+
+- **2022-12-20**: Staging cannot load docker container images because it has run out of disk space. Increased boot volume to 16GB.
+- **2022-12-27**: Staging deploys fail complaining about lack of disk space, again. Dropping `stdout/err` in favor of LogDNA recommended.
+- **2023-01-03**: Staging `run` service crashed, cannot be deployed automatically. Deployment code written, better logging and auto-restart added to `run` service. Discord altering on staging/production status revived.
+- **2023-01-08**: Staging instance becomes unreachable. Somehow the DHCP daemon failed to get an IP address assigned by AWS. Manually stopping and starting the server revived it.
+- **2023-01-11**: Staging cannot deploy. Authorization to ghcr.io failed. Replaced Github Personal Access Token. Recommend monthly calendar task to replace the token until ghcr.io is replaced with dockerhub (potentially).
```
depombo, line 19: hmm having a CI job that pushed to a public ECR repo in the bootstrap AWS account is probably best?

yrobla (approved):

postmortems/002 - Postmortem for v0.0.23 release.md
```diff
@@ -0,0 +1,31 @@
...
+- David Ellis <david@iasql.com>
+
+## Summary
+
+We have been unable to release v0.0.23, a significant rearchitecting and feature-adding version of IaSQL, for approximately a month, *primarily* due to issues with the new infrastructure required in staging and production. There are five incidences we have observed on staging (none on production as production is not yet in use) and we have applied mitigations to staging and production for them all. This document will outline them.
+
+## Timeline
+
+- **2022-12-20**: Staging cannot load docker container images because it has run out of disk space. Increased boot volume to 16GB.
+- **2022-12-27**: Staging deploys fail complaining about lack of disk space, again. Dropping `stdout/err` in favor of LogDNA recommended.
```
aguillenv, line 16: do we have a task for this already?

postmortems/002 - Postmortem for v0.0.23 release.md
```diff
@@ -0,0 +1,31 @@
...
+- **2023-01-03**: Staging `run` service crashed, cannot be deployed automatically. Deployment code written, better logging and auto-restart added to `run` service. Discord altering on staging/production status revived.
+- **2023-01-08**: Staging instance becomes unreachable. Somehow the DHCP daemon failed to get an IP address assigned by AWS. Manually stopping and starting the server revived it.
+- **2023-01-11**: Staging cannot deploy. Authorization to ghcr.io failed. Replaced Github Personal Access Token. Recommend monthly calendar task to replace the token until ghcr.io is replaced with dockerhub (potentially).
+
+## Detection
+
+Our test suite for the Dashboard uses staging, so tests there start failing, but with low activity over the holidays, this was delayed for the January 3rd situation. The first and last issue happened during deployment so they were noticed immediately. The lack of disk space on December 27th and the instance being unreachable on January 8th were only spotted by deployment failures (but not caused by them). They should be spotted sooner now that Discord health notifications have been revived.
```
aguillenv, line 23: In the case of an instance lack of disk space, does the instance becomes unhealthy or should we try to create another check to the instance so we can get notified through discord?

aguillenv (approved):

postmortems/002 - Postmortem for v0.0.23 release.md
dfellis: It was already done in the engine, but not the run service. I did it for the run service at the same time I improved the logging and added the auto restart code.

postmortems/002 - Postmortem for v0.0.23 release.md
dfellis: Running out of space on the boot disk immediately causes chaos with Docker and a couple of other services on the instance, so we *should* be alerted.

postmortems/002 - Postmortem for v0.0.23 release.md
dfellis: Staging and Production probably would have different needs here, and the "run" service will always be private. I went with ghcr.io instead of a public or private ECR because we pay a flat fee for ghcr.io baked into our github org versus variable billing from AWS. The credential rotation is annoying, but not a *huge* deal.

I mentioned moving the engine to a public dockerhub and using that because that's where we're gonna go eventually, anyways? But thinking about it it won't remove this credential rotation requirement for the run service.

depombo (approved):

PR #1792, opened 2023-01-16
feat: add sns topic support

yrobla:
Closes: #1772


depombo:
awesome, let's add SNS to https://github.com/iasql/iasql-engine#aws-services-with-basic-api-coverage

src/modules/aws_sns/mappers/topic.ts
```diff
@@ -0,0 +1,248 @@
...
+    'tracingConfig',
+  ];
+
+  capitalize = (s: string) => {
+    return s.charAt(0).toUpperCase() + s.slice(1);
+  };
+
+  async topicMapper(t: string, region: string, ctx: Context) {
+    let out = new Topic();
+    if (!t) return undefined;
```
aguillenv, line 81:
```suggestion
    if (!t) return undefined;
    let out = new Topic();
```

src/modules/aws_sns/mappers/topic.ts
```diff
@@ -0,0 +1,248 @@
...
+    'policy',
+    'signatureVersion',
+    'tracingConfig',
+  ];
+
+  capitalize = (s: string) => {
+    return s.charAt(0).toUpperCase() + s.slice(1);
+  };
+
+  async topicMapper(t: string, region: string, ctx: Context) {
```
aguillenv, line 79: Can we rename this variable to something more explicit like `arn` or  `topicArn`?

aguillenv:
now that https://github.com/iasql/iasql-engine/pull/1801 has been merged you could rebase and add a test to avoid regressions.

aguillenv (approved):

src/modules/aws_sns/mappers/topic.ts
```diff
@@ -0,0 +1,258 @@
...
+    out.region = region;
+    out.name = parseArn(arn).resource;
+    return out;
+  }
+
+  cloud: Crud2<Topic> = new Crud2({
+    create: async (es: Topic[], ctx: Context) => {
+      const out = [];
+      for (const e of es) {
+        if (!e.name || !e.region) continue; // cannot create topic without name or region
```
mtp1376, line 121: `name` and `region` are both non-nullable. I think we have the extra check for `region` and it has to be a valid region (due to FK). So maybe this is redundant.

src/modules/aws_sns/mappers/topic.ts
```diff
@@ -0,0 +1,258 @@
...
+        const client = (await ctx.getAwsClient(e.region)) as AWS;
+        const attr = new Map();
```
mtp1376, line 123: Wondering why not creating an empty object and setting the properties by using `attr[property] = value` syntax.

src/modules/aws_sns/mappers/topic.ts
```diff
@@ -0,0 +1,258 @@
...
+          Attributes: Object.fromEntries(attr),
+          DataProtectionPolicy: e.dataProtectionPolicy,
+        };
+        const result = await this.createTopic(client.snsClient, input);
+        if (!result) throw new Error('Error creating SNS topic');
+
+        // update topic with the ARN
+        e.arn = result;
+        await this.module.topic.db.update(e, ctx);
+        out.push(e);
```
mtp1376, line 143:
```suggestion
        const arn = await this.createTopic(client.snsClient, input);
        if (!arn) throw new Error('Error creating SNS topic');

        // update topic with the ARN
        e.arn = arn;
        await this.module.topic.db.update(e, ctx);
        out.push(e);
```

src/modules/aws_sns/mappers/topic.ts
```diff
@@ -0,0 +1,258 @@
...
+          // if data protection policy is set and is fifo topic, we restore it
+          if (
+            !Object.is(e.arn, cloudRecord.arn) ||
+            !Object.is(e.fifoTopic, cloudRecord.fifoTopic) ||
+            (e.fifoTopic && e.dataProtectionPolicy)
+          ) {
+            cloudRecord.id = e.id;
+            await this.module.topic.db.update(cloudRecord, ctx);
+            out.push(cloudRecord);
+            break;
```
mtp1376, line 194: Maybe this should be a `continue;` instead?

src/modules/aws_sns/mappers/topic.ts
yrobla: i was actually getting some complains from typescript when using that method, something like Type 'string' is not assignable to type 'never' . So i took this other approach , reading feedback from other people hitting the same

src/modules/aws_sns/entity/topic.ts
```diff
@@ -0,0 +1,165 @@
...
+  /**
+   * @public
+   * Set to true to create a FIFO topic
+   * @see https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-sns/interfaces/settopicattributescommandinput.html
+   */
+  @Column({
+    type: 'boolean',
+    default: false,
+  })
+  fifoTopic: boolean;
```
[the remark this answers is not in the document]
yrobla: mm personally for me i find more convenient , even in postgres, using a boolean, than using a enum... because when using a boolean i always know that possible values are true/false. Maybe i can rename it to "isFifoTopic" ?

mtp1376 (approved):

PR #1793, opened 2023-01-16
Enabling hot reload

aguillenv:
I did not benchmark it but doing the `yarn build` and then starting seems faster than using `ts-node` directly with the `index.ts`


yrobla:
I tested that and improvement is clear for me, it reloads without the need of stopping all the procedures, waiting, rebuilding etc... i can say it's a great improvement in productivity

package.json
```diff
@@ -13,15 +13,17 @@
     "coverage": "IASQL_ENV=test jest -i --coverage --forceExit",
     "coverage:local": "IASQL_ENV=test export $(cat .env | xargs) && yarn coverage",
     "serve": "node --max_old_space_size=8192 dist/index.js",
-    "serve:debug": "node --inspect=0.0.0.0:9229 --max_old_space_size=8192 dist/index.js",
+    "serve:debug": "yarn build && node --max_old_space_size=8192 dist/index.js",
     "start": "yarn serve",
-    "start:debug": "nodemon -e ts -w ./src -x yarn serve:debug",
+    "start:debug": "nodemon --inspect=0.0.0.0:9229 -e ts -w ./src -x yarn serve:debug",
```
dfellis, line 18: We should `EXPOSE` port 9229, as well, so we can hook up the Chrome Node Debugger to the engine during dev.

Dockerfile.dev
```diff
@@ -0,0 +1,45 @@
...
+
+# Default ENVs that can be overwritten
+ARG IASQL_ENV=local
+ENV IASQL_ENV=$IASQL_ENV
+ARG DB_USER=postgres
+ENV DB_USER=$DB_USER
+ARG DB_PASSWORD=test
+ENV DB_PASSWORD=$DB_PASSWORD
+
+EXPOSE 5432
```
dfellis, line 44: Eg:

```suggestion
EXPOSE 5432
EXPOSE 9229
```

aguillenv: Committing this change but this is merely informative, the exposure is already being done in the `docker run` command with the port forwarding option `-p 9229:9229`.

Quoting from [`EXPOSE` docker docs](https://docs.docker.com/engine/reference/builder/#expose):
> The EXPOSE instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published. To actually publish the port when running the container, use the -p flag on docker run to publish and map one or more ports, or the -P flag to publish all exposed ports and map them to high-order ports.

yrobla (approved):

package.json
```diff
@@ -13,15 +13,17 @@
...
     "gen-sql": "IASQL_ENV=ci ts-node ./node_modules/.bin/typeorm migration:generate -n",
     "migrate": "IASQL_ENV=ci ts-node ./node_modules/.bin/typeorm migration:run",
     "gen-module": "IASQL_ENV=ci ./.module_migration.sh",
     "docker-build": "docker build -t iasql:latest .",
+    "docker-build:dev": "docker build -f ./Dockerfile.dev -t iasql:latest-dev .",
     "docker-run": "docker run -p 5432:5432 --name iasql iasql:latest",
     "docker-run:detach": "docker run -d -p 5432:5432 --name iasql iasql:latest",
+    "docker-run:dev": "docker run -p 5432:5432 -p 9229:9229 --name iasql-dev -v $PWD/src:/engine/src -v /engine/node_modules iasql:latest-dev",
```
mtp1376, line 26: `-v /engine/node_modules` going to make issues if the host and the docker container are from two different architectures.

dfellis (approved):

depombo (approved):

Dockerfile.dev
dfellis: Huh. I wonder if it used to be required?

package.json
mtp1376: I think you're not overwriting the local `node_module`. Just signaling the persistence, right?

In that case it's ok

PR #1795, opened 2023-01-16
Add coverage for InternetGateway

mtp1376:
Signed-off-by: Mohammad Teimori Pabandi <mtp1376@gmail.com>


test/modules/aws-vpc-internet-gateway.ts
```diff
@@ -0,0 +1,243 @@
...
+    ),
+  );
+  it('applies creation of the vpc', commit());
+
+  it('starts a transaction', begin());
+  it(
+    'adds a new internet gateway to the vpc in the region',
+    query(
+      `
+          INSERT INTO internet_gateway (vpc_id, tags, region)
```
yrobla, line 118: can we have 2 internet gateways for the same vpc? what happens if you do that?

mtp1376: Good question. I've changed the relationship to a `nullable` 1-1 relationship between InternetGateway and VPC.

src/modules/aws_vpc/mappers/internet_gateway.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    return (await client.createInternetGateway(input)).InternetGateway;
+  }
+
+  async internetGatewayMapper(ig: AwsInternetGateway, region: string, ctx: Context) {
+    const out = new InternetGateway();
+    out.internetGatewayId = ig.InternetGatewayId;
+    out.tags = convertTagsFromAws(ig.Tags);
+    if (!!ig.Attachments && ig.Attachments.length)
```
aguillenv, line 65: so we have defined an `OneToOne` relation between`Vpc` and `InternetGateway` but based on this it seems you can have one `InternetGateway` attached to multiple `Vpc`s? Should we handle that?

mtp1376: No despite AWS is treating that as a list, but a VPC can only have one InternetGateway. Maybe the other attachments are from another type? (Like Egress-only InternetGateway) I will check it.

mtp1376: I actually can't find a reason why the `Attachments` is a list. The doc says this:
```ts
    /**
     * <p>Any VPCs attached to the internet gateway.</p>
     */
    Attachments?: InternetGatewayAttachment[];
```
But the AWS console and this StackOverflow answer both say that VPC and InternetGateway have a 1-1 relationship.
https://stackoverflow.com/a/49356420/2594912

aguillenv (approved):

PR #1801, opened 2023-01-17
Update `undefined`s before save

### LINKED ISSUE #1806: Error updating optional properties back to null
Entities with possible `undefined` values cannot be set back to `undefined` after they have an actual value because Typeorm `.save()` method ignores `undefined`s when inserting/updating, so they have to be explicitly set to `null` to be taken into account

aguillenv:
Closes #1806

### DIFF
--- a/src/modules/aws_vpc/mappers/nat_gateway.ts
+++ b/src/modules/aws_vpc/mappers/nat_gateway.ts
@@ -197,8 +197,9 @@ export class NatGatewayMapper extends MapperBase<NatGateway> {
           if (!input.AllocationId) throw new Error('Elastic ip need to be created first');
         } else if (!e.elasticIp && e.connectivityType === ConnectivityType.PUBLIC) {
           const elasticIp = new ElasticIp();
-          // Attach the same tags in case we want to associate them visualy through the AWS Console
+          // Attach the same tags in case we want to associate them visually through the AWS Console
           elasticIp.tags = e.tags;
+          elasticIp.region = e.region;
           const newElasticIp = await this.module.elasticIp.cloud.create(elasticIp, ctx);
           if (!newElasticIp || newElasticIp instanceof Array) continue;
           input.AllocationId = newElasticIp.allocationId;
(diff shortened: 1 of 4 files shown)


yrobla:
i tested locally, works fine for me

src/modules/aws_vpc/mappers/nat_gateway.ts
```diff
@@ -197,8 +197,9 @@ export class NatGatewayMapper extends MapperBase<NatGateway> {
           if (!input.AllocationId) throw new Error('Elastic ip need to be created first');
         } else if (!e.elasticIp && e.connectivityType === ConnectivityType.PUBLIC) {
           const elasticIp = new ElasticIp();
-          // Attach the same tags in case we want to associate them visualy through the AWS Console
+          // Attach the same tags in case we want to associate them visually through the AWS Console
           elasticIp.tags = e.tags;
+          elasticIp.region = e.region;
```
yrobla, line 202: that is related to the PR?

aguillenv: Yes, the test for this mapper was consistently failing because it was trying to `db.update` with region `undefined`, that was inserting a `null` and the region cannot be `null` for the elastic ip entity

yrobla (approved):

dfellis (approved):

src/modules/subscribers.ts
```diff
@@ -9,6 +9,15 @@ function updateNulls(entity: any) {
...
 
+function updateUndefined(entity: any) {
+  if (entity) {
+    const that: any = entity;
+    Object.keys(entity).forEach(k => {
+      if (that[k] === undefined) that[k] = null;
+    });
+  }
+}
+
```
mtp1376, line 20:
```suggestion
function updateUndefined(entity: any) {
  if (entity) {
    Object.keys(entity).forEach(k => {
      if (entity[k] === undefined) entity[k] = null;
    });
  }
}
```
I think this does the same thing. Is there a reason why you have created the `that` variable?

mtp1376 (approved):

src/modules/subscribers.ts
aguillenv: I think you are right, this is copied from the other method in this file and came from a legacy code inside the entities' class code, which was messing up with `this`. I'll try to update it in a follow-up PR to unblock PRs waiting for this one.

PR #1808, opened 2023-01-18
feat: add support for topic subscriptions

yrobla:


src/modules/aws_sns/mappers/subscription.ts
```diff
@@ -0,0 +1,108 @@
...
+              if (mappedSubscription) out.push(mappedSubscription);
+            }
+          }),
+        );
+
+        return out;
+      }
+    },
+    update: async (es: Subscription[], ctx: Context) => {
+      // no action
```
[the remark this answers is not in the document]
aguillenv: this one still pending

aguillenv (approved):

test/modules/aws-sns-integration.ts
```diff
@@ -27,7 +31,33 @@ const installAll = runInstallAll.bind(null, dbAlias);
...
 
-const modules = ['aws_sns'];
+const lambdaFunctionName = `${prefix}${dbAlias}`;
+const lambdaFunctionCode =
+  'UEsDBBQAAAAIADqB9VRxjjIufQAAAJAAAAAIABwAaW5kZXguanNVVAkAAzBe2WIwXtlidXgLAAEE9QEAAAQUAAAANcyxDoIwEIDhnae4MNFIOjiaOLI41AHj5NLUA5scV3K9Gojx3ZWB8R++H5c5iWb78vwkFDgD+LxygKFw0Ji4wTeythASKy5q4FPBFjkRWkpjU3f3zt1O8OAaDnDpr85mlchjHNYdcyFq4WjM3wpqEd5/26JXQT85P2H1/QFQSwECHgMUAAAACAA6gfVUcY4yLn0AAACQAAAACAAYAAAAAAABAAAApIEAAAAAaW5kZXguanNVVAUAAzBe2WJ1eAsAAQT1AQAABBQAAABQSwUGAAAAAAEAAQBOAAAAvwAAAAAA';
+// Base64 for zip file with the following code:
+// exports.handler =  async function(event, context) {
+//   console.log("EVENT: \n" + JSON.stringify(event, null, 3))
+//   return context.logStreamName
+// }
```
depombo, line 41: nit- having the function as a string and encoding using the node.js std

depombo (approved):

dfellis:
This looks like a legit error https://github.com/iasql/iasql/actions/runs/4326576045/jobs/7554282560#step:9:35269

The solution may just be a small change to the delete sql statement to also delete the subscription first, however.

test/modules/aws-sns-integration.ts
yrobla: it's not just base64, it will need to zip it as well , needing an extra dependency. So i prefered to just leave it simple as it was only used here

PR #1809, opened 2023-01-18
Add workflow to run a script on all test accounts

mtp1376:


.github/workflows/run-on-all-test-accounts.yml
```diff
@@ -0,0 +1,134 @@
+name: Run On All Test Accounts
+on:
+  push:
+    branches: [ add-run-on-all-test-accounts-job ]
+  workflow_dispatch:
+    inputs:
+      script_address:
+        required:  true
+        description: 'Specify the script to run on all the accounts'
```
dfellis, line 9: This input is not used.

.github/workflows/run-on-all-test-accounts.yml
```diff
@@ -0,0 +1,133 @@
+name: Run On All Test Accounts
+on:
+  workflow_dispatch:
+    inputs:
+      script_address:
```
aguillenv, line 5: more like `script_path`?

.github/workflows/run-on-all-test-accounts.yml
```diff
@@ -0,0 +1,133 @@
...
+        required:  true
+        description: 'Specify the script to run on all the accounts'
```
aguillenv, line 7: Could you please add an example here? I'm assuming it would be a relative path, right? something like `src/scripts/my-script.sh`?

.github/workflows/run-on-all-test-accounts.yml
```diff
@@ -0,0 +1,133 @@
...
+  run-command:
+    runs-on: ubuntu-latest
+    needs:
+      - setup-variables
+      - build-and-push
+    name: run on account ${{ matrix.module }}
+    strategy:
+      fail-fast: false
+      matrix:
+        module: ${{ fromJson(needs.setup-variables.outputs['test-accounts']) }}
```
aguillenv, line 57: Maybe instead of `module` could be `index`?

aguillenv (approved):

yrobla (approved):

PR #1813, opened 2023-01-20
Update "Run On All Test Accounts" workflow

aguillenv:


mtp1376:
So the accounts start from 0 and end in 99? If so this should be merged.

mtp1376 (approved):

yrobla (approved):

aguillenv:
> So the accounts start from 0 and end at 99? If so this should be merged.

yes, something we missed. Also, the test on account 100 was failing (because it does not exists) but the test was succeeding, so I just added the early exist if `psql` command fails

dfellis (approved):

PR #1823, opened 2023-01-21
restore upgrade

dfellis:
The upgrade logic works!

Still a draft because I need to:

- [ ] Revive the upgrade test to prevent regressions
- [ ] Add some code to clean up the temporary files created in the new upgrade flow
- [ ] Re-organize the code, particularly in `src/index.ts` that probably should be put somewhere else
- [ ] I was hoping to eliminate the migration glitch with this one, but there's still one situation where it shows itself, and I'm hoping to tackle it


src/services/iasql.ts
```diff
@@ -231,3 +233,99 @@ export async function dump(dbId: string, dataOnly: boolean) {
...
+          SELECT iasql_install($1);
+        `,
+          [mod],
+        );
+      }
+      execSync(`rm -rf /tmp/upgrade/${db}`);
+      dbsDone[db] = true;
+      clearInterval(upgradeHandle);
+      logger.info(`Part 3 of 3 for ${db} complete!`);
+      if (Object.keys(dbsDone).sort().join(',') === dbs.sort().join(',')) {
```
aguillenv, line 324: what should we do if they are not equal? 😨

src/services/iasql.ts
```diff
@@ -231,3 +233,99 @@ export async function dump(dbId: string, dataOnly: boolean) {
...
+    for (const line of auditLogLines) {
+      await conn.query(
+        `
+        INSERT INTO iasql_audit_log (ts, "user", table_name, change_type, change, message) VALUES
+        ($1, $2, $3, $4, $5, $6);
+      `,
+        [line.ts, line.user, line.table_name, line.change_type, line.change, line.message],
+      );
+    }
+    logger.info(`Part 2 for ${db} complete!`);
```
aguillenv, line 261:
```suggestion
    logger.info(`Part 2 of 3 for ${db} complete!`);
```

src/services/iasql.ts
```diff
@@ -231,3 +233,99 @@ export async function dump(dbId: string, dataOnly: boolean) {
     await conn2?.close();
   }
 }*/
+
+export async function upgrade() {
+  const dbs = readdirSync('/tmp/upgrade');
+  const dbsDone: { [key: string]: boolean } = {};
+  for (const db of dbs) {
+    logger.info(`Starting Part 2 of upgrading ${db}`);
```
aguillenv, line 241: Just for consistency 
```suggestion
    logger.info(`Starting Part 2 of 3 for ${db}`);
```

src/scripts/upgrade.sh
```diff
@@ -0,0 +1,247 @@
...
+				INNER JOIN pg_enum e ON t.oid = e.enumtypid
+				INNER JOIN pg_catalog.pg_namespace n ON n.oid = t.typnamespace
+				WHERE n.nspname = 'public';
+			")
+      for ENUM in $ENUMS
+      do
+        psql $DB -c "DROP TYPE \"$ENUM\";"
+      done
+
+      echo Part 1 for $DB complete!
```
aguillenv, line 244:
```suggestion
      echo Part 1 of 3 for $DB complete!
```

test/upgrade/run-upgrade.sh
```diff
@@ -3,5 +3,94 @@
...
+  SELECT iasql_begin();
+"
+psql $CONNSTR -c "
+  INSERT INTO aws_credentials (access_key_id, secret_access_key)
+  VALUES ('${AWS_ACCESS_KEY_ID}', '${AWS_SECRET_ACCESS_KEY}');
+"
+psql $CONNSTR -c "
+  SELECT * FROM iasql_commit();
+"
+psql $CONNSTR -c "
```
yrobla, line 53: why us-east-1 here?

src/scripts/upgrade.sh
```diff
@@ -0,0 +1,247 @@
...
+					" > /tmp/upgrade/$DB/default_region
+				fi
+			done
+
+			# All IaSQL DBs have an audit log, so this doesn't need a conditional check
+			psql $DB -tc "
+        SELECT array_to_json(ARRAY(SELECT row_to_json(iasql_audit_log) FROM iasql_audit_log));
+			" > /tmp/upgrade/$DB/audit_log
+
+			# TODO 2: For now just show the outputs of these various queries, actually start deleting stuff later
```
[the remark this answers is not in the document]
dfellis: Oh, yes, that was a note to me while I was debugging the Postgres outputs.

src/services/iasql.ts
dfellis: They usually will be not equal. They're only equal once every single DB has been upgraded.

src/scripts/upgrade.sh
```diff
@@ -0,0 +1,247 @@
...
+
+		if [[ $SHOULD_UPGRADE ]]; then
+			echo $DB to be upgraded!
+
+			# It's not possible to actually re-install the new modules without the engine, the RPC SQL is
+			# generated by Typescript code, so instead we're gonna grab the relevant info, stuff it into
+			# temporary files, then continue in Typescript hooking into the startup process, which will
+			# look for the `/tmp/upgrade/` path to decide if it needs to execute or not.
+
+			mkdir -p /tmp/upgrade/$DB
```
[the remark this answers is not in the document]
dfellis: Yes, at the very end of the third part of the upgrade process we check if every DB has been upgraded and if they have been, then we delete the whole thing.

test/upgrade/run-upgrade.sh
```diff
@@ -3,5 +3,94 @@
...
+# Clear mutations, if any
+git reset --hard HEAD
+git checkout ${CURRENTGITSHA}
+
+mkdir -p ${PWD}/../db
+yarn docker-build
+yarn docker-run-vol &
+while ! psql postgres://postgres:test@localhost:5432/iasql_metadata -b -q -c "SELECT iasql_engine_health()"; do sleep 1 && echo -n .; done;
+
+UPGRADECHECKCOUNT=300
```
yrobla, line 75: this is just to act as the limit of retries?

src/scripts/upgrade.sh
```diff
@@ -0,0 +1,247 @@
...
+	")
+	for DB in $USER_DATABASES
+	do
+		echo Checking $DB ...
+
+		# Get the engine and DB versions, figure out which is the newest and if that newest version is
+		# a beta release. *Always* re-run the upgrade on beta releases, and if not a beta release, run
+		# the upgrade only if the DB is an earlier version than the engine
+
+		ENGINE_VERSION=$(jq -r .version /engine/package.json | xargs) # xargs is the recommended way to 'trim' strings. Oh Bash...
```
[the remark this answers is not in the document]
dfellis: true. I put them together because it made reading through this file easier, but if you don't think there'd be any confusion, I can pull it out of the loop.

src/scripts/upgrade.sh
```diff
@@ -0,0 +1,247 @@
...
+				fi
+			done
+
+			# All IaSQL DBs have an audit log, so this doesn't need a conditional check
+			psql $DB -tc "
+        SELECT array_to_json(ARRAY(SELECT row_to_json(iasql_audit_log) FROM iasql_audit_log));
+			" > /tmp/upgrade/$DB/audit_log
+
+			# TODO 2: For now just show the outputs of these various queries, actually start deleting stuff later
+
```
dfellis, line 102:
```suggestion
```

src/scripts/upgrade.sh
```diff
@@ -0,0 +1,247 @@
...
+# TODO: Also need to add an `UPGRADE` audit log type we can add in there to make it clear when prior logs are no longer fully "trustworthy" for the purposes of re-generating SQL statements from them.
+
+# Check if there's any tables in the metadata database, in case this is a first-time run
+HAS_IASQL_DATABASE=$(psql iasql_metadata -qtc "
+	SELECT table_name
+	FROM information_schema.tables
+	WHERE table_name = 'iasql_database';
+")
+
+if [ "$HAS_IASQL_DATABASE" != "" ]; then
```
yrobla, line 30: what are we doing if there is no database? shall we throw some warn message?

src/scripts/upgrade.sh
```diff
@@ -0,0 +1,247 @@
...
+
+			for MODULE in $MODULE_LIST
+			do
+				if [[ "$MODULE" == "aws_account" ]]; then
+					# TODO: Figure out a more generic way of dealing with non-synced user data
+					psql $DB --csv -tc "
+						SELECT access_key_id, secret_access_key FROM aws_credentials;
+					" > /tmp/upgrade/$DB/creds
+          # Temporarily print the creds to see what's going on here
+          cat /tmp/upgrade/$DB/creds
```
yrobla, line 84: shall it be removed?

src/scripts/upgrade.sh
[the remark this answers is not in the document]
aguillenv: yeah probably is better to follow having it here, let's keep it, I don't think it will penalize the performance that much!

src/scripts/upgrade.sh
dfellis: Yep. The end of step 3 deletes `/tmp/upgrade/$DB` and after they're all done `/tmp/upgrade` is also deleted

test/upgrade/run-upgrade.sh
dfellis: No reason, I just needed a default.

test/upgrade/run-upgrade.sh
dfellis: Yep

src/scripts/upgrade.sh
aguillenv: I think Yolanda means we should remove this `cat` call

dfellis: ohh, yes definitely should delete it :)

src/scripts/upgrade.sh
```diff
@@ -0,0 +1,245 @@
...
```
dfellis, line 84:
```suggestion
```

src/scripts/upgrade.sh
```diff
@@ -0,0 +1,245 @@
...
+# Alternatively, instead of storing the logic there into the metadata database, maybe just fork a disconnected child process and sleep for a few seconds, then continue, running pure IaSQL code to do the rest? (Well, re-running the `up` migration code for the `iasql_*` modules would need to be done in the script, but that part could be hacked into the shell script since it doesn't depend on using the AWS SDK. Then the restoration of the database is IaSQL-powered SQL only.
+# Note: The audit log data would be restored immediately after the `iasql_*` modules are added, but not sure if we can do that with a "normal" SQL script or if we need to do something special there to get it in.
```
dfellis, line 20:
```suggestion
```

src/scripts/upgrade.sh
dfellis: Oh, this is to bypass the upgrade script on the first-time run of the engine.

aguillenv (approved):

src/services/iasql.ts
```diff
@@ -283,48 +283,75 @@ export async function upgrade() {
...
-        );
+        try {
+          await conn.query(
+            `
+            SELECT iasql_install($1);
+          `,
+            [mod],
+          );
+        } catch (e) {
+          logger.warn('Failed to install a module on upgrade', { e, });
```
aguillenv: since we have the module in `mod`, would be nice if this message points to the troublemaker

dfellis: Sure

src/services/iasql.ts
```diff
@@ -231,3 +233,135 @@ export async function dump(dbId: string, dataOnly: boolean) {
...
+        }
+        try {
+          await conn.query(
+            `
+            UPDATE aws_regions SET is_default = TRUE WHERE region = $1;
+          `,
+            [defaultRegion],
+          );
+        } catch (e) {
+          logger.warn('Failed to set the default region', { e });
```
dfellis, line 339:
```suggestion
          logger.warn('Failed to set the default region', { e, defaultRegion });
```

src/services/iasql.ts
```diff
@@ -231,3 +233,135 @@ export async function dump(dbId: string, dataOnly: boolean) {
...
+      for (const mod of moduleList) {
+        try {
+          await conn.query(
+            `
+            SELECT iasql_install($1);
+          `,
+            [mod],
+          );
+        } catch (e) {
+          logger.warn('Failed to install a module on upgrade', { e });
```
dfellis, line 353:
```suggestion
          logger.warn('Failed to install a module on upgrade', { e, mod });
```

yrobla (approved):

PR #1832, opened 2023-01-23
Better typing for `Context` object

mtp1376:


src/modules/interfaces.ts
```diff
@@ -17,7 +18,12 @@ import logger from '../services/logger';
 
 export type IdFields = { [key: string]: string };
 
-export type Context = { [key: string]: any };
+export type Context = {
+  [key: string]: any;
+  getEnabledAwsRegions: (fullEntities?: boolean) => Promise<string[]>;
+  getAwsClient: (region?: string) => Promise<AWS>;
+  getDefaultRegion: () => Promise<string>;
+};
```
dfellis, line 26: This isn't a good approach here because it breaks encapsulation of the modules.

Better if we add helper logic to construct a `Context` type in each module that combines the "Base" context with the memo object with the `module.provides.context` object. That would help with the typing but also keep the code better organized and keep this more correct (when we start supporting more than AWS, those three methods *won't* be on the context object for those other modules, for instance).

PR #1838, opened 2023-01-23
Add `customSslCertificate` and `alternateDomainNames` Coverage to CloudFront `distribution`

mtp1376:


src/modules/aws_cloudfront/index.ts
```diff
@@ -117,17 +118,30 @@ class DistributionMapper extends MapperBase<Distribution> {
       WebACLId: e.webACLId,
       Origins: { Quantity: e.origins.length, Items: e.origins },
       PriceClass: distributionConfig.DistributionConfig?.PriceClass,
-      Aliases: distributionConfig.DistributionConfig?.Aliases,
+      Aliases: { Quantity: e.alternateDomainNames?.length ?? 0, Items: e.alternateDomainNames ?? [] },
```
yrobla, line 121: why do we track Quantity here? is not just length of items?

src/modules/aws_cloudfront/index.ts
```diff
@@ -117,17 +118,30 @@ class DistributionMapper extends MapperBase<Distribution> {
...
       OriginGroups: distributionConfig.DistributionConfig?.OriginGroups,
       Restrictions: distributionConfig.DistributionConfig?.Restrictions,
-      ViewerCertificate: distributionConfig.DistributionConfig?.ViewerCertificate,
     };
 
+    if (e.customSslCertificate?.arn)
+      req.ViewerCertificate = {
+        CloudFrontDefaultCertificate: false,
+        ACMCertificateArn: e.customSslCertificate.arn,
+        MinimumProtocolVersion: 'TLSv1.2_2021',
```
yrobla, line 135: maybe put that in some constant, so it's easier to modify in the future

src/modules/aws_cloudfront/index.ts
mtp1376: It is. I'm not sure if I follow, since it's being evaluated dynamically.

yrobla: ah ok ... that explain, if that's mandatory from AWS... looked as strange parameter to me

src/modules/aws_cloudfront/entity/distribution.ts
```diff
@@ -188,4 +189,20 @@ export class Distribution {
     nullable: true,
   })
   status?: string;
+
+  /**
+   * @public
+   * Associate a certificate from AWS Certificate Manager. The certificate must be in the US East (N. Virginia) Region (us-east-1).
```
aguillenv, line 195: hmm, interesting, why is this? We might need to add a constraint or a trigger to check it.

mtp1376: That's AWS' constraint. Seems like `CHECK` [isn't able](https://stackoverflow.com/a/27107470/2594912) to enforce a cross-table constraint. However, it's possible to do it using `TRIGGER`s. I'll do it as a follow-up.

src/modules/aws_cloudfront/index.ts
```diff
@@ -117,17 +120,30 @@ class DistributionMapper extends MapperBase<Distribution> {
...
       OriginGroups: distributionConfig.DistributionConfig?.OriginGroups,
       Restrictions: distributionConfig.DistributionConfig?.Restrictions,
-      ViewerCertificate: distributionConfig.DistributionConfig?.ViewerCertificate,
     };
 
+    if (e.customSslCertificate?.arn)
+      req.ViewerCertificate = {
+        CloudFrontDefaultCertificate: false,
+        ACMCertificateArn: e.customSslCertificate.arn,
+        MinimumProtocolVersion: MinimumProtocolVersion.TLSv1_2_2021,
```
dfellis, line 137: Will we expose these parameters later, or is this one of those situations where you have to provide a specific value or it fails and the argument itself is deprecated?

dfellis:
Will a new test be added for this PR?

src/modules/aws_cloudfront/index.ts
mtp1376: Yes. I think we should expose those values later, but since they're rarely used I've not added support for those yet.

mtp1376:
> Will a new test be added for this PR?

I think it'd be to change the value and re-sync the state to see if the change persists. If you think that helps we can add. Not sure if we need that kind of test for each property.

dfellis:
> > Will a new test be added for this PR?
> 
> I think it'd be to change the value and re-sync the state to see if the change persists. If you think that helps we can add. Not sure if we need that kind of test for each property.

Yes, we should have at least a smoke test for this functionality.

aguillenv (approved):

mtp1376:
> > > Will a new test be added for this PR?
> > 
> > 
> > I think it'd be to change the value and re-sync the state to see if the change persists. If you think that helps we can add. Not sure if we need that kind of test for each property.
> 
> Yes, we should have at least a smoke test for this functionality.

Oops. This one needs a test because it's more than just a property sync. Forgot that. I need to first import a certificate and then use it in the distribution to test it.

yrobla (approved):

aguillenv (approved):

PR #1840, opened 2023-01-23
fix: update lambda for cleaning

yrobla:


dfellis (approved):
It finally worked, but it doesn't seem to release accounts properly, so I think there's an issue with the lambda.

PR #1847, opened 2023-01-24
do not run django/prisma if there are no accounts

yrobla:


.github/workflows/prisma.yml
```diff
@@ -87,7 +87,6 @@ jobs:
           echo "::set-output name=account_index::$(echo ${ACCOUNT_INDEX})"
 
       - name: Setup IaSQL database
-        if: always()
         uses: nick-fields/retry@v2
```
aguillenv, line 90: This step should not be using the retry logic

aguillenv (approved):

mtp1376 (approved):

dfellis (approved):

PR #1849, opened 2023-01-24
`aws_sdk` Module: Invoke AWS SDK Requests Via IaSQL

mtp1376:
Example:
```sql
SELECT iasql_install('aws_sdk');

SELECT invoke_s3('listBuckets', '{}', 'us-east-1'); 

SELECT invoke_ec2(
    'describeVpcs',
    json_build_object(
        'VpcIds', (SELECT array(SELECT vpc_id FROM vpc WHERE region = 'eu-central-1' AND is_default))
    ),
    'eu-central-1'
);
```
Last query gives:
```json
{
  "Vpcs": [
    {
      "CidrBlock": "172.31.0.0/16",
      "DhcpOptionsId": "dopt-0440578b12fae1c05",
      "State": "available",
      "VpcId": "vpc-02b2a6c506171b3fd",
      "OwnerId": "680703303227",
      "InstanceTenancy": "default",
      "CidrBlockAssociationSet": [
        {
          "AssociationId": "vpc-cidr-assoc-019f46e758465b04e",
          "CidrBlock": "172.31.0.0/16",
          "CidrBlockState": {
            "State": "associated"
          }
        }
      ],
      "IsDefault": true
    }
  ]
}
```
https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#describeVpcs-property

More complex query:
```sql
-- describes all vpcs in all regions, grouped by their regions
SELECT agg.region, invoke_ec2('describeVpcs', json_build_object('VpcsIds', agg.vpc_ids), agg.region)
FROM (SELECT region, json_agg(vpc_id) as vpc_ids FROM vpc GROUP BY region) as agg;
```
returns:
image

invoke methods


src/modules/aws_ec2/rpcs/invoke.ts
```diff
@@ -0,0 +1,63 @@
...
+  RpcBase,
+  RpcResponseObject,
+} from '../../interfaces';
+
+/**
+ * Method for invoking an AWS EC2 SDK function
+ */
+export class InvokeRpc extends RpcBase {
+  preTransactionCheck = PreTransactionCheck.NO_CHECK;
+  postTransactionCheck = PostTransactionCheck.NO_CHECK;
```
aguillenv, line 20: Since this method could invoke whatever function and that could imply changes in the cloud I suggest running this without a transaction check, this would make it try to call `iasql_begin` before starting and then call `iasql_commit` when it's done:
```suggestion
```

mtp1376: Yes I'm going to do that. I just added those two for testing purposes to not slow down the testing process.

mtp1376: BTW now that I'm thinking, I think we should keep it as-is. The user can sync on demand, but having this would slow down the process for them and since this is a pro feature, we can communicate that they should sync the state periodically.

src/modules/aws_ec2/index.ts
```diff
@@ -22,6 +22,7 @@ export class AwsEc2Module extends ModuleBase {
 
   keyPairImport: KeyPairImportRpc;
   keyPairRequest: KeyPairRequestRpc;
+  invokeEc2: AwsSdkInvoker;
```
dfellis, line 25: I think these changes are also supposed to be reverted.

src/modules/aws_lambda/index.ts
```diff
@@ -9,12 +9,12 @@ export class AwsLambdaModule extends ModuleBase {
...
   lambdaFunction: LambdaFunctionMapper;
 
-  invokeLambda: LambdaFunctionInvokeRpc;
+  invokeLambdaFunction: LambdaFunctionInvokeRpc;
 
   constructor() {
     super();
     this.lambdaFunction = new LambdaFunctionMapper(this);
-    this.invokeLambda = new LambdaFunctionInvokeRpc(this);
+    this.invokeLambdaFunction = new LambdaFunctionInvokeRpc(this);
```
dfellis, line 17: Same here

src/modules/aws_sdk/migration/aws_sdk.ts
```diff
@@ -0,0 +1,10 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class awsSdk implements MigrationInterface {
+  // dummy migration
+  name = 'awsSdk';
```
dfellis, line 5: What does this line do?

src/modules/interfaces.ts
```diff
@@ -397,6 +401,162 @@ export class RpcBase {
...
+                ? `${rpcOutputEntries
+                    .map(([col, typ]) => `try_cast(j.s->>'${col}', NULL::${typ}) as ${col}`)
+                    .join(', ')}`
+                : ''
+            }
+          from (
+            select json_array_elements(_content::json) as s
+          ) as j;
+        end;
+        $$;
```
dfellis, line 524: Can you add another line here that defines a second function the same as the first, but auto-populates the third arg with `default_aws_region()`?

src/modules/interfaces.ts
```diff
@@ -397,6 +401,162 @@ export class RpcBase {
...
+      `;
+    const beforeUninstallSql = `
+        DROP FUNCTION "${snakeCase(key)}";
```
dfellis, line 527: If you do the function overloading suggested above, this now needs to be two `DROP FUNCTION` calls with the arg types following it, eg:

```suggestion
        DROP FUNCTION "${snakeCase(key)}"(varchar, json);
        DROP FUNCTION "${snakeCase(key)}"(varchar, json, varchar);
```

src/modules/aws_lambda/index.ts
dfellis: Ohh. Got it.

src/modules/aws_ec2/rpcs/invoke.ts
aguillenv: hmm, I understand it will delay a little bit the process but I think it could be risky because this lead to a case where the user can do 2 operations at the same time, while this rpc could be creating something in the cloud another `apply` could be running and deleting what has just been created. 

The right  solution actually is to use: 
- `preTransactionCheck = PreTransactionCheck.WAIT_FOR_LOCK;`
- `postTransactionCheck = PostTransactionCheck.UNLOCK_ALWAYS;`

With these, it will be safe to run in a transaction but it will not add the delay for the `commit`

dfellis:
Can you add one or two of the example SDK calling queries we discussed on Discord to a new test for the new module?

src/modules/aws_sdk/migration/aws_sdk.ts
```diff
@@ -0,0 +1,10 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
aguillenv, line 1: do we need this?

src/modules/aws_ec2/migration/1674575646446-aws_ec2.ts
```diff
@@ -0,0 +1,165 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
aguillenv, line 1: Do we need this migration change?

src/modules/interfaces.ts
```diff
@@ -397,6 +401,162 @@ export class RpcBase {
...
+      `;
+    return { beforeUninstallSql, afterInstallSql };
+  }
+}
+
+export class AwsSdkInvoker extends RpcBase {
+  preTransactionCheck = PreTransactionCheck.NO_CHECK;
+  postTransactionCheck = PostTransactionCheck.NO_CHECK;
+
```
aguillenv, line 467: As commented [here](https://github.com/iasql/iasql-engine/pull/1849#discussion_r1086934117) I think this should be:
```suggestion
  preTransactionCheck = PreTransactionCheck.WAIT_FOR_LOCK;
  postTransactionCheck = PostTransactionCheck.UNLOCK_ALWAYS;

```

src/modules/aws_ec2/index.ts
```diff
@@ -1,4 +1,4 @@
-import { ModuleBase } from '../interfaces';
+import { AwsSdkInvoker, ModuleBase } from '../interfaces';
```
dfellis, line 1: This line should also be reverted, I think.

src/modules/aws_ec2/migration/1674575646446-aws_ec2.ts
dfellis: We shouldn't, anymore. Probably easier to just `git checkout main -- src/modules/aws_ec2` to undo all of this.

src/modules/interfaces.ts
mtp1376: You're right. I changed it to the way you suggested.

src/modules/aws_sdk/migration/aws_sdk.ts
mtp1376: Yes it does. If it doesn't exist the system will break. So I created this dummy migration.

dfellis (approved):

depombo (approved):
nice!

mtp1376:
TODO:
- [ ] Add hints for the `method_name` argument.
- [ ] Add links in the typedoc

I'll do the above in a follow-up PR.

PR #1852, opened 2023-01-24
fix ec2 typewriter queries

depombo:


README.md
```diff
@@ -21,7 +21,7 @@
...
+<a href="https://app.iasql.com#/button/INSERT%20INTO%20instance%20%28ami%2C%20instance_type%2C%20tags%29%0A%20%20VALUES%20%28%27resolve%3Assm%3A%2Faws%2Fservice%2Fcanonical%2Fubuntu%2Fserver%2F20.04%2Fstable%2Fcurrent%2Famd64%2Fhvm%2Febs-gp2%2Fami-id%27%2C%20%27t2.micro%27%2C%20%27%7B%22name%22%3A%22i-1%22%7D%27%29%3B%0A%0AINSERT%20INTO%20instance_security_groups%20%28instance_id%2C%20security_group_id%29%20SELECT%0A%20%20%28SELECT%20id%20FROM%20instance%20WHERE%20tags%20-%3E%3E%20%27name%27%20%3D%20%27i-1%27%29%2C%0A%20%20%28SELECT%20id%20FROM%20security_group%20WHERE%20group_name%3D%27default%27%20AND%20vpc_id%20%3D%20%28SELECT%20id%20FROM%20vpc%20WHERE%20is_default%20%3D%20true%20AND%20region%3Ddefault_aws_region%28%29%29%29%3B">
```
aguillenv, line 24: Should we add the `SELECT * FROM iasql_install('aws_ec2');` to this? otherwise, the query will fail for a new database, and if they select to install the module, the query will be replaced.

dfellis: To the button link I think yes, the animation no, as it would be distracting?

depombo: > To the button link I think yes, the animation no, as it would be distracting?

Agreed, done

aguillenv (approved):

dfellis (approved):

PR #1855, opened 2023-01-25
Revoke permissions for iasql_audit_log table

### LINKED ISSUE #1842: Do not allow users that are not admin to delete the `iasql_audit_table`

aguillenv:
Resolves #1842


src/services/db-manager.ts
```diff
@@ -83,6 +83,7 @@ export function grantPostgresGroupRoleQuery(user: string, dbId: string) {
   const groupRole = getGroupRole(dbId);
   return `
     GRANT SELECT, INSERT, UPDATE, DELETE ON ALL TABLES IN SCHEMA public TO ${groupRole};
+    REVOKE INSERT, UPDATE, DELETE ON iasql_audit_log FROM ${groupRole};
```
dfellis, line 86: Perhaps it would make sense to have this in the `iasql_platform` module instead of here?

aguillenv: where exactly? does not need to be here per user basis? since we are granting everything for all tables in public? Or should we move the audit log to another schema?

depombo: I think so, all those tables should be out of bounds

aguillenv:
Should I delete this test since this should not happen? https://github.com/iasql/iasql/actions/runs/4015278274/jobs/6898902736#step:9:1420

depombo:
> Should I delete this test since this should not happen? https://github.com/iasql/iasql/actions/runs/4015278274/jobs/6898902736#step:9:1420

this is necessary to test upgrade functionality, no? this should be allowed by the admin credentials, just not the user scoped ones?

depombo:
> Should I delete this test since this should not happen? https://github.com/iasql/iasql/actions/runs/4015278274/jobs/6898902736#step:9:1420

this is necessary to test upgrade functionality, no? this should be allowed by the admin credentials, just not the user scoped ones?

aguillenv:
> > Should I delete this test since this should not happen? https://github.com/iasql/iasql/actions/runs/4015278274/jobs/6898902736#step:9:1420
> 
> this is necessary to test upgrade functionality, no? this should be allowed by the admin credentials, just not the user scoped ones?

yes, my bad, it was just to be adapted to be run by the admin and not any user

depombo (approved):

src/modules/iasql_platform/sql/after_install.sql
```diff
@@ -0,0 +1,30 @@
+ -- ALTER DEFAULT PRIVILEGES allows you to set the privileges that will be applied to objects created in the future. (https://www.postgresql.org/docs/current/sql-alterdefaultprivileges.html)
+ALTER DEFAULT PRIVILEGES
+  FOR USER CURRENT_USER IN SCHEMA public
+GRANT
+SELECT
+,
+  INSERT,
+UPDATE
+,
+  DELETE ON TABLES TO PUBLIC;
```
dfellis, line 10: Nit: What's with the weird formatting here?

dfellis (approved):

yrobla (approved):

PR #1857, opened 2023-01-25
add pricing page

depombo:
Screenshot 2023-01-25 at 4 15 08 PM


dfellis (approved):

aguillenv:
after going multi-region just installing the vpc module adds around 240 records. Not sure if still makes sense to have the limit of 500 records and we will see or set it a bit higher?

PR #1880, opened 2023-01-28
feat: add new UI changes for documentation

yrobla:


yrobla:
image
image
image

site/docusaurus.config.js
```diff
@@ -21,8 +21,8 @@ const theButton = require('./src/rehype/thebutton');
     url: config.url,
     baseUrl: '/',
     trailingSlash: true,
-    onBrokenLinks: 'throw',
-    onBrokenMarkdownLinks: 'throw',
+    onBrokenLinks: 'ignore', // we need to allow temporarily until the path structure for current version and next version converge, otherwise it fails on references
```
depombo, line 24: so there will be a PR that sets this back?

depombo (approved):

site/docusaurus.config.js
yrobla: yes, as soon as the stable documentation has the same paths as the next one, i can issue a PR to return again to throw.

aguillenv (approved):

PR #1893, opened 2023-01-31
Add route mapper

### LINKED ISSUE #1845: Changes done to entities that don't own a mapper are not being detected

aguillenv:
Resolves #1845


yrobla (approved):

src/modules/aws_vpc/mappers/route.ts
```diff
@@ -154,15 +154,17 @@ export class RouteMapper extends MapperBase<Route> {
...
+        es.map(async e => {
+          if (e.gatewayId === 'local' && ctx.memo?.db?.Route?.[this.entityId(e)]) {
+            // created by AWS, can't be deleted by the user but we need to remove it from the memo
+            delete ctx.memo.db.Route[this.entityId(e)];
+            return;
+          } else if (e.gatewayId === 'local') return;
+          const client = (await ctx.getAwsClient(e.region)) as AWS;
+          await this.deleteRoute(client.ec2client, e);
+        }),
+      );
```
dfellis, line 167: Why is this being done in parallel? We try not to do that so we don't slam into AWS API rate limiting.

aguillenv: I was keeping the implementation close to the previous one. Also, I do not expect to have a lot of deletes and most of them will try to delete the `local` one that is in all routetables and that one will never reach the AWS api. But I can serilize this to avoid issues.

src/modules/aws_vpc/entity/route.ts
```diff
@@ -23,124 +24,122 @@ export class Route {
...
-   * A managed prefix list is a set of one or more CIDR blocks.
-   * @see https://docs.aws.amazon.com/vpc/latest/userguide/managed-prefix-lists.html
-   */
-  @Column({ nullable: true })
-  @cloudId
-  DestinationPrefixListId?: string;
+  @Check(`
+    "destination" ~ '^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)/(3[0-2]|[12][0-9]|[0-9])$'
+    OR "destination" ~ '^(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:))/(12[0-8]|1[01][0-9]|[1-9]?[0-9])$'
+    OR "destination" LIKE 'pl-%'`)
```
dfellis, line 50: Nit: put the `"destination" LIKE 'pl-%'` at the beginning to short-circuit the cheapest check first

src/modules/aws_vpc/mappers/route.ts
```diff
@@ -0,0 +1,217 @@
...
+  equals = (a: Route, b: Route) => {
+    return this.eqRoute(a, b);
+  };
+  cidrIPv4Pattern = new RegExp(
+    '^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?).){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)/(3[0-2]|[12][0-9]|[0-9])$',
+  );
+  cidrIPv6Pattern = new RegExp(
+    '^(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1,4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:))/(12[0-8]|1[01][0-9]|[1-9]?[0-9])$',
+  );
+  prefixedListPattern = new RegExp('^pl-[wd]+$');
```
dfellis, line 38: Nit: This check is slightly stricter than the postgres check, that could cause records that pass in the DB but fail in the engine

src/modules/aws_vpc/entity/route.ts
aguillenv: I have my doubts about that here because most of the time the `destination` will be a cidr and not an aws prefixed list. But following your line of thinking I think it would make sense to put it in the second position since maybe ipv6 cidrs are less frequent than aws prefixed lists?

depombo (approved):

src/modules/aws_vpc/entity/route.ts
```diff
@@ -23,124 +24,123 @@ export class Route {
...
-   * @public
-   * A managed prefix list is a set of one or more CIDR blocks.
-   * @see https://docs.aws.amazon.com/vpc/latest/userguide/managed-prefix-lists.html
-   */
-  @Column({ nullable: true })
-  @cloudId
-  DestinationPrefixListId?: string;
+  @Check(`
+    "destination" ~ '^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)/(3[0-2]|[12][0-9]|[0-9])$'
+    OR "destination" LIKE 'pl-%'
```
dfellis, line 49: I was thinking of this one as the *first* in the list since it's the fastest to check.

dfellis (approved):

PR #1902, opened 2023-02-02
Fix local dev Chrome debugging of the engine

dfellis:


depombo (approved):

src/config/dev.ts
```diff
@@ -0,0 +1,27 @@
...
+    multiUser: false,
+  },
+  logger: {
+    debug: true,
+    test: false,
+    forceLocal: false,
+  },
+  telemetry: {
+    posthogKey: 'phc_xvAQWfpHug7G0SuU5P9wwAbvP9ZawgAfIEZ9FUsiarS',
+  },
```
aguillenv, line 24: do we want telemetry for dev?

aguillenv (approved):

src/config/dev.ts
depombo: no, good point

dfellis: I had just copy-pasted this from the `local` environment, because that's what we were using for the `dev` builds before. But I can remove it here.

dfellis:
```suggestion
```

PR #1903, opened 2023-02-02
feat: add new top level module page

yrobla:


yrobla:
image

src/modules/aws_ec2_metadata/index.ts
```diff
@@ -118,4 +118,29 @@ class AwsEc2MetadataModule extends ModuleBase {
     super.init();
   }
 }
+
+/**
+ * ## Code examples
+ *
+ *
```
depombo, line 125: any reason for the extra space?

depombo (approved):

src/modules/aws_ec2_metadata/index.ts
yrobla: formatting... otherwise lint doesn't pass

aguillenv (approved):

PR #1907, opened 2023-02-02
Add signature for RPC functions

mtp1376:
Probably I've not taken the best approach, so I'd really appreciate your feedbacks. This is what I've done:
- Add `inputTable` field to the RpcBase class. It's from `RpcInput` type which is a list of different things needed for [creation of Postgres function input](https://www.postgresql.org/docs/current/sql-createfunction.html):
```typescript
export type RpcInput = { ArgMode?: string; ArgName: string; ArgType: string; Default?: string }[];
```
```
( [ [ argmode ] [ argname ] argtype [ { DEFAULT | = } default_expr ] [, ...] ] )
```

- Create the input strings based on the above syntax and put it into `rpcInputArgs` variable. 
- If there's just a single `VARIADIC` input, just pass it along to `http_post` (and not wrap into another `ARRAY`), but if the arguments are different, make a new `SELECT array[${rpcInputArgsForPost}]::varchar[]` out of them.
- Name of the inputs and the outputs can't be the same, so I've added checks in our code so that we don't have to wait for Postgres to trigger the error.

### DIFF
--- a/src/modules/aws_acm/rpcs/import.ts
+++ b/src/modules/aws_acm/rpcs/import.ts
@@ -35,10 +35,10 @@ export class CertificateImportRpc extends RpcBase {
   } as const;
 
   inputTable: RpcInput = [
-    { ArgName: 'certificate', ArgType: 'varchar' },
-    { ArgName: 'privateKey', ArgType: 'varchar' },
-    { ArgName: 'region', ArgType: 'varchar' },
-    { ArgName: 'options', ArgType: 'varchar' },
+    { argName: 'certificate', argType: 'varchar' },
+    { argName: 'privateKey', argType: 'varchar' },
+    { argName: 'region', argType: 'varchar' },
+    { argName: 'options', argType: 'varchar' },
   ];
 
   /**
--- a/src/modules/aws_acm/rpcs/import.ts
+++ b/src/modules/aws_acm/rpcs/import.ts
@@ -33,7 +33,9 @@ export class CertificateImportRpc extends RpcBase {
     status: 'varchar',
     message: 'varchar',
   } as const;
-
+  /**
+   * @internal
+   */
   inputTable: RpcInput = {
     certificate: 'varchar',
     privateKey: 'varchar',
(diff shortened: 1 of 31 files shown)


src/modules/aws_acm/rpcs/import.ts
```diff
@@ -40,6 +40,13 @@ export class CertificateImportRpc extends RpcBase {
     message: 'varchar',
   } as const;
 
+  inputTable: RpcInput = [
```
yrobla, line 43: currently the documentation for rpcs was quite manual, explaining input/outputs just in manual comments. Have you tried adding comments at this level to see how they display, and if they may be more informative than what we have now?

yrobla (commented):
comments about doc

src/modules/aws_acm/rpcs/import.ts
mtp1376: I'll investigate that as a follow-up. Is that ok?

yrobla: sure

yrobla (approved):

mtp1376:
Talks with David on this:
https://discord.com/channels/920684574231646229/949065192275316736/1070765778057580675

dfellis (approved):

src/modules/aws_acm/rpcs/request.ts
```diff
@@ -53,6 +53,13 @@ export class CertificateRequestRpc extends RpcBase {
     message: 'varchar',
   } as const;
 
+  inputTable: RpcInput = {
```
aguillenv, line 56: I'm not completely sure, but not adding the `@internal` decorator is the same as having it? or it will be taken into account for the docs generation?

aguillenv (approved):

src/modules/aws_acm/rpcs/request.ts
mtp1376: This is how it's shown currently:
image

So you're right. I should add that decorator.

aguillenv: oh nice, this is good to know!

Issue #1870, opened 2023-01-27
Investigate why install is taking longer after multiregion/two way mode


depombo:

aguillenv:
The context of my tests:
- The installations occurred on a newly created db
- One test installing `aws_ec2` and another installing `aws_ecs_fargate`.

Current `main` install time:
- aws_ec2: 1.1min
![image](https://user-images.githubusercontent.com/5898639/216382918-5f0e8137-1115-4a9e-ae9c-daeaade9aa1a.png)

- aws_ecs_fargate: **3.2 min!!!!**
![image](https://user-images.githubusercontent.com/5898639/216384385-befa02c3-250e-414e-a087-80330b2d230b.png)

Issues seen:
- We have some mappers taking ~2s each doing a `cloud.read` (log group, metrics alarm, repository, vpc, iam user, sg rules, listener)
- We have mappers taking 3s to 4s each doing the `cloud.read` (iam role, access key, target groups)
- We have mappers taking more than 10s each running the `cloud.read`. (certificates ~10s, ecs cluster ~17s, ecs service ~17s, ecs task def ~20 s)
- We also have ~7 to ~18s delay **on every install for the `aws_vpc` module** since it usually tries to install subnets first and always fail since the VPCs are not in the db yet... Once the VPCs are in the db there's no issue.
- On every install we are running `iasql_commit` at the end unnecessarily. This is adding ~30s for my tests.

I have a branch where I have improved the times as follows:
- aws_ec2: 42s
![image](https://user-images.githubusercontent.com/5898639/216369054-e57ef788-62f1-4ae5-a2b4-d17d9a41f0de.png)

- aws_ecs_fargate: 51s
![image](https://user-images.githubusercontent.com/5898639/216369428-ed611373-3af8-4eb1-bff5-ba0a325f1647.png)

What I have done to reduce this time:
- Update cluster, service, task def and certificate to do the `cloud.read` in parallel for all enabled regions.
- Define the proper pre- and post-transaction hooks for `iasql_install` RPC. There's no need to run `iasql_commit` at the end of the install, we just need to close the transaction.

Some other  things we could do to improve:
- Going through the mappers taking between ~2s and ~5s to see if they can perform better somehow.
- I'm proposing to add a new `order` property to `MapperBase` or a new property defined in the `module.json` to be able to sort the mappers and make it easier for the lazy loader. With something like this, we can reduce the ~10s per `aws_vpc` install due to the subnets issue.

[...]

PR #1908, opened 2023-02-02
Initial approach to reduce install times

aguillenv:
Part of #1870

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ name: CI Tests
 # events but only for the main branch.
 on:
   push:
-    branches: [main]
+    branches: [main, install-times]
     paths:
       [
         '.github/workflows/ci.yml',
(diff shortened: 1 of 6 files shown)


depombo:
what about other places like the repository images?

aguillenv:
> what about other places like the repository images?

what do you mean about repository images? As I said in the issue #1870 these are the changes done for now to reduce the times as shown since these mappers were the worst performers. The other modules might be follow-up tasks?

depombo:
> what do you mean about repository images? 

https://github.com/iasql/iasql/blob/main/src/modules/aws_ecr/index.ts#L482. I think this one runs sequentially.

> As I said in the issue #1870 these are the changes done for now to reduce the times as shown since these mappers were the worst performers. The other modules might be follow-up tasks?

ah got it, a follow-up makes sense.

depombo (approved):

mtp1376 (approved):

PR #1910, opened 2023-02-02
Memory Recovery Hackery

dfellis:
Just an experiment to run on staging. I hope we don't have to merge (a slightly better version of) this.


mtp1376 (approved):
This is tempoarily, right?

dfellis:
This was just a Plan B approach. Properly fixed in https://github.com/iasql/iasql/pull/1911

Issue #1622, opened 2022-12-01
Add inline code samples documentation for missing modules


depombo:
https://github.com/jbreckmckye/testdoc looks like it might hold potential. https://github.com/jbreckmckye/testdoc/blob/master/src/api.ts#L16-L34 is the key bit. It's very fixed function, but it's internals look pretty clean, so I think we could use them directly. Parse with the typescript parser, grab the AST directly and generate our own markdown

yrobla:
My idea would be to reuse the current plugin that we are using for documentation and add those queries , to replace the current @example ones that we do manually. We could have a custom annotation where we pass the reference to the lines of code to use, and generate the snippet for us.

PR #1930, opened 2023-02-06
feat: generate code examples automatically from tests

yrobla:
Closes: #1622


yrobla:
image

depombo:
Hard to appreciate from the screenshot, but if the code samples are getting buried down it might make sense to create another top-level section in the TOC called `Sample Queries`. `Reference` can stay as-is and `Sample Queries` can list every module as well, but include the queries only. That way we can also exclude the `Code Samples` title too. Something like this:

```

Sample Queries >
   iasql_functions
   aws_account
Reference >
   iasql_functions
   aws_account

```

More of a thought than a proper suggestion. Curious to hear your opinion @yrobla. It is hard for me to know if this is substantially more difficult or makes the code structure more logical

test/modules/aws-codepipeline-integration.ts
```diff
@@ -709,7 +709,7 @@ describe('AwsCodepipeline Integration Testing', () => {
     try {
       query(
         `
-      UPDATE pipeline_declaration SET region='${nonDefaultRegion} WHERE name='${prefix}-${dbAlias}');
+      UPDATE pipeline_declaration SET region='${nonDefaultRegion}' WHERE name='${prefix}-${dbAlias}');
```
aguillenv, line 712: I wonder how was this working 👀

yrobla:
image

yrobla:
> 

i divided into folders, and added collapsible content, as we discussed

test/modules/aws-cloudfront-integration.ts
```diff
@@ -170,7 +170,7 @@ describe('Cloudfront Integration Testing', () => {
       () => ({ username, password }),
     )((e?: any) => {
       if (!!e) return done(e);
-      done();
+      done(undefined);
```
aguillenv, line 173: why is this change needed?

depombo (approved):

test/helpers.ts
```diff
@@ -200,3 +200,7 @@ export function defaultRegion(overrideList?: string[]): string {
   ];
   return regionList[Math.floor(Math.random() * regionList.length)];
 }
+
+export function itDocs(...args: any[]): any {
+  return (it as unknown as (...args: any[]) => {})(...args);
```
aguillenv, line 205: Maybe this solve then issue of the `undefined`s around on the `done` functions?
```suggestion
  return (it as unknown as (...args?: any[]) => {})(...args);
```
(just thinking out loud)

yrobla: no, i tried already... A rest parameter cannot be optional.ts(1047)

dfellis (approved):
Hope the new docs work in Dark Mode :)

PR #1934, opened 2023-02-06
add dashboard

depombo:


Dockerfile
```diff
@@ -20,14 +20,44 @@ RUN ["bash", "-c", "curl -fsSL https://deb.nodesource.com/setup_16.x | bash -"]
...
+COPY dashboard/yarn.lock yarn.lock
+RUN yarn install
+
+COPY dashboard/assets assets
+COPY dashboard/public public
+COPY dashboard/tsconfig.json tsconfig.json
+COPY dashboard/tailwind.config.js tailwind.config.js
+COPY dashboard/craco.config.js craco.config.js
+COPY dashboard/src src
+ENV REACT_APP_IASQL_ENV=local
```
dfellis, line 36: Perhaps we should move the `IASQL_ENV` definition below up here and then have this use that as the default value if not specified? (Not a blocker on this PR, thinking out lout)

docker-entrypoint.sh
```diff
@@ -2,6 +2,17 @@
 
 set -vex
 
+# Dashboard
+cd /dashboard
+REACT_APP_IASQL_ENV=local yarn start:build &
+
+# Run service
+cd /dashboard/run
+IASQL_ENV=local yarn start &
```
dfellis, line 11: Similarly, these should just use the env vars that were passed in to the Docker container. But won't block this PR on it.

dfellis (approved):

PR #1936, opened 2023-02-07
push to dockerhub on release

depombo:


.github/workflows/release-dockerhub.yml
```diff
@@ -0,0 +1,40 @@
+name: Publish Docker image
+
+on:
+  workflow_dispatch:
+    tag:
+      description: 'Tag for docker image'
```
aguillenv, line 6: whats the difference between this one and the one under `inputs`?

PR #1937, opened 2023-02-07
RFC 008: Pure-SQL Modules

mtp1376:


rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+- [ ] Implemented: [One or more PRs](https://github.com/iasql/iasql-engine/some-pr-link-here) YYYY-MM-DD
+- [ ] Revoked/Superceded by: [RFC ###](./000 - RFC Template.md) YYYY-MM-DD
+
+## Author(s)
+
+- Mohammad Teimori Pabandi <m@iasql.com>
+
+## Summary
+
+Having Pure-SQL modules would help us expand more. One key advantage is that we don't need a PR on our engine to add coverage for a new service, we just use a syntax like this and everything JustWorks™.
```
depombo, line 28: > Having Pure-SQL modules would help us expand more. 

We should better explain what values lie behind unlocking this. It would be great to get an understanding of how this ties to `aws_sdk` vs other AWS modules as well.

depombo, line 28: > Having Pure-SQL modules would help us expand more. 

We should better explain what values lie behind unlocking this. It would be great to get an understanding of how this ties to `aws_sdk` vs other AWS modules as well.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+
+A positive point for having standalone mappers is that it's language agnostic: mapper developers can write the logic in whatever language they want, and it's fine, until they follow the guidelines for the Engine -> Mapper communication protocol.
+
+One negative point is that they'll be provided with the AWS credentials, and they can perform out-of-band data extraction on it on both of the above ways.
+
+### In-IaSQL-Container Mappers
+
+In this approach, the logic of the mappers will be executed on IaSQL: either on the Postgres server or on the Engine. Our current mappers are under this category.
+- Mapper logic is in IaSQL engine: `aws_acm`'s `CertificateMapper`, `aws_ecr`'s `RepositoryImageMapper`, etc.
+- Mapper logic is in Postgres server: This is where we initially called Pure-SQL module. Note that `aws_ecs_simplified` module can't be considered under this category, since it introduces no new mappers. It's playing in module-level (not mapper level) and handling a new Entity using already-existing mappers.
```
depombo, line 73: Depending on how loosely you define a mapper, I would say that the function the triggers call are acting as mappers themselves. They are mapping an existing entity to another entity. However, my understand was that we were trying to get a JSON response from `aws_sdk` into a mapper.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+
+Before diving into "Pure-SQL Modules" category, first let's see if our current mappers are surjective to the space of "Mapper logics inside IaSQL engine".
+
+#### Mapper Logic Inside IaSQL Engine
+
+Currently, we're developing our mappers in this category. But there might be other ways to integrate the logic inside the engine, without having to exactly follow our current pattern. I'm getting inspired by [Strapi](https://github.com/strapi/strapi), which is a headless CMS. You can create models using a GUI from Strapi's admin panel and the database will be changed on-air, new API routes will be added with proper logic, and it achieves these without losing performance or being in need of pushing new files to the codebase (but the codebase is changing on-air).
+
+So we can get inspired and create on-air mapper codes. But to ensure safety, we might take one of the two approaches: either by narrowing down the available variables, or the by narrowing down the available functionality.
```
yrobla, line 81: i like this idea of code generators, but the AWS SDK is very complex in terms of methods and types. What effort do you think it's needed to achieve that? not even talking about the GUI, but just some language based on yaml for example?

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+    - `{entity}_cloud_update(old: json, new: json)`
+    - `{entity}_cloud_delete(object: json)`
+- Call other mappers: 
+  - `call_mapper('{module_name}', '{entity_name}', 'cloud|db', 'create|read|update|delete', {}::json) -> json`
+  - It would be easy to add support for normal modules with the above definition. It will call `ec2_cloud_read` if that Postgres function exists. Otherwise, it'll send an RPC to the engine to ask engine to call that function from the mapper.
+
+We also need some other things for this to work:
+- Dynamically creating a module that provides the same interface as `MapperBase`, but calls corresponding SQL functions. 
+
+### Alternatives Considered
```
depombo, line 107: Technically the options above are the alternatives considered and once we settle on one the rest go here? There is an alternative I am surprised to not see here- a cron job that periodically calls `aws_sdk` and populates some tables/view from the JSON payload.

rfcs/007 - Alternative Ways to Create Mappers.md
mtp1376: I'm defining a mapper to have these properties:
- A cloud id function
- An equality checker function
- An entity, that the mapper manages it
- Cloud CRUD functions
- Ability to call other mappers' CRUD functions

In that definition, I mean, the `ecs_simplified` does not fit since it doesn't provide a new mapper.

rfcs/007 - Alternative Ways to Create Mappers.md
mtp1376: We've already achieved that functionality in `aws_sdk` module. We can dynamically call AWS SDK methods and receive the results as JSON (inputs are JSON, as well).

I think the main issue would be how that DSL looks like. Do we want to take the burden of designing the DSL, and then write compilers for that new language, etc. 

But I added that here because David, Luis, and Alejandro have worked on Alan Language and probably this isn't something hard for them.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+  - The only issue that comes to my mind is that we will be vulnerable to SSRF attacks, which can be dangerous [in EC2](https://scalesec.com/blog/exploit-ssrf-to-gain-aws-credentials/). Our other vulnerability to SSRF would be related to the engine's Express server, which will be accessible by the running code. Therefore, we'll need authorization measures for that internal API.
+- Providing a new "safe" Domain Specific Language for IaSQL: developers can use this language to create new mappers, and we can translate that to TypeScript code on-air, while narrowing down the DSL interface to something that is "safe" (no code-execution, no unwanted database access, etc). The DSL does not have to be capable of creating all possible mappers (at least not at first) since if they want to do something very professional, they can just submit a PR to the engine.
+  - For example: we can supply a base provider like AWS. And using DSL people can dynamically invoke AWS client's methods and write their logic. The DSL then gets parsed and converted to JavaScript code, and we'll use its interface when we need the mapper.
+
+
+#### Mapper Logic Inside Postgres
```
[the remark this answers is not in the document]
mtp1376: Agreed. Thanks for commenting this, though. So I think I should focus more on this part and try to introduce more alternatives under this section.

rfcs/007 - Alternative Ways to Create Mappers.md
mtp1376: Yes. This PR is still in draft mode and I've not yet decided on which to go with. So I wanted some input from you to know which part I should focus on.

But a problem on my side was that I didn't remove this section. It's a leftover from the last commit. I'll delete it and just keep the ideas as siblings.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+  - `call_mapper('{module_name}', '{entity_name}', 'cloud|db', 'create|read|update|delete', {}::json) -> json`
+  - It would be easy to add support for normal modules with the above definition. It will call `ec2_cloud_read` if that Postgres function exists. Otherwise, it'll send an RPC to the engine to ask engine to call that function from the mapper.
+
+We also need some other things for this to work:
+- Dynamically creating a module that provides the same interface as `MapperBase`, but calls corresponding SQL functions. 
+
+### Alternatives Considered
+
+I also considered the functions to be strongly-typed inside Postgres. More info here:
+https://rounded-apology-58d.notion.site/What-Does-a-Mapper-Need-9402ce7e35244405876f4cf660365c86
```
dfellis, line 110: Why is this part in a separate site instead of being embedded markdown in this file?

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+
+- Mohammad Teimori Pabandi <m@iasql.com>
+
+## Summary
+
+Having Pure-SQL modules would help us expand more. One key advantage is that we don't need a PR on our engine to add coverage for a new service, we just use a syntax like this and everything JustWorks™.
+```postgresql
+SELECT iasql_install('https://github.com/someModuleDeveloper/awsEc2Module');
+```
+That's where we started, but it might be useful to assess different paths we might take for creation of an IaSQL module/mapper.
```
dfellis, line 32: It would be good for others and for future-us to include all of the reasons why we're considering this.

1. Enables 3rd Party Modules for an actual user ecosystem to exist.
2. Reduces pressure on us to cover 100% of all AWS services before getting our first customer, because users can customize their modules to their own needs.
3. Potentially improves development/testing turn-around times by segmenting the engine testing from module testing (changes to a module would only trigger tests within the module's repo, changes to the engine would still trigger all officially-supported modules for testing, but would potentially be done much less often, and primarily for confirming newer version(s) of the AWS SDK still work with our modules.
4. Adding support for other clouds/SaaSes would be simpler within the engine.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Proposal
+
+To create a module, we need a mapper. I'll continue this RFC with my proposed solution for mappers, since a module is a set of mappers, and entities (with a bit of oversimplification).
+
+First, let's see what we need to create a mapper:
+- A cloud id function
+- An equality checker function
+- An entity, that the mapper manages it
```
dfellis, line 41: The conclusion we came to earlier and that @aguillenv is working on to have 1 mapper per entity *ought* to make equality-checking a potentially automate-able behavior, then this would not need to be defined by hand.

We already have it for some entities, but we may be able to subsume it entirely into the engine.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+- Cloud CRUD functions
```
dfellis, line 42: We also have a few DB CRUD functions still required in the current engine. We need to evaluate if any of them would still be required once we're working *inside* Postgres, but it's possible.

rfcs/007 - Alternative Ways to Create Mappers.md
dfellis, line 41: Specifically for the line `An entity, that the mapper manages it` this is a good thing to call out. These modules won't have access to TypeORM.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+- Ability to call other mappers' CRUD functions
```
dfellis, line 43: We should quantify how many of these are actually necessary once we're inside of Postgres where we can literally just `SELECT * FROM other_entity WHERE some_condition = TRUE`. And perhaps triggering Cloud CRUD functions ought to be done by INSERT/UPDATE/DELETE on the database and letting the apply/commit logic just deal with it?

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+![In-house Standalone Mapper](./assets/008/Mapper-1.png)
+2. Standalone mappers served on servers other than IaSQL: the mapper providers can host that mapper anywhere, and provide endpoints to the engine which will be called. This keeps the maintenance of the mapper on the mapper developer side, but increases the latency of mapper.
+![Self-hosted Standalone Mapper](./assets/008/Mapper-2.png)
+
+In the 2nd way, we should post the AWS credentials to the mapper through the internet, which should be done in a secure way. In both ways I don't see another possible approach than posting the AWS credentials to the mapper, since it needs them to query AWS APIs. Anyway, we should warn people to not use a mapper from untrusted authors, because the credentials will be shared with them.
+
+A positive point for having standalone mappers is that it's language agnostic: mapper developers can write the logic in whatever language they want, and it's fine, until they follow the guidelines for the Engine -> Mapper communication protocol.
+
+One negative point is that they'll be provided with the AWS credentials, and they can perform out-of-band data extraction on it on both of the above ways.
```
dfellis, line 67: This is a pretty massive negative. With the other approaches the 3rd party module developers are constrained to the APIs we give them so exfiltrating secrets should be much more difficult, and potentially even a tractable problem we can lock down, while these two are basically the wild west -- end-users have to trust us *and* everyone who contributes code to the ecosystem? That's a much taller order.

dfellis: One other thing this doesn't cover is: *all* of the modules have some portion that runs within our engine and within Postgres -- they have to somehow transmit the relevant schema install/uninstall logic to define entities and the engine has to register a mapper object related to it and then funnel the DB/Cloud CRUD operations (and equality, etc) to these external mappers.

*That* means this entire approach, beyond being a worse security situation, is *also* much harder for us to build/maintain. This is also something worth calling out in the RFC.

rfcs/007 - Alternative Ways to Create Mappers.md
dfellis: Yes. It may be necessary to have a separate section on 3rd party modules vs 3rd party mappers, and also why 3rd party modules without 3rd party mappers was deemed insufficient. Specifically, `ecs_simplified` and other modules like it can only reduce down the functionality that has been already exposed through mappers we have written, it can't go beyond any functionality we didn't hand-expose, which makes us the bottleneck for the whole ecosystem.

Then it would be worth discussing why we need mapper-like behavior when the `aws_sdk` is now exposed directly. What do we lose if we just had a collection of triggers that directly call `aws_sdk` `invoke_` functions on INSERT/UPDATE/DELETE and execute in a cron job to do INSERT/UPDATE/DELETE calls into that table based on cloud changes.

(No eventual consistency, for one, but other issues arise, too. Would like it if you think about it. :) )

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+- Mapper logic is in IaSQL engine: `aws_acm`'s `CertificateMapper`, `aws_ecr`'s `RepositoryImageMapper`, etc.
+- Mapper logic is in Postgres server: This is where we initially called Pure-SQL module. Note that `aws_ecs_simplified` module can't be considered under this category, since it introduces no new mappers. It's playing in module-level (not mapper level) and handling a new Entity using already-existing mappers.
+
+Before diving into "Pure-SQL Modules" category, first let's see if our current mappers are surjective to the space of "Mapper logics inside IaSQL engine".
+
+#### Mapper Logic Inside IaSQL Engine
+
+Currently, we're developing our mappers in this category. But there might be other ways to integrate the logic inside the engine, without having to exactly follow our current pattern. I'm getting inspired by [Strapi](https://github.com/strapi/strapi), which is a headless CMS. You can create models using a GUI from Strapi's admin panel and the database will be changed on-air, new API routes will be added with proper logic, and it achieves these without losing performance or being in need of pushing new files to the codebase (but the codebase is changing on-air).
```
dfellis, line 79: This is still a huge security issue. With local usage it's kinda on the user if they loaded an untrustworthy module, but that's not a great story to tell them. But in the SaaS, if we don't *really* rethink how we handle the SaaS product a single user could compromise everyone. It would be good to bring up the pros and cons here, with this being a major con.

rfcs/007 - Alternative Ways to Create Mappers.md
dfellis: We can bring up the pros and cons of a DSL, it is what the original IaC Terraform did, but I feel like the cons will outweigh the pros.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+
+So we can get inspired and create on-air mapper codes. But to ensure safety, we might take one of the two approaches: either by narrowing down the available variables, or the by narrowing down the available functionality.
+- Create a sandbox and run mapper's JavaScript code on IaSQL engine: in that sandbox, we can narrow down the context and not pass variables that might cause problems (not sure if that's technically possible currently, but I've found see some links: [1](https://nodejs.org/api/vm.html#vm_vm_runinthiscontext_code_options), [2](https://github.com/patriksimek/vm2)).
```
dfellis, line 82: An alternative is to use WebWorkers and provide an event-based API into and out of the worker. We could expose something similar to the `invoke_` functions in Postgres as events and event payloads (also including a worker-defined unique token) and they call the same named event back into the worker with the result.

This would be much safer as a pure JS context not allowed to `require` or `import` anything, but is a very awkward experience for the authors. That awkwardness can be alleviated somewhat by providing our own SDK to take more normal-looking Typescript that automatically has some implicit global functions baked in and then generating the JS bundle out.

The entities and Postgres upgrade/downgrade code could also be written using this IaSQL SDK and perhaps structured fairly closely to the current module code...

This approach may have more legs to it than I originally anticipated, but would still keep 3rd Party Modules restricted to JS/TS code. This also has its pros and cons, but I would argue may be more pro than con?

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+  - The only issue that comes to my mind is that we will be vulnerable to SSRF attacks, which can be dangerous [in EC2](https://scalesec.com/blog/exploit-ssrf-to-gain-aws-credentials/). Our other vulnerability to SSRF would be related to the engine's Express server, which will be accessible by the running code. Therefore, we'll need authorization measures for that internal API.
```
dfellis, line 83: The WebWorker approach with a very constrained event-based communication protocol *shouldn't* be vulnerable to the SSRF issue as the `invoke_*` functions we currently expose in the `aws_sdk` don't require the AWS credentials to be passed through them, so we could keep those credentials hidden from the module authors.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+- Providing a new "safe" Domain Specific Language for IaSQL: developers can use this language to create new mappers, and we can translate that to TypeScript code on-air, while narrowing down the DSL interface to something that is "safe" (no code-execution, no unwanted database access, etc). The DSL does not have to be capable of creating all possible mappers (at least not at first) since if they want to do something very professional, they can just submit a PR to the engine.
```
dfellis, line 84: We'd be generating JS on the air, not TS, but the point stands.

The downsides of DSLs is that people have to learn your DSL. I think a TS SDK that compiles into a WebWorker JS bundle is functionally identical while having a superset of pros and a subset of cons as the DSL approach would require 3rd party contributors to learn a language that they cannot use anywhere else.

This is still definitely good to bring up in the RFC.

rfcs/007 - Alternative Ways to Create Mappers.md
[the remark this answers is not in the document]
dfellis: This is the approach that I was originally leaning towards, and I still think it has some solid long-term merit to it (will get to that in a bit) but it will require much more up-front effort than the TS/JS SDK into WebWorker approach, so it may not be the right choice for now.

The advantage of moving into pure Postgres is that we only need to write a `saas_account` and `saas_sdk` module in Typescript (and in this future we'd probably merge them together?) and then we can write just Postgres entirely outside of the engine and start implementing modules. No SDK *required*, and they just need a built set of code that can run inside of Postgres (whether it's pl/pgsql, pl/python, whatever) and could write the logic in anything that can eventually build such a block of Postgres for them.

But as shown below there are a *lot* of Postgres functions that need to be written per Postgres table managed in the mapper way. There is also the IaSQL RPC overhead to deal with here. Each `aws_sdk` invoke is converted into an http request into the engine, then executed, and then returned to Postgres to continue on. For AWS SDK calls that has been fine because those APIs are so slow, but this may be an awkward performance bottleneck for better SaaS APIs.

On top of that, the collection of functions are kept together just by a naming convention, which feels very C-like (versus the C++-like classes we currently use). Not *terrible*, but there's less feedback that a Postgres function is missing here than in Typescript where a missing method to implement causes your editor to highlight the class as incorrect and prod you into fixing it.

That also brings up the other general issue with pl/pgsql and friends: the actual developer experience is poorer than Typescript because there hasn't been a huge demand for a good experience. Most developers stick to ORMs (even we are, to a degree), and many SQL jockeys are from non-programming backgrounds, so they don't even know what they're missing there.

But for developers of a pure-SQL mapper, the lack of guidance from their IDE/text editor could make the experience much more grueling, especially with respect to the opaque `invoke_*` function calls that basically eat away the type safety of the AWS SDK.

I came into this review feeling like this was going to be the approach we go for, but it's much more mixed than I expected, and I am actually leaning towards the WebWorker JS bundle, now.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+    - `{entity}_cloud_create(object: json)`
+    - `{entity}_cloud_read() -> json list`
+    - `{entity}_cloud_update(old: json, new: json)`
+    - `{entity}_cloud_delete(object: json)`
+- Call other mappers: 
+  - `call_mapper('{module_name}', '{entity_name}', 'cloud|db', 'create|read|update|delete', {}::json) -> json`
+  - It would be easy to add support for normal modules with the above definition. It will call `ec2_cloud_read` if that Postgres function exists. Otherwise, it'll send an RPC to the engine to ask engine to call that function from the mapper.
+
+We also need some other things for this to work:
+- Dynamically creating a module that provides the same interface as `MapperBase`, but calls corresponding SQL functions.
```
dfellis, line 105: This is going to be required for *all* approaches, right? (Well, replace "SQL functions" with "WebWorker events" or "gRPC calls" etc)

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+  - For example: we can supply a base provider like AWS. And using DSL people can dynamically invoke AWS client's methods and write their logic. The DSL then gets parsed and converted to JavaScript code, and we'll use its interface when we need the mapper.
+
+
+#### Mapper Logic Inside Postgres
+
+We can put the mapper's logic inside Postgres. And then provide an interface for the engine to call mapper functions. In this approach, the interface will be provided through the Postgres, and therefore this is the approach that is most consistent with "Pure-SQL". The DSL idea from the above way can still be applied here (developer writes DSL, PL/pgSQL code is generated).
+
+So let's consider one possible implementation of the mapper structure (all as Postgres functions):
+- CloudId: `{entity}_cloud_id(object: json) -> string` a Postgres function that takes the object as json and returns its cloud id as string.
```
dfellis, line 93: Should they be `json` types or Postgres `Row` types where the sub types match the table definition of the entity?

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,122 @@
...
+
+A minor update. Just adding some Postgres functions, the ability to install a Pure-SQL module from remote link, and the engine having the ability to work with Pure-SQL modules. 
+
+## Affected Components
+
+Mostly the engine
+
+## Expected Timeline
+
+I believe it takes at least 2 weeks for us to create a working example of this and port an easy module to it.
```
dfellis, line 122: I think we could get something quicker than 2 weeks if we're okay with cut corners for the demonstration that we then clean up.

rfcs/007 - Alternative Ways to Create Mappers.md
mtp1376: > And perhaps triggering Cloud CRUD functions ought to be done by INSERT/UPDATE/DELETE on the database and letting the apply/commit logic just deal with it?

I think we still need to be able to call the mappers' functions directly. Yes, we can trigger cloud CRUD functions by manipulating the data inside Postgres, but we can't wait for the result (or I don't know of a way to do so) of those functions. That's kind of event-driven, but without the callback yet.

If a module's cloud CRUD functions depend on these functions on another module, then we'll need to provide that interface.

rfcs/007 - Alternative Ways to Create Mappers.md
mtp1376: The only issue that comes to my mind regarding this, is about FKs. Because we have some logics there.

mtp1376: For the current module, it can just perform `db.X` functions itself by directly executing SQL commands. For other modules, it can use the following function to access their objects.
`call_mapper('{module_name}', '{entity_name}', 'cloud|db', 'create|read|update|delete', {}::json) -> json`

dfellis: True, but we can inspect the schema of tables in Postgres itself using the [information_schema](https://stackoverflow.com/questions/1152260/how-to-list-table-foreign-keys#1152321) namespace and tables, so we *should* be able to replicate any meta-programming we're doing in Typescript in pure Postgres.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,159 @@
...
+- Mohammad Teimori Pabandi <m@iasql.com>
+
+## Summary
+
+Having Pure-SQL modules would help us expand more. One key advantage is that we don't need a PR on our engine to add coverage for a new service, we just use a syntax like this and everything JustWorks™.
+```postgresql
+SELECT iasql_install('https://github.com/someModuleDeveloper/awsEc2Module');
+```
+
+By the way we already have the `aws_ecs_simplified` module, and it's written purely in SQL. But it does not add a new mapper functionality. It maps a new entity to the entities that already exist, therefore leveraging those entities' mappers. Specifically, `ecs_simplified` and other modules like it can only reduce down the functionality that has been already exposed through mappers we have written, it can't go beyond any functionality we didn't hand-expose, which makes us the bottleneck for the whole ecosystem.
```
dfellis, line 33: Nit

```suggestion
We already have the `aws_ecs_simplified` module, and it's written purely in SQL, but it does not add a new mapper functionality. It maps a new entity to the entities that already exist, therefore leveraging those entities' mappers. Specifically, `ecs_simplified` and other modules like it can only reduce down the functionality that has been already exposed through mappers we have written, it can't go beyond any functionality we didn't hand-expose, which makes us the bottleneck for the whole ecosystem.
```

rfcs/007 - Alternative Ways to Create Mappers.md
dfellis: I think it is worth explicitly calling out in the RFC that the security profile of this approach is *much* worse than other approaches because you can't inspect the source code to make sure it's not doing anything malicious, so any responsible developer would have to assume malicious intent.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,159 @@
...
+- Registering/de-registering logic for the module entities be transmitted to the engine: security concerns here, since the module should not be able to `DROP` tables for other modules (for example).
+
+### Security Can't Be Guaranteed
+Suppose this scenario:
+> We have a new Pure-SQL module. It has access to do `INSERT`s and `SELECT`s on behalf of the user. So it can create a CodeBuild project that calls a curl, with the AWS credentials coming from the `aws_credentials` table, right? So I think security can't be guaranteed in neither of the above ways, given our tables keep the current permissioning system.
+
+One way to fix the above issue, is to create a new Postgres role for each user (like `{username}_module`) which will be used by the Pure-SQL modules to execute SQL commands. It won't have access to the `aws_credentials` table.
+
+But the above idea can be also bypassed:
+> Use `aws_sdk` module to create a new set of AWS AccessKeyId and SecretAccessKey. Then create a CodeBuild project to send them to the malicious endpoint :D.
```
dfellis, line 147: As we are discussing in chat, being explicit about module capabilities and dependencies up-front can help mitigate this. A third party module that depends on CodeBuild and AWS account key management APIs should declare that it needs these things and if they seem incongruous to the purpose of the module that's another flag for users.

We could also classify certain API calls as requiring manual confirmation because of their sensitivity (particularly the aws credential management APIs, but also anything that allows arbitrary code execution, such as codebuild or even the startup script you can give to an ec2 instance) as further hardening in the future.

rfcs/007 - Alternative Ways to Create Mappers.md
mtp1376: This idea was also checked, and it's going to need an extra casting step more than this one.

rfcs/007 - Alternative Ways to Create Mappers.md
mtp1376: One new thing here is for a situation where we're using `policiesAreEqual`. In that case, the exact content of two `policy` fields may not be equal, but they in-fact might be the same thing. Automating that might be a tricky thing. Or we'd need to provide a "customized" behavior for that case.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,165 @@
...
+- Call other mappers: 
+  - `call_mapper('{module_name}', '{entity_name}', 'cloud|db', 'create|read|update|delete', {}::json) -> json`
+  - It would be easy to add support for normal modules with the above definition. It will call `ec2_cloud_read` if that Postgres function exists. Otherwise, it'll send an RPC to the engine to ask engine to call that function from the mapper.
+
+We can also consider the above functions to be strongly-typed inside Postgres. More info here:
+https://rounded-apology-58d.notion.site/What-Does-a-Mapper-Need-9402ce7e35244405876f4cf660365c86
+
+##### Security Can't Be Guaranteed
+In the Pure-SQL way, it's not easy to guarantee security. Suppose this scenario:
+> We have a new Pure-SQL module. It has access to do `INSERT`s and `SELECT`s on behalf of the user. So it can create a CodeBuild project that calls a curl, with the AWS credentials coming from the `aws_credentials` table, right? So I think security can't be guaranteed in neither of the above ways, given our tables keep the current permissioning system.
```
aguillenv, line 138: Can't this happen with any of the alternatives exposed in this RFC?

aguillenv (commented):
So, correct me if I'm wrong but it seems that some of the takeaways from this RFC are:
- we will be enabling 3rd party modules
- it might make sense to create/provide a common interface and/or a set of tools for consistency and UX.
- in terms of security we might need to run this code in an isolated environment and we might need to "forbid" some tables, modules or functions that can expose sensitive information or can manipulate sensitive resources (or at least alert about it).

I think this RFC does an exhaustive analysis and expose important issues that maybe we haven't thought about before. We might come up with a good proposal here but IMO the implementation of this RFC would not occur in the immediate future since we do not have active users neither we have an active community wanting to contribute, yet. Also, it seems to me that it deviates from the original issue which was trying to find a way to create modules easier/faster.

dfellis:
> So, correct me if I'm wrong but it seems that some of the takeaways from this RFC are:
> 
>     * we will be enabling 3rd party modules
> 
>     * it might make sense to create/provide a common interface and/or a set of tools for consistency and UX.
> 
>     * in terms of security we might need to run this code in an isolated environment and we might need to "forbid" some tables, modules or functions that can expose sensitive information or can manipulate sensitive resources (or at least alert about it).
> 
> 
> I think this RFC does an exhaustive analysis and expose important issues that maybe we haven't thought about before. We might come up with a good proposal here but IMO the implementation of this RFC would not occur in the immediate future since we do not have active users neither we have an active community wanting to contribute, yet. Also, it seems to me that it deviates from the original issue which was trying to find a way to create modules easier/faster.

On the last point, the use of `aws_sdk` to make pure-sql modules was explored and found to 1) not be much easier/faster if we don't want the functionality to suffer (we could make blocking INSERT/UPDATE/DELETE operations, probably, but transactions wouldn't work at all in that structure) 2) it has the worst or second worst security profile of them all.

To cover your other comment [here](https://github.com/iasql/iasql/pull/1937#discussion_r1112068940) not exactly. The SDK or DSL approaches can limit access by literally making the ability to express certain data accesses impossible. The Pure-SQL approach gives the modules access to all of the same data the owner of the database has, so there's no good way to firewall sensitive data from them.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Common Parts
+We also need these added functionalities in order for any of the above solutions to work:
+- Dynamically creating a mapper that provides the same interface as `MapperBase`, but calls corresponding SQL functions.
+- Registering/de-registering logic for the module entities be transmitted to the engine: security concerns here, since the module should not be able to `DROP` tables for other modules (for example).
+- Dependency declaration mechanism: each module should tell IaSQL engine which modules it needs to work properly. And we should be able to get user's manual approval in case it's using sensitive modules. 
+
+## Expected Semver Impact
+
+A minor update. Just adding some Postgres functions, the ability to install a Pure-SQL module from remote link, and the engine having the ability to work with Pure-SQL modules.
```
dfellis, line 157: This comment is no longer correct if we're going with the SDK proposal.

rfcs/007 - Alternative Ways to Create Mappers.md
```diff
@@ -0,0 +1,165 @@
...
+
+Before diving into "Pure-SQL Modules" category, first let's see if our current mappers are surjective to the space of "Mapper logics inside IaSQL engine".
+
+### Mapper Logic Inside IaSQL Engine
+
+Currently, we're developing our mappers in this category. But there might be other ways to integrate the logic inside the engine, without having to exactly follow our current pattern. I'm getting inspired by [Strapi](https://github.com/strapi/strapi), which is a headless CMS. You can create models using a GUI from Strapi's admin panel and the database will be changed on-air, new API routes will be added with proper logic, and it achieves these without losing performance or being in need of pushing new files to the codebase (but the codebase is changing on-air).
+
+So we can get inspired and create on-air mapper codes. But to ensure safety, we might take one of the two approaches: either by narrowing down the available variables, or the by narrowing down the available functionality.
+- Create a sandbox and run mapper's JavaScript code on IaSQL engine: in that sandbox, we can narrow down the context and not pass variables that might cause problems.
```
dfellis, line 91: This option should be pulled out and turned into the Proposal with all of the rest put into Alternatives Considered.

dfellis: It would also be good to bring up the SDK vs external module install are orthogonal.

We can break this into a few different tasks:
- Creating the VM2-based code runner and RPC layer, which can be validated with a handwritten internal module to flesh it out, perhaps simply porting a smaller existing module, like the S3 module, and renaming it to `aws_s3_vm2` for this first piece.
- Creating the SDK that generates the bundle for VM2 to run, re-porting the chosen test module again, and seeing how close it looks to the original module.
- Modifying `iasql_install` to handle installing modules from a URL, where the URL is pointing to the bundled `.js` file that the SDK can generate (perhaps with some sort of signing mechanism to be more sure it was built via the SDK?) and the relevant changes necessary to deal with dependency declaration and management in that world.

Technically each one depends on the one before it, but a lot of the foundational work of those three main tasks can be done in parallel, but finishing it off requires the first to be done first, second done second, etc.

dfellis (commented):
@depombo @aguillenv @yrobla even if we aren't going to implement this RFC in the near future, it would be good if we're all in agreement on what will be done here, I think. Can you review my last few comments and let us know if you agree with that plan?

rfcs/007 - Alternative Ways to Create Mappers.md
aguillenv: > * Creating the VM2-based code runner and RPC layer, which can be validated with a handwritten internal module to flesh it out, perhaps simply porting a smaller existing module, like the S3 module, and renaming it to `aws_s3_vm2` for this first piece.

do we expect any performance penalization running the code in this isolated environment?

aguillenv: Extracting the modules from the engine technically would make easier the testing since when any change is done in a module does not need to wait for all other modules, but this raises a couple of questions:

- what happens if a module depends on another one? If the other module change, the dependent module would be affected afterwards? would we need to introduce some versioning?
- Any change done in the engine should trigger integration tests in all modules? or again we would need to add some versioning to the engine too?

dfellis: > do we expect any performance penalization running the code in this isolated environment?

A minor amount that should not be measurable vs the CPU-bound XML parsing and TypeORM object-to-query/query-to-object transforms and the IO-bound network requests to AWS.

The last time I checked our own code was less than 1% of the CPU time and far less than 1% of the total time, depending on the operation being performed (eg, the RDS tests taking half an hour, but code we wrote taking up approximately 1 second of that time).

dfellis: > Extracting the modules from the engine technically would make easier the testing since when any change is done in a module does not need to wait for all other modules, but this raises a couple of questions:
> 
>     * what happens if a module depends on another one? If the other module change, the dependent module would be affected afterwards? would we need to introduce some versioning?

We still have the versioning in the database to support modules with version X depending on other modules with version Y in the `iasql_module` table. The majority of work would be tweaks to the dashboard to understand module versions better.

As far as how testing of modules outside of the repo would work, the SDK would have to require something like the `module.json` but with more precision on the dependency list, probably closer to Node's `package.json`? But we'd probably hew closer to a lockfile and not let modules float their dependencies, even if that causes more versioning churn on "high level" dependencies, to reduce risks with bad module interactions.

Then testing wouldn't really be any different: the module test installs the dependencies that it needs and confirms everything still works.

>     * Any change done in the engine should trigger integration tests in all modules? or again we would need to add some versioning to the engine too?

This is why I made it an error for a module to not depend on `iasql_platform`, as this version will basically become the engine version in the future, and external modules need to re-publish to function on a newer version of the engine.

In a true ecosystem this would slow down adoption of the newest version of the engine, but in this world updates to the engine itself would be shorter (and once we're actually on semver we could perhaps make it so that you can only pin to `X.Y` and not `X.Y.Z` so bugfix releases aren't blocked on module authors to re-target).

aguillenv: Ok, I got it. That sounds great to me then. I'd suggest including a bit about the versioning in the final proposal, and maybe talk a bit more on what we expect to expose in the SDK without going into implementation details.

PR #1940, opened 2023-02-07
Multistage build

aguillenv:


aguillenv:
Latest image size 
image

We can still improve. We could easily go down at least to 1GB. We need to figure out if we need all in here:
`RUN apt install postgresql-client-14 postgresql-14 postgresql-14-cron pgbouncer postgresql-server-dev-14 libcurl4-openssl-dev git make g++ locales-all -y
` since this line is the one making the image to grow by 1GB

These are all the sizes per layer in the Dockerfile:
```log
ENTRYPOINT ["/engine/docker-entrypoint.sh"]: 0B
EXPOSE map[8888/tcp:{}]: 0B
EXPOSE map[5432/tcp:{}]: 0B
ENV DB_PASSWORD=test: 0B
ARG DB_PASSWORD=test: 0B
ENV DB_USER=postgres: 0B
ARG DB_USER=postgres: 0B
ENV IASQL_ENV=local: 0B
ARG IASQL_ENV=local: 0B
COPY /engine/dist dist # buildkit: 3.74MB
COPY /engine/src/scripts src/scripts # build…: 52kB
COPY /engine/node_modules node_modules # bui…: 142MB
COPY /engine/package.json /engine/docker-ent…: 7.71kB
WORKDIR /engine: 0B
COPY /dashboard/build build # buildkit: 5.57MB
WORKDIR /dashboard: 0B
COPY /run/dist dist # buildkit: 39.5kB
COPY /run/node_modules node_modules # buildk…: 49.4MB
COPY /run/package.json package.json # buildk…: 991B
WORKDIR /dashboard/run: 0B
COPY ./src/scripts/pg_hba.conf /etc/postgres…: 5.45kB
COPY ./src/scripts/postgresql.conf /etc/post…: 36.8kB
RUN /bin/sh -c npm install -g yarn # buildkit: 8.85MB
RUN /bin/sh -c apt install nodejs -y # build…: 129MB
RUN bash -c curl -fsSL https://deb.nodesourc…: 2.79MB
RUN /bin/sh -c cd pgsql-http && make && make…: 768kB
RUN /bin/sh -c git clone --depth 1 https://g…: 134kB
RUN /bin/sh -c apt install postgresql-client…: 1.01GB
RUN /bin/sh -c apt upgrade -y # buildkit: 0B
RUN /bin/sh -c apt update # buildkit: 984kB
RUN bash -c echo 'deb http://apt.postgresql.…: 64B
RUN bash -c curl https://www.postgresql.org/…: 3.49kB
RUN /bin/sh -c apt install curl ca-certifica…: 360MB
RUN /bin/sh -c apt update # buildkit: 18MB
/bin/sh -c #(nop)  CMD ["bash"]: 0B
/bin/sh -c #(nop) ADD file:cc6a56676703ec7ab…: 124MB
```

dashboard/package.json
```diff
@@ -64,7 +63,6 @@
     "@playwright/test": "^1.24.2",
     "@trivago/prettier-plugin-sort-imports": "^3.3.0",
     "@types/amplitude-js": "^8.16.2",
-    "@types/flux": "^3.1.11",
```
depombo, line 67: `universal-cookie` is also not used

aguillenv: kinda, it is there and in the code, but we should remove it... we can create a task and do it as a follow-up. The good thing is that we don't need the node_modules from the dashboard build so that's not sizing up the final image.

dfellis (approved):

depombo (approved):

PR #1954, opened 2023-02-09
Dockerfile updates

aguillenv:
- Keep `dev` up-to-date
- Better Dockerfile formatting
- Reorder some commands for better performance
- Now wait for engine to be ready before starting the dashboard

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,7 +76,7 @@ FROM base AS engine-stage
 WORKDIR /engine
 
 ## Copy files
-COPY docker-entrypoint.sh ormconfig.js tsconfig.json ./
+COPY ormconfig.js tsconfig.json ./
 COPY src src
 
 ## Install stage dependencies
@@ -94,10 +94,6 @@ RUN yarn install --production
 # Main stage
 FROM base AS main-stage
 
-## Copy files
-COPY ./src/scripts/postgresql.conf /etc/postgresql/14/main/postgresql.conf
-COPY ./src/scripts/pg_hba.conf /etc/postgresql/14/main/pg_hba.conf
-
 ## Copy from run-stage
 WORKDIR /dashboard/run
 COPY --from=run-stage /run/package.json package.json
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -1,99 +1,131 @@
 # Base image
 FROM debian:bullseye AS base
-# Install OS Packages
+
+## Install OS Packages
 RUN apt update
-RUN apt install curl ca-certificates gnupg build-essential git jq -y
+RUN apt install curl ca-certificates gnupg build-essential git jq make g++ locales-all -y
```
depombo, line 6: what are the extra packages for? I'm worried that this puts it over the 2GB default memory threshold and we have to add documentation so people recognize the error I was getting yesterday and need to increase it.

aguillenv: These ones were already there, they are not new additions. I guess they are necessary, I just moved them from [line 14](https://github.com/iasql/iasql/pull/1954/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R14) to this `install` since these are general packages and those are `postgres` related packages

depombo: non blocker, but do you think it makes sense to add a test that makes sure the size of the image doesn't increase or posts it in the PR or this would be overkill?

depombo (approved):

Dockerfile
```diff
@@ -1,99 +1,131 @@
...
-
+## Copy from run-stage
 WORKDIR /dashboard/run
 COPY --from=run-stage /run/package.json package.json
 COPY --from=run-stage /run/node_modules node_modules
 COPY --from=run-stage /run/dist dist
 
+## Copy from dashboard-stage
 WORKDIR /dashboard
 COPY --from=dashboard-stage /dashboard/build build
```
dfellis, line 105: When you run this and connect to localhost:8888 does it work? I would expect it not to with how it is structured.

aguillenv: what do you mean? It works and all tests have passed. also, nothing has changed in this particular line from yesterday?

dfellis: Oh, because the way the run service hosts the dashboard build artifact assumes it's located in `../build` but now it needs to be in `../dashboard/build` so I honestly don't understand how that part would be working?

aguillenv: I made sure that was taken into account when structuring the directories in the container. The run service code runs under `dashboard/run` and it's `index.ts` is in `dashboard/run/dist` directory and the relative path in the run service to host the dashboard is `../../build`, so the first `..` goes back to `dashboard/run` and the second to `dashboard` where we have the `build` directory.

dfellis: Ahh, I see now, the `WORKDIR` is `/dashboard/run`. I saw all of the `COPY --from=run-stage /run/...` and thought it was also put in the top-level.

dfellis (approved):

Dockerfile
aguillenv: a test might be overkill... We might be able to update the current message showing the install all modules time info to be more general giving us relevant info on the build process like time and size?

depombo: > a test might be overkill... We might be able to update the current message showing the install all modules time info to be more general giving us relevant info on the build process like time and size?

that sounds good

PR #1958, opened 2023-02-09
Remove DbDoc schema from Navbar now that our docs have it

depombo:


dfellis (approved):
Should this also remove the dbdoc CI job?

PR #1963, opened 2023-02-10
Remove reference to 'iasql_upgrade' function from 'iasql_help'

dfellis:
I noticed while working on other things last night that `iasql_help` still referred to an `iasql_upgrade` function that hasn't existed for a few releases now.

We should also see about potentially automating the output here so it can also include the RPCs from other modules if/when installed.


mtp1376:
+1 for automating.

mtp1376 (approved):

depombo (approved):
We have a task to try to automate this https://github.com/iasql/iasql/issues/1844

PR #1964, opened 2023-02-10
Remove unneeded `iasql_{begin,commit}` from `site/`

mtp1376:


dfellis (approved):
This is good, but are these the only two unnecessary begin/commit calls left in the documentation?

depombo:
> This is good, but are these the only two unnecessary begin/commit calls left in the documentation?

Agreed, there are still a lot more in other blogs and code samples

mtp1376:
> This is good, but are these the only two unnecessary begin/commit calls left in the documentation?

Yes. Seems so. The other instances of these two functions are really needed, either by an `iasql_preview` | `iasql_rollback` or an immediate variable scan afterwards (which was coming from the cloud). 
I also checked the doc strings we had in `<module_name>/entites/*.ts` files and we seem to no longer have examples there.
But I've not yet checked the `itDocs` ones. Going for them now.

test/modules/aws-acm-import-integration.ts
```diff
@@ -88,7 +88,7 @@ describe('AwsAcm Import Integration Testing', () => {
   `),
   );
 
-  itDocs(
+  it(
```
depombo, line 91: this is not related to a transaction, why are we excluding it from the documentation?

mtp1376: Because there is the same exact query a number of blocks ahead. It just did the same thing so I kept the other one.

depombo (approved):

dfellis (approved):

PR #1972, opened 2023-02-11
WIP: Dashboard+Run merged into Nextjs project

dfellis:


Dockerfile
```diff
@@ -119,6 +87,7 @@ COPY src/scripts /engine/src/scripts
 ## Default ENVs that can be overwritten
 ARG IASQL_ENV=local
 ENV IASQL_ENV=$IASQL_ENV
+ENV NEXT_PUBLIC_IASQL_ENV=$IASQL_ENV
```
depombo, line 90: didn't we do this already in line 29? why repeat it?

Dockerfile.dev
```diff
@@ -22,46 +22,20 @@ RUN npm install -g yarn
...
-
-#####################################################################################################################################################
-
 # Dashboard
 FROM base AS dashboard-stage
 
 WORKDIR /dashboard
 
 ## Copy files
-COPY dashboard/tsconfig.json dashboard/tailwind.config.js dashboard/craco.config.js ./
```
depombo, line 49: no more craco!

dashboard/.gitattributes
```diff
@@ -1,3 +0,0 @@
-yarn.lock binary
```
depombo, line 1: nit - we might want to keep this around?

dashboard/src/components/Navbar.tsx
```diff
@@ -76,14 +79,16 @@ export default function Navbar({ userPic }: { userPic: string }) {
...
                   onClick={() => {
-                    const newIsDarkMode = !isDarkMode;
-                    localStorage.setItem('theme', newIsDarkMode ? 'dark' : 'light');
-                    if (newIsDarkMode) {
-                      document.documentElement.classList.add('dark');
-                    } else {
-                      document.documentElement.classList.remove('dark');
-                    }
-                    setDarkMode(newIsDarkMode);
+                    (() => {
```
depombo, line 82: nit - why do we need this function nesting?

dashboard/src/server-config/index.ts
```diff
@@ -2,10 +2,10 @@ import { ConfigInterface, throwError } from './config';
 
 // To prevent side-effects for other environments' error handling, the specific config
 // needs to be dynamically `require`d.
-if (!['production', 'staging', 'local', 'test', 'dev'].includes(process.env.IASQL_ENV ?? '')) {
-  throwError(`Invalid environment ${process.env.IASQL_ENV}`);
+if (!['production', 'staging', 'local', 'test', 'dev'].includes(process.env.NEXT_PUBLIC_IASQL_ENV ?? '')) {
+  throwError(`Invalid environment ${process.env.NEXT_PUBLIC_IASQL_ENV}`);
```
depombo, line 6: nit - why not just use `IASQL_ENV`?

depombo (approved):

Dockerfile
dfellis: I think Alejandro added the one in line 29 back. This is to make the environment available at actual runtime. I'm not sure if the one he added actually does anything because we don't go through a `build` step now.

dashboard/.gitattributes
dfellis: We should. It was dropped when I used the next.js docs to create a new repo and then dropped the old code and shoved the new code in its place. I will revert this.

dashboard/src/components/Navbar.tsx
dfellis: I do not fully understand why, but making it an IIFE seems to resolve the "first click glitch" on the darkmode/lightmode toggle.

dashboard/src/server-config/index.ts
dfellis: Next doesn't make any env var that is not prefixed `NEXT_PUBLIC_` available to the front-end. We technically could use `IASQL_ENV` in the `server-config`, but since those are both going to be the same thing every time, it doesn't seem appropriate to me?

Dockerfile
aguillenv: and do we need this variable at build time or not anymore?

dashboard/src/server-config/index.ts
depombo: got it

depombo (approved):

dashboard/src/components/Navbar.tsx
aguillenv: oh I never moved the `isDarkMode` to app state... So probably the `setDarkMode` was the one causing the rendering... Some tech debt to be cleaned up later

mtp1376 (approved):

dashboard/src/components/Navbar.tsx
dfellis: Hmm. I don't think so. Or at least, it's not directly related to this bug here, as this one only triggers when you click on the icon to toggle dark mode and light mode.

PR #1980, opened 2023-02-13
subquery vs join blog post

depombo:


site/blog/unlisted/subqueries.md
```diff
@@ -0,0 +1,48 @@
...
+slug: subquery-vs-join
+title: Subquery vs JOIN
+authors: [depombo]
+tags: [sql]
+---
+
+As a developer, you are probably familiar with the concept of joins in SQL databases. Joins are a powerful tool for combining data from multiple tables, but there are certain situations where subqueries may be a better choice. In this blog post, we will look at the advantages of using subqueries instead of joins in PostgreSQL, and why it may be the better option for your project.
+
+### Improved Performance
+One of the biggest advantages of using subqueries is improved performance. Subqueries can be optimized and executed more efficiently than joins, especially when dealing with large amounts of data. In PostgreSQL, subqueries can be executed in parallel, which can significantly improve performance. By breaking down the query into smaller parts, the database can use its indexing and caching mechanisms to optimize the performance of the query. In contrast, joins can be slower and more resource-intensive, especially when dealing with complex queries or large datasets.
```
dfellis, line 11: There should be a caveat here. The output of the sub-query itself will not be indexed, so the cardinality of the output needs to be much smaller than the inputs for the performance benefit to actually occur.

site/blog/unlisted/subqueries.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Better Readability
+Another advantage of using subqueries is improved readability. Subqueries can make your SQL code cleaner and easier to understand, especially for developers who are new to your project. Joins can be difficult to follow, especially when there are many tables involved or when the relationships between tables are complex. Subqueries can help simplify the code and make it easier for others to follow or reuse your logic.
+
+### Increased Flexibility
+Subqueries also offer increased flexibility compared to joins. With subqueries, you have more control over the order in which the data is retrieved, and you can easily modify the query to return the data you need. Joins are often more rigid, and making changes to the query can be more complex and time-consuming. In PostgreSQL, subqueries can be used to retrieve data from a single table and then join the results with another table, eliminating the need for complex join conditions.
+
+### Simplified Maintenance
```
dfellis, line 19: Would be good to explicitly call out composability here. You could literally have the different sub queries defined in other source files and included in the top-level query by templating. Improvements or schema changes that affect one sub query then are encapsulated away from the logic of the high-level query.

For instance you can call that out in the example below by introducing a new column on the department that affects the WHERE clause. Maybe 'Engineering' is split between 'Software' and 'Hardware Development' (so a simple `LIKE %Engineering` won't work) but they both are joined to `cost_center` table under the `R&D` category that could use a third sub-query?

depombo: Implemented!

mtp1376 (commented):
The overall post is good, easy-to-follow, and the reasons behind the titles are logical. I had two ideas for making this post even better:
1. It's a bit one-sided. The two examples you've provided are both from a situation where subqueries are better than joins. But is there a situation in which it's more efficient to use joins instead? I believe there is, and I think it's for when you want to "annotate" each row with a value from other table, and maybe applying a aggregate function afterwards. So maybe we can add that at the end, calling out that if joins are completely useless compared to subqueries, then why in earth they should even exist.
2. Are you sure that performance is increased when you're using the subqueries? 😆 I'd suggest you add `EXPLAIN` results from Postgres, or run it on a big table to compare the performance of those two queries. That'll make the ideas proposed in the post much more tangible.
3. nit: Despite the fact that you don't need the schema for `employee` and `department` tables to understand the post, but I think adding pictures showing those tables, or maybe visualizing the join vs subquery will make the post easier to follow.

yrobla (commented):
Maybe it can be backed up with some benchmarking? Is there some public data we can reference about that?

aguillenv (commented):
In general terms I like it. My comment would be to add maybe a case or be more explicit when JOINS need to be used instead of sub-queries like if you need to retrieve data from both tables, it is more straightforward with a JOIN.

site/blog/unlisted/subqueries.md
```diff
@@ -68,4 +75,12 @@ WHERE cost_center_id = (
...
+
+PostgreSQL offers a query analyzer that can help you better understand the performance characteristics of a given query. Simply prepend `EXPLAIN ANALYZE` before any given query. Historically, explicit joins have better performance, but query optimizers are getting better over time so it is best to write queries first in a logically coherent way, and then restructure if performance is a problem.
```
dfellis, line 84: This blog post would benefit from looking at two queries and picking at their `EXPLAIN` outputs, I think.

depombo: The thing is that the `EXPLAIN` output from a query is highly dependent on the data in the database

dfellis: That's true, but at least having an example of what that output can look like and how to interpret what it's telling you can be useful, even if the details differ for the users? We can just tell them that.

site/blog/unlisted/subqueries.md
```diff
@@ -0,0 +1,86 @@
...
+
+Subqueries are used to retrieve data from one table based on the results of a query on another table. They are enclosed in parentheses and can be used in different parts of a SQL statement, such as the `WHERE` clause, `HAVING` clause, and `FROM` clause. Subqueries can be correlated or non-correlated. Correlated subqueries are executed for each row of the outer query, while non-correlated subqueries are executed only once before the outer query is executed.
+
+### Know when the hammer suits the nail
+
+Joins and subqueries have their strengths and weaknesses. Joins are useful when retrieving data from multiple tables with related data, performing aggregation functions on related data, and when there are a small number of tables involved in the query. On the other hand, subqueries are useful when retrieving data from a table based on the results of a query on another table, performing complex calculations or conditional statements based on data from another table, and when there are multiple tables involved in the query and joins may become complex and difficult to manage.
```
aguillenv, line 18: this paragraph seems a bit off to me. Firsts I think we are repeating too closely `related data`. And at the end I did not get if for complex queries is better subqueries or joins?

site/blog/unlisted/subqueries.md
```diff
@@ -0,0 +1,86 @@
...
+
+```sql
+SELECT employee.name
+FROM employee
+INNER JOIN cost_center
+    ON employee.cost_center_id = cost_center.id
+WHERE cost_center.name = 'R&D';
+```
+
+The query had to change completely which seems negligible at this scale but adds up for complex queries. On the other hand, using subqueries requires fewer changes:
```
aguillenv, line 66: hmm maybe this is not the best example since the query seems the same as before but using another table?

PR #1986, opened 2023-02-14
Add an env var to toggle telemetry

### LINKED ISSUE #1955: Allow turning off telemetry locally with an environment variable

dfellis:
Resolves #1955


aguillenv (approved):

depombo (approved):

src/config/local.ts
```diff
@@ -19,9 +19,11 @@ const config: ConfigInterface = {
     test: false,
     forceLocal: false,
   },
-  telemetry: {
-    posthogKey: 'phc_xvAQWfpHug7G0SuU5P9wwAbvP9ZawgAfIEZ9FUsiarS',
-  },
 };
+if (process.env.IASQL_TELEMETRY === 'on') {
+  config.telemetry = {
+    posthogKey: 'phc_xvAQWfpHug7G0SuU5P9wwAbvP9ZawgAfIEZ9FUsiarS',
```
depombo, line 25: there is no sentry in the run/server part?

dfellis: Nope. However in Next.js land I have no idea if the setup for Sentry applies to both or not. The `server-config` section is meant for stuff that absolutely shouldn't live in the front-end.

PR #2004, opened 2023-02-16
feat: allow to specify root volume settings, or no volume

yrobla:


src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,264 @@
...
+    create: async (es: InstanceBlockDeviceMapping[], ctx: Context) => {
+      const out: InstanceBlockDeviceMapping[] = [];
+      for (const e of es) {
+        // read instance details
+        const instance: Instance = await ctx.orm.findOne(Instance, {
+          id: e.instanceId,
+        });
+
+        // if we have different regions we may be in a multi-region update situation, just skip
+        if (e.region !== instance.region) continue;
```
dfellis, line 80: Can we actually get into this situation? Or is this just a safeguard in the code?

Because I think we can prevent this sort of craziness by interlocking `region` joins where the `InstanceBlockDeviceMapping` has a `region` column that is FK joined to both the Instance and GPV entities.

src/modules/aws_ec2/rpcs/describe_ami.ts
```diff
@@ -0,0 +1,105 @@
...
+          attributes: (result ?? [])[0],
+          status: 'OK',
+          message: '',
+        },
+      ];
+    } catch (e) {
+      return [
+        {
+          attributes: '',
+          status: 'KO',
```
dfellis, line 93: That's a Knock-Out! ;)

src/modules/aws_ec2/entity/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * @public
+   * Internal AWS ID for the volume
+   */
+  @Column({
+    nullable: true,
+    comment: 'Unique identifier provided by AWS once the volume is provisioned',
+  })
+  @cloudId
+  cloudVolumeId?: string;
```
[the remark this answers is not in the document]
yrobla: i found that i actually needed to have some @cloudId on the instance_block_device_mapping entity itself. Otherwise it was not able to reconciliate between cloud and database. And what is representing a set of valid cloudId here is cloudInstanceId+cloudVolumeId+region.
It is different than the instanceId and volumeId defined here, as those values are pointing to database ids, while the cloud ones are pointing to aws identifiers

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
yrobla: it actually happened when i update an instance moving to a different region. So region is updated at instance but mapping still points to the older region, until actually things get reconciled. There is a region field on the instance_block_device_mapping but is actually independent on the instance/gpv, and i just enforce it by code. How would you enforce that from the database itself?

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,264 @@
...
+                    instanceId: instance?.id ?? undefined,
+                    instance,
+                    volumeId: volume?.id ?? undefined,
+                    volume,
+                    deviceName: newMap.DeviceName,
+                    cloudInstanceId: i.InstanceId,
+                    cloudVolumeId: newMap.Ebs.VolumeId,
+                    region,
+                    deleteOnTermination: newMap.Ebs.DeleteOnTermination ?? true,
+                  };
```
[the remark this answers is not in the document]
yrobla: indeed i can do that, i started just iterating on it because this entity was special, but it is possible to add into the mapper as well, will do

src/modules/aws_ec2/rpcs/describe_ami.ts
yrobla: i have been using this OK/KO for my error codes. Happy to change, but it's quite visual!

test/modules/aws-codedeploy-integration.ts
```diff
@@ -733,16 +733,9 @@ SELECT * FROM codedeploy_deployment_group WHERE application_id = (SELECT id FROM
     it(
       'deletes all ec2 instances',
       query(
-        `
-      BEGIN;
-        DELETE FROM general_purpose_volume
-        USING instance
-        WHERE instance.id = general_purpose_volume.attached_instance_id AND 
-          (instance.tags ->> 'name' = '${instanceTag}');
-
```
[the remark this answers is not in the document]
yrobla: no, with the new workflow the volume gets automatically removed if the DeleteOnTermination flag is set to true (by default it is like that on root devices)

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
dfellis: So moving an instance to another region becomes *very* tough when you do this because now it will fail when the FK relation on this `InstanceBlockDeviceMapping` table still has the old region.

One example of where we do it is with [`Subnet`](https://github.com/iasql/iasql/blob/main/src/modules/aws_vpc/entity/subnet.ts#L37-L52)s joining to the Availability Zone. You could follow a pattern like that including the region in the FK join of `InstanceBlockDeviceMapping` to both `Instance` and `GeneralPurposeVolume`

It would, for the most part, make mutating the `instance`'s region impossible from the FK constraints because it's volumes are in the same region. There would be a trick in a transaction to first delete the `InstanceBlockDeviceMapping` record(s), then change the `region` in GPV and Instance, then re-insert the `InstanceBlockDeviceMapping` record(s), *then* run `iasql_commit`, but that could reasonably be treated by the engine as wholly separate entities and not migrating the volume data between regions would be reasonable.

src/modules/aws_ec2/mappers/instance.ts
```diff
@@ -91,6 +103,42 @@ export class InstanceMapper extends MapperBase<Instance> {
...
+        // volume mapping
+        const vol: InstanceBlockDeviceMapping[] = [];
+        for (const map of instance.BlockDeviceMappings ?? []) {
+          if (map.DeviceName && map.Ebs?.VolumeId) {
+            const volume = await this.module.generalPurposeVolume.db.read(
+              ctx,
+              this.module.generalPurposeVolume.generateId({ volumeId: map.Ebs.VolumeId ?? '', region }),
+            );
+
+            const entry: InstanceBlockDeviceMapping = {
```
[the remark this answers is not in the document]
yrobla: so the is_root_device is set by the user itself when it generates the mapping. It means that the volume will not be created and will be just used to setup the initial mapping. In `generateBlockDeviceMapping` i take all the mappings with root_device=true and compare their device name against the mapping provided by the AMI, generating the initial mapping structure

src/modules/aws_ec2/mappers/instance.ts
```diff
@@ -290,6 +338,85 @@ export class InstanceMapper extends MapperBase<Instance> {
...
+  ) {
+    // start reading the block device mapping from the image
+    const client = await ctx.getAwsClient(instance.region);
+    const amiImage = (await this.describeImages(client.ec2client, [ami]))?.Images?.pop();
+
+    if (amiImage) {
+      const imageMapping = amiImage.BlockDeviceMappings;
+
+      // check if there is any mapped volume that doesn't exist on instance mapping, and error
+      // or if there is any volume mapped that is not set as root
```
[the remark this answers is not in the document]
yrobla: it means that i look at is_root_device_flag. If there is an instance_block_device_mapping associated to this instance but is not set as root device it will throw an error, as only the volumes flagged as root device can be used on instance creation.
The alternative if we find a case like that (an instance that is not created, but with some mapping already associated), is to ignore this volume and later we can associate as an extra one.

src/modules/aws_ec2/mappers/instance.ts
```diff
@@ -339,84 +475,92 @@ export class InstanceMapper extends MapperBase<Instance> {
...
+
           const instanceId = await this.newInstance(client.ec2client, instanceParams);
           if (!instanceId) {
             // then who?
             throw new Error('should not be possible');
           }
+
+          // read block device mapping from instance and wait for volumes in use
+          const mapping = await this.getInstanceBlockDeviceMapping(client.ec2client, instanceId);
+          for (const map of mapping ?? []) {
```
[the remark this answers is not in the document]
yrobla: mm it was a leftover for when i was initially creating the volumes. I needed to wait for them to be totally created to properly delete them later. not needed now

src/modules/aws_ec2/mappers/instance.ts
```diff
@@ -339,84 +475,92 @@ export class InstanceMapper extends MapperBase<Instance> {
...
+            if (map.DeviceName && map.Ebs?.VolumeId) {
+              await this.module.instanceBlockDeviceMapping.waitUntilInUse(
+                client.ec2client,
+                map.Ebs.VolumeId!,
+              );
+
+              // if we had a root volume associated with the instance, we need to update the volume id
```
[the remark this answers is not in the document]
yrobla: is it true, what is doing is that it waits for the volume to be created and reads the details. If the entry in the database already existed, it simply reuses that, updating the volume id, and setting root_device to false. So the ids for the mappings remain and i only need to update volume data.
I only create new entries for volumes and mappings if there was no initial volume/mapping defined.

src/modules/aws_ec2/mappers/general_purpose_volume.ts
```diff
@@ -246,10 +197,9 @@ export class GeneralPurposeVolumeMapper extends MapperBase<GeneralPurposeVolume>
     create: async (es: GeneralPurposeVolume[], ctx: Context) => {
       const out = [];
       for (const e of es) {
+        if (e.isRootDevice) continue; // cannot create root volumes, skip
+        if (e.volumeId) continue; // cannot create a volume that is already created, skip
```
dfellis, line 201: If someone mutates the definition of a volume that's already created, this will cause it to always get reverted during `commit`. Is that the desired behavior?

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,282 @@
...
+      if (id) {
+        // decompose the id
+        const { instanceId, volumeId, region } = this.idFields(id);
+        const client = (await ctx.getAwsClient(region)) as AWS;
+
+        // check if we can find the instance in database
+        const instance = await this.module.instance.db.read(
+          ctx,
+          this.module.instance.generateId({ instanceId: instanceId ?? '', region }),
+        );
```
dfellis, line 168: Since this is the `cloud.read` path, perhaps it should call `this.module.instance.cloud.read` first and if that fails it can fall back to the `db.read` path?

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,282 @@
...
+              if (!i.InstanceId) continue; // if we do not have an id we skip
+
+              // check instance block device mappings
+              const mapping = await this.module.instance.getInstanceBlockDeviceMapping(
+                client.ec2client,
+                i.InstanceId,
+              );
+
+              // check if we can find the instance in database
+              const instance = await this.module.instance.db.read(
```
dfellis, line 203: Depending on the `db.read` path in a `cloud.read` function feels like the wrong choice.

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,282 @@
...
+          }
+        }
+      } else {
+        const out: InstanceBlockDeviceMapping[] = [];
+        const enabledRegions = (await ctx.getEnabledAwsRegions()) as string[];
+        await Promise.all(
+          enabledRegions.map(async region => {
+            const client = (await ctx.getAwsClient(region)) as AWS;
+            const rawInstances = (await this.module.instance.getInstances(client.ec2client)) ?? [];
+            for (const i of rawInstances) {
```
dfellis, line 190: I do not understand the reasoning behind using the `getInstances` SDK call directly here. It seems like you could just use `this.module.instance.db.read(ctx)` and skip the conditional `continue` calls before *and* skip the `await this.module.instance.db.read(ctx, ...generateId(...)` step below.

src/modules/aws_ec2/mappers/general_purpose_volume.ts
yrobla: i noticied that behaviour, and what i am doing in case of the replace case in update is something like ```        } else {
          // Replace, but first remove volume id
          e.volumeId = undefined;
          const newVolume = await this.module.generalPurposeVolume.cloud.create(e, ctx);
          await this.module.generalPurposeVolume.cloud.delete(cloudRecord, ctx);
          out.push(newVolume);
        }
```
so i let the volume be created again

File: src/modules/aws_ec2/mappers/instance_block_device_mapping.ts (hunk 11 above)
Comment (yrobla) [reply src/modules/aws_ec2/mappers/instance_block_device_mapping.ts:168]: i did it because this mapper actually needs to have the records existing in the database, see the mapper as 
```      const res: InstanceBlockDeviceMapping = {
        instanceId: instance?.id ?? undefined,
        volumeId: volume?.id ?? undefined,
        instance,
        volume,
        deviceName: newMap.DeviceName,
        region,
        deleteOnTermination: newMap.Ebs?.DeleteOnTermination ?? true,
      };
```
so i thought that it would not make sense to query for records that are not already created

src/modules/aws_ec2/mappers/general_purpose_volume.ts
aguillenv: hmm but if the `volumeId` was edited that should not trigger a replacement but restore the record, right?

src/modules/aws_ec2/mappers/instance.ts
```diff
@@ -290,6 +331,81 @@ export class InstanceMapper extends MapperBase<Instance> {
...
+        const vol = maps?.find(item => item.deviceName === map.DeviceName);
+        if (vol) {
+          if (vol.volumeId) {
+            if (vol.volume) {
+              // map it to the ebs mapping
+              if (!vol.volume.isRootDevice) throw new Error('Error mapping volume that is not root');
+
+              let snapshotId;
+              if (vol.volume.snapshotId && instance.region === vol.region) snapshotId = vol.volume.snapshotId;
+              else snapshotId = map.Ebs?.SnapshotId;
```
aguillenv, line 370: can we refactor this using early returns to avoid the nested conditionals to make it easier to follow?

src/modules/aws_ec2/mappers/instance.ts
```diff
@@ -290,6 +331,81 @@ export class InstanceMapper extends MapperBase<Instance> {
...
+      // check if there is any mapped volume that doesn't exist on instance mapping, and error
+      // or if there is any volume mapped that is not set as root
+      for (const dev of maps ?? []) {
+        // try to find the device name on instance mapping
+        const vol = imageMapping?.find(item => item.DeviceName === dev.deviceName);
+        if (!vol) throw new Error('Error mapping volume to a device that does not exist for the AMI');
+        if (!dev.volume?.isRootDevice) throw new Error('Error mapping volume that is not root');
+      }
+      const region = instance.region;
+      for (const map of imageMapping ?? []) {
```
aguillenv, line 359: also, can we update a bit the variable naming inside this `generateBlockDeviceMapping` function? For example, I thought `map` was coming from `maps`, not `imageMapping`. It also took me some time to realize that `dev` was referring to device

src/modules/aws_ec2/mappers/instance.ts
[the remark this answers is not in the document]
aguillenv: ok got it, can we add this explanation as a comment in the code?

aguillenv (commented):
Some of my comments still need to be addressed, I think:
https://github.com/iasql/iasql/pull/2004/files#r1126156526
https://github.com/iasql/iasql/pull/2004/files#r1126160231
https://github.com/iasql/iasql/pull/2004/files#r1126166102

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,257 @@
...
+      !Object.keys(fields).every(fk => requiredFields.includes(fk))
+    ) {
+      throw new Error(`Id generation error. Valid fields to generate id are: ${requiredFields.join(', ')}`);
+    }
+    return `${fields.instanceId}|${fields.volumeId}|${fields.region}`;
+  };
+  entityId = (e: InstanceBlockDeviceMapping) => {
+    return this.module.instanceBlockDeviceMapping.generateId({
+      instanceId: e.instance?.instanceId ?? '',
+      volumeId: e.volume?.volumeId ?? '',
```
dfellis, line 26: Why fall back on these to empty string? Are there legitimate `InstanceBlockDeviceMapping` records with this being the case? (If so, how do you distinguish between multiple of them, if, say, a new instance is defined with 3 new volume definitions? There would be 3 of these records that all have the same ID so the engine would think there's only one of them to operate on.)

test/modules/aws-ec2-gpv-integration.ts
```diff
@@ -119,6 +139,158 @@ describe('EC2 General Purpose Volume Integration Testing', () => {
 
   it('starts a transaction', begin());
 
+  itDocs('creates an instance wiht specific size', (done: (arg0: any) => any) => {
```
dfellis, line 142:
```suggestion
  itDocs('creates an instance with a specific size', (done: (arg0: any) => any) => {
```

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
yrobla: actually the volume is optional here because it actually can be the case in the initial mapping of cleaning volumes with NoDevice. But it will only exist in the database and in the initial mapping definition, not in the cloud

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,259 @@
...
+    return `${fields.instanceId}|${fields.deviceName}|${fields.volumeId}|${fields.region}`;
+  };
+  entityId = (e: InstanceBlockDeviceMapping) => {
+    return this.module.instanceBlockDeviceMapping.generateId({
+      instanceId: e.instance?.instanceId ?? '',
+      deviceName: e.deviceName,
+      volumeId: e.volume?.volumeId ?? '',
+      region: e.region,
+    });
+  };
```
aguillenv, line 30: having this here should be overriding your recently added `@cloudId` decorators for this entity. Ideally, we should just keep one of them to avoid confusions.

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,259 @@
...
+
+        // if instance is not created we are in the first step, no need to create anything
+        if (!e.instance.instanceId) continue;
+        const client = (await ctx.getAwsClient(e.region)) as AWS;
+
+        // read volume details
+        if (!e.volumeId) throw new Error('Cannot attach empty volumes to an instance already created');
+        if (e.volume) {
+          if (e.volume.region !== e.instance.region)
+            throw new Error('Cannot create a mapping between different regions');
```
aguillenv, line 127: Is possible for this to happen? Isn't this handled by the database with the FKs?

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,259 @@
...
+          } else continue;
+          out.push(e);
+        }
+      }
+      return out;
+    },
+    read: async (ctx: Context, id?: string) => {
+      if (id) {
+        // decompose the id
+        const { instanceId, volumeId, region } = this.idFields(id);
```
aguillenv, line 141: are we sure this is returning the right values? I'm asking since we have the mix between overrides and `@cloudId`s

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
```diff
@@ -0,0 +1,259 @@
...
+              e.instance.instanceId,
+              mappings,
+            );
+            if (result) out.push(e);
+          }
+        } else {
+          // the only case for update is to change the device name, we will need to delete and recreate
+          await this.module.instanceBlockDeviceMapping.cloud.delete(e, ctx);
+          const newMap = await this.module.instanceBlockDeviceMapping.cloud.create(e, ctx);
+          if (!newMap) continue;
```
aguillenv, line 231: since we have this case here we should also have the informative fn `updateOrReplace` to show to the users correctly when it is an update and when a replace

src/modules/aws_ec2/mappers/instance.ts
```diff
@@ -91,6 +102,35 @@ export class InstanceMapper extends MapperBase<Instance> {
       ));
     out.hibernationEnabled = instance.HibernationOptions?.Configured ?? false;
     out.region = region;
+
+    // check if we can find the instance in database
+    if (instance.InstanceId) {
+      const instanceObj = await this.module.instance.db.read(
```
aguillenv, line 108: this could be 
```suggestion
      const instanceObj = await this.db.read(
```
right?

src/modules/aws_ec2/mappers/instance.ts
```diff
@@ -91,6 +102,35 @@ export class InstanceMapper extends MapperBase<Instance> {
...
+        ctx,
+        this.module.instance.generateId({ instanceId: instance.InstanceId, region }),
```
aguillenv, line 110: same here, I think
```suggestion
        this.generateId({ instanceId: instance.InstanceId, region }),
```

src/modules/aws_ec2/mappers/instance.ts
```diff
@@ -290,6 +330,82 @@ export class InstanceMapper extends MapperBase<Instance> {
     _res => undefined,
   );
 
+  // given an instance reads the mapping from the associate AMI , match
```
aguillenv, line 333:
```suggestion
  // given an instance reads the mapping from the associate AMI, match
```

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
yrobla: i need to have it here, as the fields for id are complex.. i actually tried to get rid of @cloudId but the tooling doesn't let me do it, it complains that if i want to map an entity, i need to have at least one cloudId.

src/modules/aws_ec2/entity/instance_block_device_mapping.ts
```diff
@@ -27,13 +27,16 @@ export class InstanceBlockDeviceMapping {
    * Device name to associate this volume to the instance
    */
   @Column()
+  @cloudId
```
dfellis, line 30: Why add these `@cloudId` decorators if you are overriding the `entityId` and `generateId` methods?

yrobla: i answered to Alejandro previously as well, if i do not add any @cloudId the tooling complains... so it didn't complain before because there was a leftover for that in `region` field. If i remove all @cloudId code fails. So i prefered to put the reasonable ones. But please suggest any other approach that would work, i would like to get rid of those cloudId as well.

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
aguillenv: You can get rid of the `@cloudId` decorators if you define the `entityId` function yourself, you have the `CodedeployDeploymentGroup` case you already worked on, for example. The only thing is that if you remove the `@cloudId`s you have to write yourself the `entityId`, `idFields` and `generateId` methods in the mapper

yrobla: i added but still getting errors

src/modules/aws_ec2/mappers/instance.ts
```diff
@@ -105,7 +105,7 @@ export class InstanceMapper extends MapperBase<Instance> {
 
     // check if we can find the instance in database
     if (instance.InstanceId) {
-      const instanceObj = await this.module.instance.db.read(
+      const instanceObj = await this.db.read(
```
dfellis, line 108: Calling it directly like this will bypass the memoization cache, which can negatively impact performance, but will guarantee it's getting the latest value from the DB. Not sure if that's what you want to do here.

src/modules/aws_ec2/mappers/instance_block_device_mapping.ts
yrobla: indeed it was an older code leftover

src/modules/aws_ec2/mappers/instance.ts
dfellis: Nevermind. We fixed this in the past and I forgot about it.

test/modules/aws-ec2-gpv-integration.ts
```diff
@@ -119,6 +139,216 @@ describe('EC2 General Purpose Volume Integration Testing', () => {
...
+  it('starts a transaction', begin());
+
+  itDocs('creates an instance with specific size and an extra volume', (done: (arg0: any) => any) => {
+    query(
+      `
+      BEGIN;
+        INSERT INTO general_purpose_volume (volume_type, size, availability_zone, tags, is_root_device)
+        VALUES ('gp2', 15, '${availabilityZone2}', '{"Name": "${gp2VolumeName}-test-1"}', TRUE);
+        INSERT INTO general_purpose_volume (volume_type, size, availability_zone, tags, is_root_device)
+        VALUES ('gp2', 20, '${availabilityZone2}', '{"Name": "${gp2VolumeName}-test-2"}', TRUE);
```
dfellis, line 233: This can be

```suggestion
        INSERT INTO general_purpose_volume (volume_type, size, availability_zone, tags, is_root_device)
        VALUES ('gp2', 15, '${availabilityZone2}', '{"Name": "${gp2VolumeName}-test-1"}', TRUE),
        VALUES ('gp2', 20, '${availabilityZone2}', '{"Name": "${gp2VolumeName}-test-2"}', TRUE);
```

dfellis (commented):
Only very minor comments. I think we're in the home stretch! :)

dfellis (approved):

PR #2005, opened 2023-02-16
Add IAM Policy Normalizer

mtp1376:


dfellis (approved):

test/common/canonical-iam-policy.ts
```diff
@@ -0,0 +1,228 @@
+import { normalizePolicy } from '../../src/services/canonical-iam-policy';
```
depombo, line 1: common tests should be running and they are not, we should fix the CI job and make sure those test pass before landing

Issue #1989, opened 2023-02-14
History of query results / tabs for each


aguillenv:
Original issue: https://github.com/iasql/dashboard/issues/207

- [x] [Add tabs to the editor and keep track of the result for each one](https://github.com/iasql/iasql/pull/1990)
- [x] [When clicking on table names (or a button close to it) open a new tab with the generated select query](https://github.com/iasql/iasql/pull/2013)
- [x] [Update install/uninstall behaviour to open a new tab](https://github.com/iasql/iasql/pull/2012)
~- [ ] Keep the history of queries+results obtained per tab per session~

aguillenv:
@depombo:
> This allows showing the result of a query made of multiple queries and visualizing the tables by clicking on it.

@aguillenv:

>So we have small tasks here:
>
> [ ] Add tabs to the editor and keep track of the result for each one
> [ ] When clicking on table names (or a button close to it) open a new tab with the generated select query
> [ ] Update install/uninstall behaviour to open a new tab
> [ ] Keep the history of queries+results obtained per tab per session

@dfellis:

> * Shouldn't we open a new issue in the monorepo instead of continuing to use this one?
> * @aguillenv how is your fourth checklist item different from your first one?

@aguillenv:
> > * Shouldn't we open a new issue in the monorepo instead of continuing to use this one?
> 
> Oh, I thought it was already in the monorepo. I'll transfer the issue.
> 
> > * @aguillenv how is your fourth checklist item different from your first one?
> 
> > > Add tabs to the editor and keep track of the result for each one
> 
> This one will just to introduce tabs to the editor. Each tab will show one result table of what you executed in each one.
> 
> > > Keep the history of queries+results obtained per tab per session
> 
> This one will keep the history per tab of all queries executed in it. Maybe with a UI like we talked about with some collapsible components.
> 
> With just tabs we do not have "history", right?

PR #2013, opened 2023-02-17
Select tables with rows

aguillenv:
Part of #1989 

image

image

### DIFF
--- a/dashboard/src/components/AppProvider.tsx
+++ b/dashboard/src/components/AppProvider.tsx
@@ -85,6 +85,7 @@ interface AppState {
     queryRes?: any | null;
     closable?: boolean;
   }[];
+  forceRun: boolean;
 }
 
 interface AppStore extends AppState {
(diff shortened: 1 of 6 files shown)


dfellis:
Does this auto-run the query on click of the :mag: ?

aguillenv:
> Does this auto-run the query on click of the 🔍 ?

Yes!

dfellis (approved):

depombo:
nit - using eye instead of a magnifying glass icon? not sure though. I don't love either

depombo (approved):

dfellis:
> nit - using eye instead of a magnifying glass icon? not sure though. I don't love either

I think the eyes might be a bit creepy all in a row?

yrobla (approved):

aguillenv:
leaving a couple of options for a follow-up:

- Search icon:
image

- Eye outline
image

- Eye solid
image

- table icon 
image

- document search icon
image

- External icon
image

aguillenv:
> and the icon only shows up if the table has a nonzero number of rows?

I forgot to add the conditional but it does now :)

depombo:
> leaving a couple of options for a follow-up:
> 
> * Search icon:
>   image
> * Eye outline
>   image
> * Eye solid
>   image
> * table icon
>   image
> * document search icon
>   image
> * External link / open in new tab icon
>   image

I think I like the table icon best, thoughts @aguillenv @dfellis ?

PR #2014, opened 2023-02-17
Use pg-format on 'iasql_uninstall' calls

dfellis:


depombo (approved):

aguillenv (approved):

yrobla (approved):

mtp1376 (approved):
I'd suggest you also add this mitigation to `iasql_install` RPC. It doesn't have the same impact, but let's do it just to increase confidence.

mtp1376 (approved):

PR #2017, opened 2023-02-17
fix telemetry when auth is off

depombo:


docker-entrypoint.sh
```diff
@@ -71,6 +71,6 @@ wait-for-url() {
 }
 
 # Dashboard service
-(cd /dashboard; wait-for-url localhost:8088/v1/version && yarn start:docker) &
+(cd /dashboard; wait-for-url localhost:8088/v1/version && NEXT_PUBLIC_UID=$(cat /proc/sys/kernel/random/uuid) yarn start:docker) &
```
dfellis, line 74: Oh wow, I had no idea The Linux kernel itself sprouted a UUID generator. That's cool :)

dashboard/package.json
```diff
@@ -43,7 +43,6 @@
     "typescript": "4.9.5",
     "unique-names-generator": "^4.7.1",
     "universal-cookie": "^4.0.4",
-    "uuid": "^9.0.0",
```
dfellis, line 46: Did you re-run `yarn` after this? I'm surprised the `yarn.lock` file was not touched.

depombo: I did and it didn't regenerate it. I was also surprised

docker-entrypoint.sh
depombo: Yes, I also learned the today!

dfellis (approved):

aguillenv (approved):

PR #2028, opened 2023-02-20
Keep the three package.json version numbers in sync

dfellis:
This was noticed in the last release. A minor glitch which should be fixed by these two script changes

### DIFF
--- a/.develop.sh
+++ b/.develop.sh
@@ -27,7 +27,6 @@ yarn format
 # Same for the dashboard
 cd dashboard
 yarn
-yarn format
 cd ..
 
 # Same for the site
--- a/.develop.sh
+++ b/.develop.sh
@@ -15,6 +15,8 @@ CONFIGVERSION="$(cat src/config/index.ts | sed "s/^  version:.*/version: '${VERS
 
 # Update the package metadata with the specified version
 PACKAGEJSON="$(jq ".version = \"${VERSION}-beta\"" package.json)" && echo "${PACKAGEJSON}" > package.json
+PACKAGEJSON="$(jq ".version = \"${VERSION}-beta\"" dashboard/package.json)" && echo "${PACKAGEJSON}" > dashboard/package.json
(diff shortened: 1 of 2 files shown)


.develop.sh
```diff
@@ -15,14 +15,27 @@ CONFIGVERSION="$(cat src/config/index.ts | sed "s/^  version:.*/version: '${VERS
...
 # Make sure the lockfiles are updated, too
 yarn
 
 # Make sure it's all formatted the way we want it
 yarn format
 
+# Same for the dashboard
+cd dashboard
+yarn
+yarn format
```
aguillenv, line 30: nit: do we need this?

.version.sh
```diff
@@ -17,15 +17,28 @@ CONFIGVERSION="$(cat src/config/index.ts | sed "s/^  version:.*/version: '${VERS
...
 # Make sure the lockfiles are updated, too
 yarn
 
 # Make sure it's all formatted the way we want it
 yarn format
 
+# Same for the dashboard
+cd dashboard
+yarn
+yarn format
```
aguillenv, line 32: same nit

aguillenv (approved):

depombo (approved):

.develop.sh
dfellis: Hmm... Does our formatter also format the `package.json` file? I know we've configured formatting to touch `.sql` files, for instance. I put that in there just to be safe that we don't accidentally auto-commit files that break our tests.

aguillenv: the `yarn format` inside the `dashboard` dir calls `npx prettier --config .prettierrc 'src/**/*.{ts,tsx}' --write` so should not be a problem

dfellis: Then I can remove it if you want. We'll just need to be careful to re-add it if we ever auto-format the `package.json` files.

dfellis, line 30:
```suggestion
```

.version.sh
dfellis, line 32:
```suggestion
```

PR #2029, opened 2023-02-20
Auto-generate `iasql_help` for RPCs

mtp1376:
image


src/modules/iasql_functions/rpcs/iasql_help.ts
```diff
@@ -0,0 +1,65 @@
...
+    sample_usage: 'varchar',
+  } as const;
+  call = async (
+    dbId: string,
+    _dbUser: string,
+    _ctx: Context,
+  ): Promise<RpcResponseObject<typeof this.outputTable>[]> => {
+    const returnList = [];
+
+    for (const [functionName, documentation] of Object.entries(this.module.provides.functions)) {
```
dfellis, line 47: Should this only be for the `iasql_functions` module, or for all modules currently installed by the user?

I think the latter would be more interesting, though we should probably:

1. Have a `module` column in the output to indicate the source module the function comes from.
2. Perhaps put that behavior behind an optional flag that has to be `'all'` or something like that if the output seems too overwhelming?

src/modules/iasql_functions/rpcs/iasql_help.ts
```diff
@@ -0,0 +1,65 @@
...
+      if (!!documentation.description && !!documentation.sample_usage)
```
dfellis, line 48: Won't this conditional skip anything that doesn't have a description or sample usage? It may be worthwhile to have an `else` branch that does a `logger.debug` call so we can potentially notice this during tests.

src/modules/iasql_functions/sql/after_install.sql
```diff
@@ -38,6 +38,9 @@ begin
 end;
 $$;
 
+COMMENT
+  ON FUNCTION iasql_modules_installed () IS '{"description": "Returns list of currently installed modules", "sample_usage": "SELECT * FROM iasql_modules_installed()"}';
```
dfellis, line 42: Clever!

src/modules/interfaces.ts
```diff
@@ -782,6 +797,62 @@ export class ModuleBase {
...
+  getFunctionsFromSql(sql: string): {
+    [key: string]: { description?: string; sample_usage?: string; signature?: string };
+  } {
+    const returnValue: {
+      [key: string]: { description?: string; sample_usage?: string; signature?: string };
+    } = {};
+    try {
+      const sqlParsed = parse(sql);
+      const functionsAndParams = sqlParsed
+        .filter((s: any) => _.has(s, ['RawStmt', 'stmt', 'CreateFunctionStmt']))
```
dfellis, line 809: Nit. `_.has` is most likely slower than:

```suggestion
        .filter((s: any) => !!s?.RawStmt?.stmt?.CreateFunctionStmt))
```

and while it's full of symbols requires a smaller body of knowledge for the reader (because it's just standard modern Javascript) so I think it'll take less time to understand what it's doing.

src/services/common.ts
```diff
@@ -2,9 +2,9 @@ export function isString(obj: unknown): obj is string {
   return typeof obj === 'string';
 }
 
-export const safeParse = (json: string) => {
+export const safeParse = (json: string | undefined) => {
```
dfellis, line 5: Nit. This should be equivalent:

```suggestion
export const safeParse = (json?: string) => {
```

src/services/db-manager.ts
```diff
@@ -12,24 +12,19 @@ export async function migrate(conn: Connection) {
   const version = iasqlPlatform.version;
   const qr = conn.createQueryRunner();
   await qr.connect();
-  if (iasqlPlatform.migrations?.beforeInstall) {
-    await iasqlPlatform.migrations.beforeInstall(qr);
-  }
+  await iasqlPlatform.migrations.beforeInstall?.(qr);
```
dfellis, line 15: We don't really need all of the safeguarding we had before when there were multiple versions of the module in the same codebase. Let's further reduce this to:

```suggestion
  await iasqlPlatform.migrations.beforeInstall(qr);
```

or delete this line if that doesn't exist.

src/services/db-manager.ts
```diff
@@ -12,24 +12,19 @@ export async function migrate(conn: Connection) {
...
   await iasqlPlatform.migrations.install(qr);
-  if (iasqlPlatform.migrations.afterInstall) {
-    await iasqlPlatform.migrations.afterInstall(qr);
-  }
-  await qr.query(`INSERT INTO iasql_module VALUES ('iasql_platform@${version}')`);
-  if (Modules.iasqlFunctions?.migrations?.beforeInstall) {
-    await Modules.iasqlFunctions?.migrations?.beforeInstall(qr);
-  }
+  await iasqlPlatform.migrations.afterInstall?.(qr);
```
dfellis, line 17: Same

```suggestion
  await iasqlPlatform.migrations.afterInstall(qr);
```

src/services/db-manager.ts
```diff
@@ -12,24 +12,19 @@ export async function migrate(conn: Connection) {
...
+  await qr.query(`INSERT INTO iasql_module
+                  VALUES ('iasql_platform@${version}')`);
+  await Modules.iasqlFunctions?.migrations?.beforeInstall?.(qr);
```
dfellis, line 20:
```suggestion
  await Modules.iasqlFunctions.migrations.beforeInstall(qr);
```

src/services/db-manager.ts
```diff
@@ -12,24 +12,19 @@ export async function migrate(conn: Connection) {
...
   await Modules.iasqlFunctions?.migrations?.install(qr);
```
dfellis, line 21:
```suggestion
  await Modules.iasqlFunctions.migrations.install(qr);
```

src/services/db-manager.ts
```diff
@@ -12,24 +12,19 @@ export async function migrate(conn: Connection) {
...
-  if (Modules.iasqlFunctions?.migrations?.afterInstall) {
-    await Modules.iasqlFunctions?.migrations?.afterInstall(qr);
-  }
-  await qr.query(`INSERT INTO iasql_module VALUES ('iasql_functions@${version}')`);
+  await Modules.iasqlFunctions?.migrations?.afterInstall?.(qr);
```
dfellis, line 22:
```suggestion
  await Modules.iasqlFunctions.migrations.afterInstall(qr);
```

src/services/typeorm.ts
```diff
@@ -83,12 +83,6 @@ export class TypeormWrapper {
     return typeorm;
   }
 
-  getMasterConnection(): Pool | undefined {
-    if (this.connection.driver instanceof PostgresDriver) {
-      return this.connection.driver.master as Pool;
-    }
-  }
-
```
dfellis, line 91: Nice catch. Does your editor tell you about unused methods?

src/modules/iasql_functions/rpcs/iasql_help.ts
aguillenv: well, we were only showing in `iasql_help()` the functions provided by us in this module. I think that would be a good idea but maybe seems like a follow-up that can be done using this work as base?

src/modules/interfaces.ts
```diff
@@ -496,6 +500,17 @@ export class RpcBase implements RpcInterface {
...
+      const documentationObject = {
+        description: this.helpDescription,
+        sample_usage: this.helpSampleUsage,
+      };
+      afterInstallSql += format(
+        'COMMENT ON FUNCTION %I IS %L;',
+        snakeCase(key),
+        JSON.stringify(documentationObject),
+      );
+    }
```
aguillenv, line 513: oh nice so we would have comments now on every function

src/modules/interfaces.ts
aguillenv: +1

src/modules/interfaces.ts
```diff
@@ -782,6 +797,62 @@ export class ModuleBase {
...
+            ...returnValue[v[0]],
+            description: v[1].description,
+            sample_usage: v[1].sample_usage,
+          };
+        });
+      return returnValue;
+    } catch (_) {
+      /** Do nothing */
+    }
+    return {};
```
aguillenv, line 853: well this could be
```suggestion
    return returnValue;
```
right? and the return in line 849 should not be necessary?

src/modules/iasql_functions/rpcs/iasql_help.ts
mtp1376: This is how will it work without an input:
image

Yet, the user can narrow down the result like normal queries:
```sql
SELECT * FROM iasql_help() WHERE module = 'iasql_functions';
```
or
```sql
SELECT * FROM iasql_help() WHERE name LIKE 'iasql_%`;
```

src/modules/iasql_functions/rpcs/iasql_help.ts
mtp1376: Changes in the new commit will take this into account.

src/modules/iasql_functions/sql/after_install.sql
mtp1376: It was propsed by @yrobla . I also think it's a cool idae.

src/services/db-manager.ts
mtp1376: We have
```ts
const beforeInstallSql = this.readSqlDir('before_install') ?? '';
```
in our code, which makes that variable possibly undefined. For that `migrations` object, we have this signature:
```ts
migrations: {
  beforeInstall?: (q: QueryRunner) => Promise<void>;
...
}
```
which makes sense.

src/services/typeorm.ts
mtp1376: Yes it does. But I also do manual checks since it sometimes misfires.

src/modules/interfaces.ts
mtp1376: In the current version it either returns a correctly created object or an empty one. In that case it may return partially-filed objects. Which one do you think we should do?

src/modules/iasql_functions/rpcs/iasql_help.ts
dfellis: Let's leave it as-is for this PR, this is already *much* nicer.

src/modules/iasql_functions/rpcs/iasql_get_sql_since.ts
```diff
@@ -96,7 +98,7 @@ async function getChangeLogs(limitDate: string, orm: TypeormWrapper): Promise<Ia
 }
 
 /** @internal */
-async function getInstalledModules(orm: TypeormWrapper): Promise<ModuleInterface[]> {
+export async function getInstalledModules(orm: TypeormWrapper): Promise<ModuleInterface[]> {
```
dfellis, line 101: So we tend to have shared logic between these RPC functions in the `src/modules/iasql_functions/iasql.ts` file. Perhaps this should be moved there and imported into both RPCs?

src/modules/iasql_functions/index.ts
```diff
@@ -64,3 +64,4 @@ export class IasqlFunctions extends ModuleBase {
  * ```
  */
 export const iasqlFunctions = new IasqlFunctions();
+export { getInstalledModules } from './iasql';
```
dfellis, line 67: I don't get this one. Why not follow the pattern in [other RPCs and just import `../iasql` directly](https://github.com/iasql/iasql/blob/main/src/modules/iasql_functions/rpcs/iasql_install.ts#L12)

mtp1376: Yes it seemed a bit awkward to me as well. I'll move to that pattern.

dfellis (approved):

src/modules/interfaces.ts
aguillenv: let's keep it as is

aguillenv (approved):

Issue #2023, opened 2023-02-20
Reducing the size of the docker image further


depombo:
- Move next.js to production build

dfellis:
Ideas to reduce it:

1. Run the Next.js build logic for each environment (and twice for `local` with and without telemetry enabled) and copy them into different directories running only the production build that matches the selected environment. Can reduce the amount of modules in `node_modules` and `node_modules` can be symlinked between all of them, which should make space usage a bit lower and speed up time-to-start for the dashboard at the cost of build complexity.
2. Figure out which `locale` in the `locales-all` meta-package we *actually* need for Postgres to have utf8 support and only install that one, which should save ~200MB
3. Find Debian packages that are installed but not actually used by any of our code and remove them, even if normally installed (like the package for `top`, though that one *is* already removed) and trim some of the ~300MB of OS deps.
4. Reduce `node_module` duplication by having the engine and dashboard use the same `node_module` pool, potentially reducing up to ~200MB
5. Explore using a base OS that's designed to be slimmer like Alpine Linux
6. Rewrite all of the bash scripts into Node.js scripts so all of the bash utilities we use like `jq`, `curl`, etc can be dropped (possibly even dropping `bash` itself). This one feels like trying to squeeze blood from a stone, though.

aguillenv:
> 1. Run the Next.js build logic for each environment (and twice for `local` with and without telemetry enabled) and copy them into different directories running only the production build that matches the selected environment. Can reduce the amount of modules in `node_modules` and `node_modules` can be symlinked between all of them, which should make space usage a bit lower and speed up time-to-start for the dashboard at the cost of build complexity.

Thinking about this point there's an issue with the UUID since it needs to be generated on the flight per run... I had another idea, which is making the build of the nextjs project in the same docker entry point script in parallel with the postgres init. In my machine, the nextjs build seems to be done in ~20 to 30s so I think both are compatible and we would not need to create unnecessary bundles per environment. This would not reduce the docker container size though but might fix the auth0 app and production experience with the dashboard.

aguillenv:
I have managed to pass env vars after building the dashboard. The env vars are available on the server side always, so on init of the dashboard we request the values and render accordingly.

PR #2043, opened 2023-02-22
Run Next.js build

### LINKED ISSUE #1948: Normalize IASQL_ENV across projects

aguillenv:
- Reduce container size from 1.6GB to 1.2 GB.
- Enable runtime config (uid and telemetry) after running the Next.js build logic just once.
- ~The `NEXT_PUBLIC_IASQL_ENV` env var still needs to be passed correctly on `build` time.~
- Refactored so `IASQL_ENV` can be defined at runtime and the config is picked correctly.
- Getting rid of the distinction between `NEXT_PUBLIC_IASQL_*` and `IASQL_*` env vars, always the latter.
- Removed a couple of unnecessary packages from the dashboard.

Part of #2023 
Close #1948


Dockerfile
```diff
@@ -55,6 +55,12 @@ COPY dashboard/src src
 COPY dashboard/package.json dashboard/yarn.lock ./
 RUN yarn install --production
 
+ARG IASQL_ENV=local
+ENV IASQL_ENV=$IASQL_ENV
+ENV NEXT_PUBLIC_IASQL_ENV=$IASQL_ENV
+
+RUN yarn build
```
depombo, line 62: so then `IASQL_ENV` is now picked up when building the container and not running it?

depombo: to preserve the behavior of https://github.com/iasql/iasql/pull/2039 the `IASQL_ENV` needs to move to the build command or it needs to be in both now?

dashboard/src/pages/index.tsx
```diff
@@ -28,6 +35,19 @@ export default function App() {
...
+    if (telemetry !== undefined && telemetry === 'on') {
+      Sentry.init();
+      Posthog.init();
+      if (!config.auth && uid !== undefined && uid === 'nouid') {
+        throwError('No uid found');
+      } else if (!config.auth && uid !== undefined && uid !== 'nouid') {
+        Sentry.identify(uid);
+        Posthog.identify(uid);
+      }
+    }
```
depombo, line 48: nit probably better to do early return/throwing instead of conditional nesting?

dashboard/src/pages/api/runtime-config.ts
```diff
@@ -0,0 +1,7 @@
+import { NextApiResponse } from 'next';
+
+const handler = ({} = {}, res: NextApiResponse) => {
+  res.status(200).json({ uid: process.env.IASQL_UID ?? 'nouid', telemetry: process.env.IASQL_TELEMETRY });
```
depombo, line 4: nit - why not just check the config.auth here and throwError if nouid?

Dockerfile
depombo: I am seeing the `to_upgrade` database back in posthog so there needs to be a change, hopefully it is only at build. In that case we should also throw if `IASQL_ENV` is passed at runtime so it doesn't get ignored or misused?

Screenshot 2023-02-22 at 7 21 43 AM
ubusercontent.com/2381303/220618385-2d994bae-42b9-4c09-b89d-1794df393aab.png">

aguillenv: Hmm yes, it would need to be sent on build time, the error might make sense, let me review that

aguillenv: > I am seeing the `to_upgrade` database back in posthog so there needs to be a change, hopefully it is only at build. In that case we should also throw if `IASQL_ENV` is passed at runtime so it doesn't get ignored or misused?
> 
> Screenshot 2023-02-22 at 7 21 43 AM

 This happened because I did not rebase. Now should not happen but you brought up a great point regarding the `IASQL_ENV`... I think we should just define a standard (maybe as you said just defining it at building time) to avoid misuse or confusion. 

Let me think a solution for this and will propose it before implementing it in another PR

depombo: > - The NEXT_PUBLIC_IASQL_ENV env var still needs to be passed correctly on build time.

hmm if the above is true even if you rebase the upgrade tests will send metrics to posthog unless there is a change to the `docker-build-vol` command

aguillenv: Ok, so I think the best solution is following the pattern I've added to this PR, request the `IASQL_ENV` to the API since it can know the env var value at run time and based on that pick the config object to use. It would not be exactly as is in different failed, probably will be a config object with the different envs and we will pick the one we need before rendering.

aguillenv: it will always be taken into account

depombo: I'm not sure what that means given there is build and runtime environment variable passing which opens several scenarios. Does this mean:
1) If `IASQL_ENV` is passed at runtime it will overwrite the one passed at build time?
2) If no `IASQL_ENV` is passed at runtime it will use the one at build time and not throw an error?
3) If no `IASQL_ENV` is passed at build time there will be an error thrown?

aguillenv: There are facts to have the context:
- Docker defines [environment variables precedence](https://docs.docker.com/compose/environment-variables/envvars-precedence/) to override when needed. Env vars defined at runtime have more precedence than the ones defined at build time so they will be overwriten.

- We are not able to know when one env var has been defined at build or run time. We cannot know if we are overwriting a variable or not.

So answering your scenarios:

> 1. If IASQL_ENV is passed at runtime it will overwrite the one passed at build time?
Yes, always.

> 2. If no IASQL_ENV is passed at runtime it will use the one at build time and not throw an error?
Yes

> 3. If no IASQL_ENV is passed at build time there will be an error thrown?
Not sure if this is possible to do.


The solution proposed here is:
- Always define `IASQL_ENV=local` in the `Dockerfile`. This will have no implications in the build process of any of the projects and will provide the environment local if you run the container and do not pass any `IASQL_ENV` to `docker run` command.
- Define `IASQL_ENV` when running `docker run` to override the environment variable and use the desired value.

Dockerfile
```diff
@@ -62,14 +65,14 @@ FROM build AS engine-stage
...
-## Copy files
-COPY .yarnrc ormconfig.js tsconfig.json ./
-COPY src src
-
 ## Install stage dependencies
 COPY package.json yarn.lock ./
 RUN yarn install
 
+## Copy files
+COPY .yarnrc ormconfig.js tsconfig.json ./
```
dfellis, line 73: The `.yarnrc` line, at least, needs to be moved back up where it was, or we will get build issues in the Docker ARM64 cross-compile because of package download timeouts.

Dockerfile
```diff
@@ -86,36 +89,34 @@ WORKDIR /
...
+COPY ./src/scripts/postgresql.conf /etc/postgresql/14/main/postgresql.conf
+COPY ./src/scripts/pg_hba.conf /etc/postgresql/14/main/pg_hba.conf
+COPY docker-entrypoint.sh /engine/docker-entrypoint.sh
+COPY src/scripts /engine/src/scripts
 
 ## Copy from engine-stage
 WORKDIR /engine
-COPY --from=engine-stage /engine/package.json ./
 COPY --from=engine-stage /engine/node_modules node_modules
+COPY --from=engine-stage /engine/package.json ./
```
dfellis, line 101: Why swap the order on these two lines?

dashboard/package.json
```diff
@@ -7,7 +7,7 @@
     "dev": "next dev",
     "build": "next build",
     "start": "next start",
-    "start:docker": "next dev -p 9876",
+    "start:docker": "node server.js",
```
dfellis, line 10: Where does this `server.js` file come from? I don't see it copied into the Dockerfile nor do I see anything to generate it?

Dockerfile
```diff
@@ -86,36 +89,34 @@ WORKDIR /
...
 ENV IASQL_ENV=$IASQL_ENV
-ENV NEXT_PUBLIC_IASQL_ENV=$IASQL_ENV
 ARG IASQL_TELEMETRY=on
 ENV IASQL_TELEMETRY=$IASQL_TELEMETRY
-ENV NEXT_PUBLIC_IASQL_TELEMETRY=$IASQL_TELEMETRY
 ARG DB_USER=postgres
 ENV DB_USER=$DB_USER
 ARG DB_PASSWORD=test
 ENV DB_PASSWORD=$DB_PASSWORD
+ENV PORT=9876
```
dfellis, line 119: Why can't the `start:docker` in the dashboard package.json define this?

dashboard/src/components/providers/ConfigProvider.tsx
```diff
@@ -0,0 +1,50 @@
...
+};
+
+const AppConfigProvider = ({ children }: { children: any }) => {
+  const [appConfig, setAppConfig] = useState({} as AppConfig);
+  useEffect(() => {
+    const getConfig = async () => {
+      const response = await fetch(`/api/config`);
+      if (response.status === 200) {
+        const configJson = await response.json();
+        if (!Object.keys(appConfig).length) {
```
dfellis, line 27: Why this conditional?

Dockerfile
```diff
@@ -46,14 +46,17 @@ FROM build AS dashboard-stage
 
 WORKDIR /dashboard
 
+## Install stage dependencies
+COPY dashboard/package.json dashboard/yarn.lock ./
+RUN yarn install
```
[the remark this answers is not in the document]
aguillenv: It would not be a bad idea. I did it in the tests because there's a breaking change with `playwright` and did not want to add more changes to this PR.

Dockerfile
aguillenv: `node_modules` layer should be defined as early as we can in the Dockerfile to take advantage of the cache. The engine's `niode_modules` take a considerable time for a small code change.

dashboard/package.json
aguillenv: the next.js `standalone` mode generates a package with everything we need and it is just a node app.

Dockerfile
aguillenv: since we are now running the prod version as a node app we cannot pass the port forwarding there, so we defined it as an env var here otherwise it took the default 3000. Do you want me to change the default one instead?

dashboard/src/components/providers/ConfigProvider.tsx
aguillenv: this method should be called just once. This is just a make sure to only set the values once and to avoid a re-rendering if for any reason it gets called again.

Dockerfile
aguillenv: Hmm I see, so the `.yarnrc` need to exists before `yarn install` is called, right?

Dockerfile
dfellis: No, I mean you could just have:


```json
"start:docker": "PORT=9876 node server.js",
```

To encapsulate that config there?

dashboard/package.json
dfellis: Yeah, but I don't see that file being copied over in the `Dockerfile`?

Dockerfile
dfellis: Yes

dashboard/package.json
aguillenv: We are copying the whole folder https://github.com/iasql/iasql/pull/2043/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R107

dfellis: Ohh... I didn't notice that `standalone` was getting copied into the `/dashboard` root.

dfellis (approved):

depombo (approved):

PR #2046, opened 2023-02-23
Deploy IaSQL on EC2 using IaSQL Post

mtp1376:


site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,184 @@
...
+```
+The first query simply creates a security group named `iasql_sg`. The second query seems a bit hard at the first sight, but it can be re-written as multiple `INSERT` queries:
+```sql
+INSERT INTO security_group_rule (is_egress, ip_protocol, from_port, to_port, cidr_ipv4, description, security_group_id)
+VALUES (FALSE, 'tcp', 22, 22, '0.0.0.0/0', 'ssh port is accessible from all ips',
+        (SELECT id FROM security_group WHERE group_name = 'iasql_sg'));
+```
+The reason we've written it that way is to avoid repeating the inner `SELECT id FROM security_group` query four times. The `security_group_id` column has an FK set to the `security_group` table's `id` column. Therefore, we need to insert the correct `id` for the security group we just created.
+
+Since the IaSQL dashboard does not have a default authorization mechanism set, remember to replace `<your-ip>` value in the above queries with your own IP address.
```
aguillenv, line 86: maybe highlight this as a warning or important note?

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,184 @@
...
+```
+- `protocol`: Specifies the type of resource that is being referenced. The currently supported protocols are `ssm` for AWS Systems Manager Parameter Store, `secretsmanager` for AWS Secrets Manager secrets, and `s3` for Amazon S3 bucket objects.
+- `value`: Specifies the path to the resource that is being referenced.
+
+Using the AWS Resolve syntax in AWS SDK or EC2 user data script allows you to dynamically reference and use the values of resources in other AWS services, which can be useful for automating the deployment of AWS resources.
+
+For example, you could use `resolve:ssm:/my/parameter/path` to reference a parameter stored in AWS Systems Manager Parameter Store, `resolve:secretsmanager:/my/secret/path` to reference a secret stored in AWS Secrets Manager, or `resolve:s3:/my/bucket/path/myobject.txt` to reference an object stored in an Amazon S3 bucket.
+:::
+
+- `instance_type`: this column contains the [instance type](https://aws.amazon.com/ec2/instance-types/) of the EC2 server. We'll use a `t2.small` instance for our deployment.
```
aguillenv, line 123: does not work on a `t2.micro`, right? just to see if we can stick to the free tier.

mtp1376:
@dfellis : trying with `USING` to simplify the query

depombo:
This is looking really good!

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,184 @@
...
+       t.from_port,
+       t.to_port,
+       t.cidr_ipv4::cidr,
+       t.description,
+       security_group.id
+FROM security_group,
+     (VALUES (FALSE, 'tcp', 22, 22, '0.0.0.0/0', 'ssh port is accessible from all ips'),
+             (FALSE, 'tcp', 9879, 9879, '<your-ip>/0', 'open dashboard to just me'),
+             (FALSE, 'tcp', 5432, 5432, '0.0.0.0/0', 'postgres port is accessible from all ips'),
+             (TRUE, 'tcp', 1, 65535, '0.0.0.0/0',
```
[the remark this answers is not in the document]
mtp1376: If we set the protocol to `-1` it will work. But if we set the protocol to `tcp`, it needs to explicitly set the port range.

site/blog/tutorials/iasql-on-ec2.mdx
mtp1376: Sure.

site/blog/tutorials/iasql-on-ec2.mdx
mtp1376: I'll investigate this. Makes sense to stick to free tier.

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,184 @@
...
+If you want to be able to access the instance through SSH, you can import your public key into AWS by executing the following IaSQL RPC:
+```sql
+SELECT * FROM key_pair_import('iasql', '<your-public-key-content>', 'us-east-1');
+```
+
+This will direct IaSQL to create an [AWS key pair](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/create-key-pairs.html#how-to-generate-your-own-key-and-import-it-to-aws) with the name `iasql` in the `us-east-1` region, and with your public key contents.
+
+:::note
+An IaSQL RPC is a function defined in Postgres that upon calling, sends a request to the IaSQL engine to run a pre-define logic and returns results to the Postgres console. In the above example [`key_pair_import`](https://iasql.com/docs/modules/aws/tables/aws_ec2_rpcs_import.KeyPairImportRpc/) is an IaSQL RPC that triggers the IaSQL engine to send a request to AWS for importing the public key.
+:::
```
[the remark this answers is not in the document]
mtp1376: It's popping in a "note" component which I've used throughout the article to say things that might be outside the context of the article itself. But yes, we need a section on this RPCs in our docs.

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,260 @@
+---
+slug: iasql-on-ec2
+title: Deploy IaSQL on EC2 using IaSQL
+authors: [mtp1376]
+---
+
+We're intending to publish a series of blog posts on deploying different services using IaSQL. IaSQL covers a wide range of AWS services, so there are lots of possibilities to explore. You can deploy containers to ECS, launch servers using EC2, and have CodePipelines that take the burden of building and deploying your applications.
```
yrobla, line 7: CodePipelines as it is is not a name used by aws... maybe "and use AWS Codepipeline to take the burden..."

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+### Deploy IaSQL on EC2 with IaSQL
+
+#### Overview
+Let's begin by taking an overview of what we're going to do to give you a general idea of the parts that are going to be deployed, before we move on to explore the specifics.
+
+To deploy IaSQL on EC2, we need to:
+
+- Launch an EC2 instance that will host IaSQL
+- Set the EC2 user data so that Docker will be installed upon instance launch, and the IaSQL container will be started
```
yrobla, line 34: set the instance user data ?

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+If you want to be able to access the instance through SSH, you can import your public key into AWS by executing the following IaSQL RPC:
+```sql
+SELECT * FROM key_pair_import('iasql', '<your-public-key-content>', 'us-east-1');
+```
+
+This will direct IaSQL to create an [AWS key pair](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/create-key-pairs.html#how-to-generate-your-own-key-and-import-it-to-aws) with the name `iasql` in the `us-east-1` region, and with your public key contents.
+
+:::note
+An IaSQL RPC is a function available in Postgres that upon calling, sends a request to the IaSQL engine to run a pre-define logic and returns results to the Postgres console. In the above example [`key_pair_import`](https://iasql.com/docs/modules/aws/tables/aws_ec2_rpcs_import.KeyPairImportRpc/) is an IaSQL RPC that triggers the IaSQL engine to send a request to AWS for importing the public key.
```
yrobla, line 57: nit, predefined

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,260 @@
...
+Let's begin by taking an overview of what we're going to do to give you a general idea of the parts that are going to be deployed, before we move on to explore the specifics.
+
+To deploy IaSQL on EC2, we need to:
+
+- Launch an EC2 instance that will host IaSQL
+- Set the EC2 user data so that Docker will be installed upon instance launch, and the IaSQL container will be started
+- Import an SSH key pair into AWS to be able to access the EC2 instance afterwards
+- Set a security group for the instance to allow SSH, while disallowing unauthorized access to the dashboard
+
+<img src={"/img/iasql-on-ec2.png"} alt={"IaSQL dashboard"} />
```
dfellis, line 38: Will there be dark-mode variants for all of these images?

mtp1376: I'll add them. It makes sense.

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,284 @@
...
+VALUES (FALSE, 'tcp', 22, 22, '0.0.0.0/0', 'ssh port is accessible from all ips', (SELECT id FROM sg)),
+       (FALSE, 'tcp', 9879, 9879, '<your-ip>/0', 'open dashboard to just me', (SELECT id FROM sg)),
+       (FALSE, 'tcp', 5432, 5432, '0.0.0.0/0', 'postgres port is accessible from all ips', (SELECT id FROM sg)),
+       (TRUE, 'tcp', 1, 65535, '0.0.0.0/0', 'allow outgoing traffic to anywhere', (SELECT id FROM sg));
+```
+The first query simply creates a security group named `iasql_sg`. The second query uses a [`WITH` Postgres query](https://www.postgresql.org/docs/current/queries-with.html) to select the `id` of our security group to be used later in the `INSERT` query. The `WITH` query can be though of as a temporary table that exists just for one query. By use of `WITH` we can simplify our query and preload the `id` value from the `security_group` table. We could've written the query using [subqueries](https://www.postgresql.org/docs/current/functions-subquery.html) as follows, but that way we should've repeated the `SELECT id FROM security_group WHERE ...` part for each set of values:
```
yrobla, line 104: /though/thought

yrobla, line 104: i'm not sure if explaining the syntax of the query and reasoning for using WITH is in the scope of this post. But open to other opinions

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,284 @@
...
+INSERT INTO security_group (description, group_name)
+VALUES ('IaSQL security group', 'iasql_sg');
+
+WITH sg as (SELECT id
+            FROM security_group
+            WHERE group_name = 'iasql_sg')
+INSERT
+INTO security_group_rule (is_egress, ip_protocol, from_port, to_port, cidr_ipv4, description, security_group_id)
+VALUES (FALSE, 'tcp', 22, 22, '0.0.0.0/0', 'ssh port is accessible from all ips', (SELECT id FROM sg)),
+       (FALSE, 'tcp', 9879, 9879, '<your-ip>/0', 'open dashboard to just me', (SELECT id FROM sg)),
```
aguillenv, line 100: Hmm, can't you just do
```suggestion
       (FALSE, 'tcp', 9879, 9879, '<your-ip>/0', 'open dashboard to just me', sg.id),
```

site/blog/tutorials/iasql-on-ec2.mdx
mtp1376: I think this post already has a lot of things that are not directly the scope of it :D. Like discussing `cloud-init` directives. Do you think it should solely talk about the deployment, and not explain the concepts that are mentioned?

aguillenv: yes, I think it will depend on how much we want to explain SQL concepts or assume it is known by the reader.

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,284 @@
...
+WHERE id = (SELECT id
+            FROM instance
+            WHERE tags ->> 'name' = 'iasql'
+            LIMIT 1);
+```
+
+Now we can go to `http://<instance-ip>:9876/` URL, and we'll see the beautiful IaSQL dashboard deployed on our EC2 server.
+
+##### Cloud-init Directives
+Now let's use cloud-init directives to do the same thing. This part is not exactly related to IaSQL, or what we intended to do in this post. It's just an exploration for other choices we have with EC2 user data field.
```
aguillenv, line 217: Since this section as you say is not directly related to the scope of this article can we explicitly mark it as optional or something like that?

site/blog/tutorials/iasql-on-ec2.mdx
mtp1376: That's what I initially tested. It would've been great if that worked, but it fails with this error:
```
[42P01] ERROR: missing FROM-clause entry for table "sg"
```

aguillenv: indeed! Well maybe instead of `VALUES` you could use `SELECT` and then `FROM sg`, but well I think it is ok as is too!! just thinking alternatives

mtp1376: How would you write it using `SELECT`? This is not working:
```sql
WITH sg as (SELECT id
            FROM security_group
            WHERE group_name = 'iasql_sg' LIMIT 1)
INSERT
INTO security_group_rule (is_egress, ip_protocol, from_port, to_port, cidr_ipv4, description, security_group_id)
SELECT (FALSE, 'tcp', 22, 22, '0.0.0.0/0', 'ssh port is accessible from all ips', sg.id),
       (FALSE, 'tcp', 9879, 9879, '0.0.0.0/0', 'open dashboard to just me', sg.id),
       (FALSE, 'tcp', 5432, 5432, '0.0.0.0/0', 'postgres port is accessible from all ips', sg.id),
       (TRUE, 'tcp', 1, 65535, '0.0.0.0/0', 'allow outgoing traffic to anywhere', sg.id)
FROM sg;
```
This won't work becuase it will treat the whole `(FALSE, 'tcp', 22, 22, '0.0.0.0/0', 'ssh port is accessible from all ips', sg.id)` as `is_egress`. Also if I remove the paranthesis then we can't insert multiple rows together.

aguillenv: oh interesting, I was somehow expecting that query to work, but I guess it makes sense. Let's keep it as is!

site/blog/tutorials/iasql-on-ec2.mdx
mtp1376: Is there a class for marking section as optional? Or something similiar?

aguillenv: no idea!

mtp1376: There is collapsible parts in `mdx`, but it's making the markdown really hard to edit.

aguillenv: maybe just add an  something like:
```suggestion
##### Advanced: Cloud-init Directives
Now let's use cloud-init directives to do the same thing. This part is not exactly related to IaSQL, or what we intended to do in this post. It's just an exploration for other choices we have with EC2 user data field.
```

or 

```suggestion
##### Optional: Cloud-init Directives
Now let's use cloud-init directives to do the same thing. This part is not exactly related to IaSQL, or what we intended to do in this post. It's just an exploration for other choices we have with EC2 user data field.
```

mtp1376: Used the second one. Thanks!

aguillenv (approved):
It looks good to me. Pending on what the other thinks about the scope.

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,284 @@
...
+Knowing the intent of each command, it's not hard to do the same functionality using `cloud-init`.
+```yaml
+#cloud-config
+package_update: true # apt-get update - this was in the `get-docker.sh` script so we didn't manually run it in our script
+package_upgrade: true # apt-get upgrade
+apt:
+  sources:
+    docker.list: # add the docker source for the package manager
+      source: deb [arch=amd64] https://download.docker.com/linux/ubuntu $RELEASE stable
+      keyid: 9DC858229FC7DD38854AE2D88D81803C0EBFCD88
```
dfellis, line 240: This section here undermines the narrative below because it is still `apt`-specific logic.

However, I think it is salvageable with some other changes.

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,284 @@
...
+```sql
+INSERT INTO instance (ami, instance_type, subnet_id, tags, user_data, key_pair_name, region)
+SELECT 'resolve:ssm:/aws/service/canonical/ubuntu/server/22.04/stable/current/amd64/hvm/ebs-gp2/ami-id',
+       't2.small',
+       subnet.id,
+       '{"name":"iasql"}',
+       '#!/bin/bash
+     sudo apt-get upgrade -y
+     curl -fsSL https://get.docker.com -o get-docker.sh
+     sudo sh get-docker.sh
```
dfellis, line 177: We can simplify this by using the 100% OSS [`docker.io`](https://packages.ubuntu.com/jammy/docker.io) package instead of Docker's official package.

```suggestion
     sudo apt-get upgrade -y
     sudo apt-get install -y docker.io
```

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,284 @@
...
+
+##### Optional: Cloud-init Directives
+Now let's use cloud-init directives to do the same thing. This part is not exactly related to IaSQL, or what we intended to do in this post. It's just an exploration for other choices we have with EC2 user data field.
+
+For the previous part, we used the following bash script as the `user_data` column in the `instance` table:
+```bash
+#!/bin/bash
+sudo apt-get upgrade -y
+curl -fsSL https://get.docker.com -o get-docker.sh
+sudo sh get-docker.sh
```
dfellis, line 224: If the suggestion to use `docker.io` above is followed, then this also becomes:

```suggestion
sudo apt-get install -y docker.io
```

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,284 @@
...
+package_update: true # apt-get update - this was in the `get-docker.sh` script so we didn't manually run it in our script
+package_upgrade: true # apt-get upgrade
+apt:
+  sources:
+    docker.list: # add the docker source for the package manager
+      source: deb [arch=amd64] https://download.docker.com/linux/ubuntu $RELEASE stable
+      keyid: 9DC858229FC7DD38854AE2D88D81803C0EBFCD88
+packages: # make sure docker is installed
+  - docker-ce
+  - docker-ce-cli
```
dfellis, line 243: If the instruction above is followed, we can trim this to:

```suggestion
packages: # make sure docker is installed
  - docker.io
```

It *doesn't* actually resolve the cross-platform issues with this Yaml format, because only Debian-derived distros call it `docker.io`, but now it doesn't ruin the text below.

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,284 @@
...
+
+Good thing about cloud-init is that it's widely supported: you can use it on GCP, AWS, DigitalOcean, and many other cloud providers to provision your servers. Their documentation says:
+> `Cloud-init` is the industry standard multi-distribution method for cross-platform cloud instance initialisation. It is supported across all major public cloud providers, provisioning systems for private cloud infrastructure, and bare-metal installations.
+
+:::note
+If you want to see examples of cloud-init, [this video](https://www.youtube.com/watch?v=-zL3BdbKyGY) and [this article](https://www.digitalocean.com/community/tutorials/an-introduction-to-cloud-config-scripting) might be a good starting point.
+:::
+
+### Next Steps
+In this post, we deployed IaSQL on EC2 server with just using IaSQL and PostgreSQL queries. We'll continue this post with example deployments of other famous apps from now on. Deploying them won't be any harder than this one!
```
dfellis, line 284: I don't think this is relevant if you follow the change at the very beginning of the post.

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,284 @@
...
+    if is_darwin; then
+      echo "ERROR: Unsupported operating system 'macOS'"
+      echo "Please get Docker Desktop from https://www.docker.com/products/docker-desktop"
+    fi
+esac
+```
+:::
+
+Good thing about cloud-init is that it's widely supported: you can use it on GCP, AWS, DigitalOcean, and many other cloud providers to provision your servers. Their documentation says:
+> `Cloud-init` is the industry standard multi-distribution method for cross-platform cloud instance initialisation. It is supported across all major public cloud providers, provisioning systems for private cloud infrastructure, and bare-metal installations.
```
dfellis, line 277: Not really a comment on this itself, but I consider this `cloud-init` yaml format misguided because you can't escape platform-specific information in it.

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -0,0 +1,284 @@
...
+        (SELECT id FROM security_group WHERE group_name = 'iasql_sg')),
+       (TRUE, 'tcp', 1, 65535, '0.0.0.0/0', 'allow outgoing traffic to anywhere',
+        (SELECT id FROM security_group WHERE group_name = 'iasql_sg'));
+```
+The `security_group_id` column has an FK set to the `security_group` table's `id` column. Therefore, we need to insert the correct `id` for the security group we just created.
+
+:::warning
+Since the IaSQL dashboard does not have a default authorization mechanism set, remember to replace `<your-ip>` value in the above queries with your own IP address.
+
+You can find your current IP address by visiting [api.ipify.org](https://api.ipify.org/). Just remember that most internet users don't have a static IP address and your IP address will be changed each time your router reconnects.
```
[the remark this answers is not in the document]
mtp1376: `curl -s api.ipify.org` 😄

site/blog/tutorials/iasql-on-ec2.mdx
mtp1376: I thought about this `cloud-init` part and I think if the commands aren't cross-distro enough, it won't make sense to have it in this post. I can come to it in some other post that we can actually leverage it to have something cool.

depombo (commented):
I can't leave comments on the images themselves, but:
- the dashboard screenshots already exist in the static img folder so let's reuse those
- the architecture diagram contains some information that is not so important like AMI and not much about EIP or EBS volume where the actual Postgres database is stored. Let's make it something like this https://www.reddit.com/r/webdev/comments/zpsseu/my_saas_architecture_tech_stack_on_aws_as_a_solo/ which is oriented around the user and how data flows wrt to them

dfellis (approved):

mtp1376:
> I can't leave comments on the images themselves, but:
> 
> * the dashboard screenshots already exist in the static img folder so let's reuse those
> * the architecture diagram contains some information that is not so important like AMI and not much about EIP or EBS volume where the actual Postgres database is stored. Let's make it something like this https://www.reddit.com/r/webdev/comments/zpsseu/my_saas_architecture_tech_stack_on_aws_as_a_solo/ which is oriented around the user and how data flows wrt to them

I think you're suggesting we use an EBS volume for the database, and persist the data. Currently we're not doing that.
@depombo

depombo:
> > I can't leave comments on the images themselves, but:
> > 
> > * the dashboard screenshots already exist in the static img folder so let's reuse those
> > * the architecture diagram contains some information that is not so important like AMI and not much about EIP or EBS volume where the actual Postgres database is stored. Let's make it something like this https://www.reddit.com/r/webdev/comments/zpsseu/my_saas_architecture_tech_stack_on_aws_as_a_solo/ which is oriented around the user and how data flows wrt to them
> 
> I think you're suggesting we use an EBS volume for the database, and persist the data. Currently we're not doing that. @depombo

got it. we should still have the EIP piece in the diagram, plus avoid duplicating the screenshots

mtp1376:
> > > I can't leave comments on the images themselves, but:
> > > 
> > > * the dashboard screenshots already exist in the static img folder so let's reuse those
> > > * the architecture diagram contains some information that is not so important like AMI and not much about EIP or EBS volume where the actual Postgres database is stored. Let's make it something like this https://www.reddit.com/r/webdev/comments/zpsseu/my_saas_architecture_tech_stack_on_aws_as_a_solo/ which is oriented around the user and how data flows wrt to them
> > 
> > 
> > I think you're suggesting we use an EBS volume for the database, and persist the data. Currently we're not doing that. @depombo
> 
> got it. we should still have the EIP piece in the diagram, plus avoid duplicating the screenshots

Also for EIP, we're not using a static IP in this version. Should I add it? I ask for both EIP and the EBS volume. We can have them if you want.

I'll remove the duplicate images.

mtp1376:
- Add ELB for SSL certificate
- Add ACM certificate
- Add Route53 entry for domain

depombo (approved):
looks good, my comments can be addressed in a follow up PR if you prefer

can we put the dashboard screenshots under the `img/screenshots` folder and the diagram pictures under either `img/diagrams` or `img/iasql-on-ec2`? otherwise the `img/` folder will be hard to keep organized

> * Add ELB for SSL certificate
> * Add ACM certificate
> * Add Route53 entry for domain

Did you want to do this in another PR? I think it would be best if it is in this same post, but it can happen over multiple PRs. That way we can have a diagram like this one that talks about how we deploy our own SaaS: https://www.reddit.com/r/webdev/comments/zpsseu/my_saas_architecture_tech_stack_on_aws_as_a_solo/. Created a follow up task: https://github.com/iasql/iasql/issues/2211

yrobla (approved):

PR #2050, opened 2023-02-24
Log Docker Image Size in PR

mtp1376:


depombo (commented):
you can probably add it to the same message as the install/uninstall timings?

mtp1376:
> 📏 Image size: 359.73MB (compressed) - 1.13GB (uncompressed) ⏳ All modules installed in: 41.562532163 seconds ⌛️ ⏳ All modules uninstalled in: .942570979 seconds ⌛️

This is the message :D

.github/workflows/tests.yml
```diff
@@ -576,20 +587,19 @@ jobs:
         run: |
           curl -X POST -H "Content-Type: application/json" -H "${SECRET_HEADER}" -H "x-iasql-drop: true" -d "[${ACCOUNT_INDEX}]" ${LAMBDA_FUNCTION}
 
-  notify-install-all-modules-time:
+  log-image-size:
```
aguillenv, line 590: lets rename the jobs to something more generic since it is not only the image size

.github/workflows/tests.yml
```diff
@@ -576,20 +587,19 @@ jobs:
...
     runs-on: ubuntu-latest
     needs:
       - get-install-times
```
aguillenv, line 593: same here, let's rename it

aguillenv (commented):
Other than the rename it looks good!

aguillenv (approved):

yrobla (approved):

PR #2069, opened 2023-02-28
Add `ph-no-capture` to query result table

mtp1376:


yrobla (approved):

dashboard/src/components/common/Copyable.tsx
```diff
@@ -1,12 +1,14 @@
...
+export default function Copyable({ children, sensitive }: { children: any[] | any; sensitive?: boolean }) {
   const child = Children.toArray(children)[0];
   if (typeof child !== 'string') throw new Error('Copyable only works with text');
   const [showTooltip, setShowTooltip] = useState(false);
   return (
     <p
-      className='mt-1 max-w-2xl text-sm cursor-pointer text-primary break-words'
+      className={
+        (sensitive ? 'ph-no-capture ' : '') + 'mt-1 max-w-2xl text-sm cursor-pointer text-primary break-words'
+      }
```
aguillenv, line 11: nit: we are using here two different approaches, maybe just use the same everywhere? Either we add the `customClasses` property here or we add a `sensitive` property to the `Box` component?

dfellis: I don't know if this is in scope for this PR, though? The Posthog class addition is separate to the refactoring of this component to use `<Box>`, right?

dfellis (approved):

dashboard/src/components/common/Copyable.tsx
aguillenv: I think it belongs to this PR because both behaviours I'm commenting on have been introduced in this PR. It's not a big deal, but since it is small we might want to solve it now.

aguillenv (approved):

PR #2075, opened 2023-02-28
Fix export buttons

mtp1376:
Should not be merged before https://github.com/iasql/meta/issues/344


aguillenv:
can you please add a test here to avoid a regression? 🙏🏼

PR #2083, opened 2023-03-01
Make sure the connection name is not duplicated

### LINKED ISSUE #2055: TypeORM connection duplication issue on IaSQL metadata

aguillenv:
Resolves #2055

### DIFF
--- a/dashboard/src/pages/api/run.ts
+++ b/dashboard/src/pages/api/run.ts
@@ -4,7 +4,7 @@ import type { NextApiRequest, NextApiResponse } from 'next';
 import pg from 'pg';
 import format from 'pg-format';
 import { parse, deparse } from 'pgsql-parser';
-import { createConnection, Connection } from 'typeorm';
+import { DataSource } from 'typeorm';
 import { PostgresConnectionOptions } from 'typeorm/driver/postgres/PostgresConnectionOptions';
 
 import { throwError } from '@/config/config';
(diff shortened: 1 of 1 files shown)


mtp1376 (approved):

dashboard/src/pages/api/run.ts
```diff
@@ -68,7 +68,7 @@ async function metaQuery(sql: string, params?: any[]): Promise<any> {
   if (!dbConns[db]?.isConnected) {
     // Trash this object and rebuild it
     delete dbConns[db];
-    dbConns[db] = await createConnection({ ...baseConnConfig, name: db, database: db });
+    dbConns[db] = await createConnection({ ...baseConnConfig, name: `${db}${Date.now()}`, database: db });
```
dfellis, line 71: I think we need a lot more than that here. The `dbConns` object is global between requests so a concurrency-induced race condition here could cause connections to be open but not exist in that object because it was overwritten with another `dbConns[db] = ...` (specifically if two requests to the same DB come in faster than `await createConnection` comes back, one will overwrite the other).

If we are giving them all different names, then we aren't planning on re-using the connections at all and this `dbConns` lookup table is an overcomplication we can drop.

Otherwise we should look into changing the type of `dbConns[db]` into a `Promise` and `await` retrieval of it, so the assignment into that data structure is blocking and other concurrent queries can see it and `await` for the connection themselves?

dfellis (approved):

dashboard/src/pages/api/run.ts
depombo: why not just use a uuid?

Issue #2071, opened 2023-02-28
Disconnect button fails silently when there's a `-` in database name or database name has uppercase characters


mtp1376:
![Image](https://user-images.githubusercontent.com/5357435/221814588-64933cab-0c20-4e87-8c61-d5100131411f.png)

response:
```json
[{"statement":"SELECT * FROM iasql_disconnect('Test-Mohammad');","result":[{"iasql_disconnect":"syntax error at or near \"-\""}]}]
```
```
query: 'REVOKE CONNECT ON DATABASE Test-Mohammad FROM PUBLIC, postgres, rb7ngw7g;'
```

Reason: we're not escaping the database name in the local environment:
https://github.com/iasql/iasql/blob/bb3c67c7a34b0b593ef1195d60e7f357b20b502e/src/services/iasql.ts#L108
(`dbId` is set as it's passed in the local env:)
https://github.com/iasql/iasql/blob/507bbaf0695754d739adb2770bdc6bcab2a9c110/src/services/db-manager.ts#L39-L41

other place that will cause issues:
https://github.com/iasql/iasql/blob/bb3c67c7a34b0b593ef1195d60e7f357b20b502e/src/services/iasql.ts#L125

PR #2086, opened 2023-03-01
Use `pg-format` to avoid issues on literals or identifiers

mtp1376:


dfellis:
Everything else passes, but the `upgrade` test consistently fails: https://github.com/iasql/iasql/actions/runs/4305917731/jobs/7513070090

So I think there is a bug in the current implementation. :/

dfellis (approved):

yrobla (approved):

aguillenv (commented):
does this fix both issues with #2071?

PR #2089, opened 2023-03-02
Enable parallel queries between tabs a track their responses

### LINKED ISSUE #2077: Keep association between the query response and the tab that made the call

aguillenv:
Resolves #2077

### DIFF
--- a/dashboard/src/components/IasqlEditor.tsx
+++ b/dashboard/src/components/IasqlEditor.tsx
@@ -32,7 +32,6 @@ export default function IasqlEditor() {
     dispatch,
     isDarkMode,
     selectedDb,
-    isRunningSql,
     installedModules,
     functions,
     token,
@@ -59,7 +58,7 @@ export default function IasqlEditor() {
   );
 
   const handleQueryToRunUpdate = useCallback(
-    (db: any, isRunning: boolean, tabIdx: number) => {
+    (db: any, tabIdx: number) => {
       const contentToBeRun = editorRef?.current?.editor?.getSelectedText()
         ? editorRef?.current?.editor?.getSelectedText()
         : editorRef?.current?.editor?.getValue();
@@ -70,7 +69,6 @@ export default function IasqlEditor() {
           data: {
             db,
             content: contentToBeRun,
-            isRunning,
             tabIdx,
           },
         });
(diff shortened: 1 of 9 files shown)


dashboard/src/components/IasqlEditor.tsx
```diff
@@ -102,11 +101,15 @@ export default function IasqlEditor() {
...
+      exec: () => handleQueryToRunUpdate(selectedDb, editorSelectedTab),
     };
-    editorRef?.current?.editor?.commands?.removeCommand(command);
-    editorRef?.current?.editor?.commands?.addCommand(command);
-  }, [handleQueryToRunUpdate, isRunningSql, selectedDb]);
+    if (editorTabs?.[editorSelectedTab]?.isRunning) {
+      editorRef?.current?.editor?.commands?.removeCommand(command);
+    } else {
+      editorRef?.current?.editor?.commands?.removeCommand(command);
+      editorRef?.current?.editor?.commands?.addCommand(command);
```
dfellis, line 110: How often is this branch executed, do you know? Per keystroke?

dashboard/src/components/common/Tab.tsx
```diff
@@ -14,56 +14,73 @@ export default function Tab({
...
   selectedIndex?: number;
   onChange?: (index: number) => void;
   onTabClose?: (i: number) => void;
+  isLoading?: boolean;
 }) {
+  const getTabContent = (
+    tab: { title: string; action?: () => void; className?: string; width?: string; closable?: boolean },
+    idx: number,
+  ) => {
+    if (tab.closable && selectedIndex === idx && tabs.length > 2) {
```
dfellis, line 31: Nit: If you moved `selectedIndex` into an argument to this function, could the function be pulled out of the component constructor and used as a normal function? The current structure will recreate a closure every time `Tab` is called.

dashboard/src/components/providers/AppProvider.tsx
```diff
@@ -446,36 +449,31 @@ const middlewareReducer = async (
       let updatedAutoCompleteRes: any;
       let runErr: any;
       try {
```
dfellis, line 451: Can this outer `try` even catch anything anymore? The only uncaught line is the `dispatch` line, which should be an infallible function call.

I think you can drop the outer try-catch-finally, deleting the catch block and always running the finally block.

dashboard/src/components/IasqlEditor.tsx
aguillenv: No, either when a new db is selected or when something change in the tabs.

dfellis (approved):

depombo (approved):

yrobla (approved):

PR #2093, opened 2023-03-03
Engine issues found upgrading Typeorm

aguillenv:
This is a mini-RFC to explain the issues found during the Typeorm upgrade and to discuss possible solutions.

Actually, I'm not sure if it should be an RFC but didn't know where I could post this to discuss it.

### DIFF
--- a/rfcs/007 - Engine tweaks.md
+++ b/rfcs/007 - Engine tweaks.md
@@ -1,0 +1,113 @@
+# 007 - Engine tweaks
+
+### Proposed
+
+2023-03-03
+
+### Accepted
+
+YYYY-MM-DD
+
+#### Approvers
+
+- David Ellis <david@iasql.com>
+- Yolanda Robla <yolanda@iasql.com>
+- Mohammad Pabandi <mohammad@iasql.com>
+- Luis Fernando De Pombo <luisfer@iasql.com>
+
+### Implementation
+
+- [ ] Implemented: [One or more PRs](https://github.com/iasql/iasql-engine/some-pr-link-here) YYYY-MM-DD
+- [ ] Revoked/Superceded by: [RFC ###](./000 - RFC Template.md) YYYY-MM-DD
+
+## Author(s)
+
+- Alejandro Guillen <alejandro@iasql.com>
+
+## Summary
+
+During the typeorm upgrade, David found a bug in the function recreating the entities. It was passing to the `findOne` the wrong attributes. Why tests in `main` were working? Because the `apply` algorithm does an `apply` with filters, but if that goes wrong (as it was happening) it catches the error, shows a warning and continues with a full `apply` (no filters, so no bug recreating entities).
+
+Fixing this bug recreating the entities has surfaced at least 2 new cases that we were not handling correctly as we could see in the tests failing (ec2 and ecs).
+
+### Cases found
+
+#### EC2 integration tests
+
+Here the issue was with the `RegisteredInstance` entity, its function is to link instances with target groups. The `entityId` it's a combination of all its fields. One of the fields is the `instanceId` coming from the `Instance` entity. The problem was that the instance was getting replaced (delete + create) and that should trigger an update for the `RegisteredInstance` entity since one of the values of the `entityId` has changed. The problem is that this latest change is not seen because we are filtering in the `apply` loop by the entities that have changed based on the audit log. Then the `sync` just removes the record from the DB.
+
+#### ECS integration tests
+
+The problem in ecs. We were calling from the `TaskDefinition`'s `cloud.update` function the `db.create` function to force the creation of a new task definition. Again, the filter in the `apply` loop is not seeing this change because it was not done by the user and it was done after the commit started.
+
+#### Temporal solution
+
+Both test suites are passing now because we were able to add the right logic in each mapper to handle this, but somehow it's breaking module encapsulation which we want to keep at a minimum or inexistent.
+
+### High-level steps of the current `commit` algorithm
+
+1. Get relevant audit logs
+2. Get relevant modules based on step 1
+3. If any change:
+    1. Execute partial `apply` (passing only the relevant changes and modules)
+    2. If error on partial `apply`, execute full `apply`
+4. Execute full `sync`
+5. If full `apply` or `sync` error, execute `revert`
+6. Return the `sync` response
+
+### High-level steps of the current `apply` algorithm
+
+Outer loop:
+1. Execute `db.read`s
+2. Recreate changes after the commit started
+
+    Inner loop:
+    1. Execute `cloud.read`s
+    2. If necessary, recreate relevant changes
+    3. Execute `diff`ing logic
+    4. If relevant changes, filter them
+    5. If no relevant changes, exclude changes done after `commit` started
+    6. Mappers execution
+
+### High-level steps of the current `sync` algorithm
+
+Outer loop:
+1. Execute `cloud.read`s
+
+    Inner loop:
+    1. Execute `db.read`s
+    2. Recreate changes after the commit started
+    3. Execute `diff`ing logic
+    5. Exclude changes done after `commit` started
+    6. Mappers execution
+
+## Proposal
+
+TBD
+
+## Alternatives Considered
+
+It's tricky to find a common solution for both cases found. For the ECS case might be simpler, we need to find a way to also see the changes done by the engine during the `commit`. In the EC2 case, we need either to ignore the relevant changes and `apply` for all modules or find a way to "mark" entities that might be affected by another entity change and somehow include them in the filter.
+
+### Executing "full" `apply`
+This is one possible solution that could solve both issues. Would be similar to the current algorithm but we do not get "relevant changes". We always `apply` for all modules and only exclude changes done **by the user** after `commit`.
+
+#### Pros 
+- Solves the 2 issues
+- Easier to implement?
+
+#### Cons
+- We lose the "faster" `apply` since we need to iterate through all modules installed and not only the ones with changes.
+
+## Expected Semver Impact
+
+This would be considered a minor version bump as it would only involve the addition of an IaSQL function.
+
+## Affected Components
+
+The IaSQL `commit` function.
+
+## Expected Timeline
+
+1 week
+


rfcs/007 - Engine tweaks.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Executing "full" `apply`
+This is one possible solution that could solve both issues. Would be similar to the current algorithm but we do not get "relevant changes". We always `apply` for all modules and only exclude changes done **by the user** after `commit`.
+
+#### Pros 
+- Solves the 2 issues
+- Easier to implement?
+
+#### Cons
+- We lose the "faster" `apply` since we need to iterate through all modules installed and not only the ones with changes.
```
dfellis, line 100: A second con: if there is a change in the cloud that hasn't yet been synced, the full apply will blow it away and then the follow-up sync step will do nothing (in fact, there is no point in doing the sync at all, anymore, because the DB becomes the sole source-of-truth and we only sync during install).

postmortems/004 - Postmortem for v0.1.3-beta TypeORM upgrade to v0.3.x.md
```diff
@@ -40,9 +23,35 @@ Here the issue was with the `RegisteredInstance` entity, its function is to link
 
 The problem in ecs. We were calling from the `TaskDefinition`'s `cloud.update` function the `db.create` function to force the creation of a new task definition. Again, the filter in the `apply` loop is not seeing this change because it was not done by the user and it was done after the commit started.
 
-#### Temporal solution
 
-Both test suites are passing now because we were able to add the right logic in each mapper to handle this, but somehow it's breaking module encapsulation which we want to keep at a minimum or inexistent.
+## Timeline
+
+- **2023-02-21**: @dfellis started working on the typeorm update
+- **2023-02-21**: Tests on ec2 and ec2 were the only ones failing
```
dfellis, line 30:
```suggestion
- **2023-02-21**: Tests on ec2 and ecs were the only ones failing
```

postmortems/004 - Postmortem for v0.1.3-beta TypeORM upgrade to v0.3.x.md
```diff
@@ -40,9 +23,35 @@ Here the issue was with the `RegisteredInstance` entity, its function is to link
...
+- **2023-02-27**: @dfellis found a bug recreating the entities and fix it using the right Typeorm property names adn structures
```
dfellis, line 31:
```suggestion
- **2023-02-27**: @dfellis found a bug recreating the entities and fix it using the right Typeorm property names and structures
```

dfellis (commented):
I found a couple of typos. Want to think for a bit on more alternatives to consider.

postmortems/004 - Postmortem for v0.1.3-beta TypeORM upgrade to v0.3.x.md
```diff
@@ -0,0 +1,107 @@
...
+Both test suites are passing now because we were able to add the right logic in each mapper to handle this, but it's breaking module encapsulation which we want to keep at a minimum or inexistent.
+
+## Cause
+
+For the 2-way mode we needed to check the audit logs and recreate the entities to then use them in a filter and only apply relevant changes per commit operation. To do that entity recreation, we needed to go trough the Typeorm internals and by mistake we were passing the wrong properties when recreating relationships.
+
+## Prevention
+
+The issue was covered by typeorm not hard failing before and our try/catch on the apply to run a full one if something failed, which made this pass for a long time silently.
+We might need to so some tweaks to the current `commit` algorithm to handle these 2 cases we found during this upgrade.
```
dfellis, line 54: To summarize the issues above, the ECS error was the `cloud.update` path making a DB change that was misinterpreted because of the user involved and then the `sync` step reverted it, while the EC2 error was no explicit change in the database itself, but the conceptual change of the entity because the instance it is pointing at changed.

The first one is an explicitly pushed change by the function, and the fix was to actually create the new `TaskDefinition` in the cloud and then let the `sync` step pull that in. I think that fix is *correct* and there is nothing special that needs to be done there. If we tried to change the user of db creation we would actually be creating a false history in the audit log. This way is actually clear in what is going on:

1. The user wants to edit a `TaskDefinition` and does so.
2. The mapper knows that the set of `TaskDefinitions` is an append-only structure within AWS so it restores the original DB record and creates a new `TaskDefinition` based on the user's change.
3. The `sync` stage sees this *new* `TaskDefinition` that wasn't *explicitly* created by the user, and pulls it into the database.

The issue with `RegisteredInstance` is much tougher. It should not be the responsibility of `Instance` to know that there's a "parasitic" entity like this attached to it and manually calling it's update logic, so the current solution *is* a hack, but there was also no DB change to indicate it should be looked at. Going through the list of changes since the last commit in the audit log will *not* show this `RegisteredInstance` record so there's nothing to "keep" in the filter.

What we need is some sort of reverse lookup mechanism. Currently there is a one entity -> one mapper relationship, and the entity in question is passed directly to the mapper logic. We need something that will:

1. Look up `N` mappers for an entity, where mappers can register an "interest" in more than one entity.
2. Take the input entity and determine the mapper's relevant entity. Such as looking for the `RegisteredInstance` record(s) given an `Instance` record.

99% of the time, these will both be identity functions; there's only one mapper relevant for a given entity, and the relevant entity for that mapper is the entity itself, so it should also have an automatic default that we override only when needed.

This could look something like:

```ts
relevantEntities: {
  RegisteredInstance: (e: RegisteredInstance): RegisteredInstance => e,
  Instance: async (e: Instance, ctx: Context): Promise<RegisteredInstance[]> => {
    return await ctx.orm.find(RegisteredInstance, {
      where: {
        instanceId: e.id,
      },
    });
  },
  ...
}
```

Then we have some internal data structure to group these by entity on the first level, the mapper on the second level, with `SomeObj[entityName].map(mapper => mapper.relevantEntities[entityName](entity))` returning an array of promises (or not but should be fine with a `Promise.all`) providing the actual set of entities to consider changed.

(To make this a set it would also need them all to be checked with `entityId` for uniqueness to eliminate any duplicates that might arise.)

It does seem like a lot of complexity, though. If you think this is a reasonable approach, let's include it in the `Prevention` section, *but* with a notice that we're turning it into a task to only pick up when we encounter such a weird entity again?

aguillenv: > To summarize the issues above, the ECS error was the `cloud.update` path making a DB change that was misinterpreted because of the user involved and then the `sync` step reverted it, while the EC2 error was no explicit change in the database itself, but the conceptual change of the entity because the instance it is pointing at changed.

Yes, perfectly summarized.


> The first one is an explicitly pushed change by the function, and the fix was to actually create the new `TaskDefinition` in the cloud and then let the `sync` step pull that in. I think that fix is _correct_ and there is nothing special that needs to be done there. If we tried to change the user of db creation we would actually be creating a false history in the audit log. This way is actually clear in what is going on:
> 
> 1. The user wants to edit a `TaskDefinition` and does so.
> 2. The mapper knows that the set of `TaskDefinitions` is an append-only structure within AWS so it restores the original DB record and creates a new `TaskDefinition` based on the user's change.
> 3. The `sync` stage sees this _new_ `TaskDefinition` that wasn't _explicitly_ created by the user, and pulls it into the database.
> 
> The issue with `RegisteredInstance` is much tougher. It should not be the responsibility of `Instance` to know that there's a "parasitic" entity like this attached to it and manually calling it's update logic, so the current solution _is_ a hack, but there was also no DB change to indicate it should be looked at. Going through the list of changes since the last commit in the audit log will _not_ show this `RegisteredInstance` record so there's nothing to "keep" in the filter.
> 
> What we need is some sort of reverse lookup mechanism. Currently there is a one entity -> one mapper relationship, and the entity in question is passed directly to the mapper logic. We need something that will:
> 
> 1. Look up `N` mappers for an entity, where mappers can register an "interest" in more than one entity.
> 2. Take the input entity and determine the mapper's relevant entity. Such as looking for the `RegisteredInstance` record(s) given an `Instance` record.
> 
> 99% of the time, these will both be identity functions; there's only one mapper relevant for a given entity, and the relevant entity for that mapper is the entity itself, so it should also have an automatic default that we override only when needed.
> 
> This could look something like:
> 
> ```ts
> relevantEntities: {
>   RegisteredInstance: (e: RegisteredInstance): RegisteredInstance => e,
>   Instance: async (e: Instance, ctx: Context): Promise<RegisteredInstance[]> => {
>     return await ctx.orm.find(RegisteredInstance, {
>       where: {
>         instanceId: e.id,
>       },
>     });
>   },
>   ...
> }
> ```
> 
> Then we have some internal data structure to group these by entity on the first level, the mapper on the second level, with `SomeObj[entityName].map(mapper => mapper.relevantEntities[entityName](entity))` returning an array of promises (or not but should be fine with a `Promise.all`) providing the actual set of entities to consider changed.
> 
> (To make this a set it would also need them all to be checked with `entityId` for uniqueness to eliminate any duplicates that might arise.)

Maybe we can call that proposed entity in the mapper `watchEntities` instead of `relevantEntities`?

Why do we need the identity in there? Why cannot be defined only with the entities we need to "watch"?

Another thing that I do not fully follow is where are we going to do this look-up, I believe it should be done *after* the `getChangesByEntity`, but *before* we start filtering the differences. Then we will need to add what the new entities to the record difference as `entitiesChanged` so they can be picked up by the engine.

> It does seem like a lot of complexity, though. If you think this is a reasonable approach, let's include it in the `Prevention` section, _but_ with a notice that we're turning it into a task to only pick up when we encounter such a weird entity again?

Agree with this plan!

dfellis: > Maybe we can call that proposed entity in the mapper watchEntities instead of relevantEntities?

That seems fine with me.

> Why do we need the identity in there? Why cannot be defined only with the entities we need to "watch"?

So, my thinking was that if this property is not defined on the mapper, it would automatically define a single entry for `this.entity` with something like this in the constructor:

```ts
if (!this.watchEntites) this.watchEntities = {
  [this.entity.name]: (e: type this.entity) => e,
}
```

(Not sure if that's the exact TS needed, but it's probably close)

If that's the case, then when you override this property, you need to define the identity case yourself.

Why? To give us the flexibility for a module to override the inclusion of a believed-to-be-changed record. For a truly read-only table that should only operate on sync, simply *blocking* all changes from the database from being considered and only implementing `cloud.read` would then work as the change would never be considered during `apply` and then during `sync` it would be reverted. That would let us reduce the amount of LOC in some modules that just have `cloud.create/update/delete` executing db operations to revert stuff.

> Another thing that I do not fully follow is where are we going to do this look-up, I believe it should be done after the getChangesByEntity, but before we start filtering the differences. Then we will need to add what the new entities to the record difference as entitiesChanged so they can be picked up by the engine.

I was thinking these calls should be integrated into both the raw diff generation *and* also applied directly to the results of `getChangesByEntity`. It should only need to be done on `entitiesChanged` of the three pools of diff results, I think, but we execute it on all of the `entitiesChanged` (using the value actually in the DB, not the old value) to append more entities to that list. Similarly we apply it to all of the entities in `getChangesByEntity` and *then* we filter like normal.

The reason we should do both is that the former is differences vs the clout and the latter is differences of the DB between commits. It is *possible* that there is a change that was identified due to user mutation in the database that *doesn't* show up when diffing against the cloud because someone also made that change manually in the cloud and now it isn't necessary. If we just appended the output of this logic to both lists instead of re-computing it on both, we could accidentally think there are changes to apply to the cloud that shouldn't be.

dfellis (approved):

postmortems/004 - Postmortem for v0.1.3-beta TypeORM upgrade to v0.3.x.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Detection
+
+The problem has been around since we landed the 2-way mode. It was working because we were doing a full `apply` when something goes wrong (as it was) and we were only reporting a warning but not `throw`ing an error.
+
+Both issues got detected during the typeorm upgrade, the tests failed and going through the engine logs to identify the unexpected behaviours.
+
+## Response
+
+Both test suites are passing now because we were able to add the right logic in each mapper to handle this, but it's breaking module encapsulation which we want to keep at a minimum or inexistent.
```
depombo, line 45: how exactly is it breaking module encapsulation?

postmortems/004 - Postmortem for v0.1.3-beta TypeORM upgrade to v0.3.x.md
```diff
@@ -0,0 +1,92 @@
...
+  },
+  ...
+}
+```
+
+Then we have some internal data structure to group these by entity on the first level, the mapper on the second level, with `SomeObj[entityName].map(mapper => mapper.relevantEntities[entityName](entity))` returning an array of promises (or not but should be fine with a Promise.all) providing the actual set of entities to consider changed (To make this a set it would also need them all to be checked with entityId for uniqueness to eliminate any duplicates that might arise.).
+
+  :::note
+
+      The proposed solution will be created as a task and its implementation will be delayed until we encounter with an use case similar to `RegisteredInstance`.
```
depombo, line 90: sounds great

depombo (approved):

yrobla (approved):

mtp1376 (commented):
For the last part using `:::note` directive, I think there's nowhere that renders that to a callout like in blog. Do we expose postmortems on our blog? If not, we can do it.

depombo:
> For the last part using `:::note` directive, I think there's nowhere that renders that to a callout like in blog. Do we expose postmortems on our blog? If not, we can do it.

This is correct. It only works on Docusaurus.

aguillenv:
> For the last part using `:::note` directive, I think there's nowhere that renders that to a callout like in blog. Do we expose postmortems on our blog? If not, we can do it.

No, we do not expose them, I think. That "note" was just to give it some emphasis.

PR #2122, opened 2023-03-06
properly emit identify

depombo:


aguillenv (commented):
hmm, what's the difference? It is just moving from one place to another but essentially the same, right?

dfellis (approved):

dfellis:
> hmm, what's the difference? It is just moving from one place to another but essentially the same, right?

I don't know for certain, but I think where it was before was before the `<Auth0Provider>` was executed, so maybe `uid` wasn't yet defined? Putting it inside the `useAuth` helper means it'll be run within a component already wrapped by the auth0 provider.

aguillenv:
> > hmm, what's the difference? It is just moving from one place to another but essentially the same, right?
> 
> I don't know for certain, but I think where it was before was before the `<Auth0Provider>` was executed, so maybe `uid` wasn't yet defined? Putting it inside the `useAuth` helper means it'll be run within a component already wrapped by the auth0 provider.

We do not use the `<Auth0Provider>` locally since there's no `config.auth` defined in that environment. The `uid` comes from the env variables we get to define the `AppConfig`so it should be defined at the same time as everything else.

depombo:
I think Alejandro is right. I am not able to pin down what exactly is going on, but this didn't consistently fix some of the events. I think it is okay as-is though. It seems identify still works for a majority of the time

PR #2125, opened 2023-03-07
Why SQL for Infrastructure Management

### LINKED ISSUE #2116: Why SQL is right for infra management

dfellis:
Resolves #2116


site/blog/unlisted/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,94 @@
...
+- By having a standardized protocol to interface with the database, you can:
+  - Perform operations on it via scripts and the [`psql` CLI](https://www.postgresql.org/docs/current/app-psql.html) tool.
+  - Integrate it into your application itself with a [postgres client library](https://node-postgres.com/) for application-driven infrastructure management (like provisioning sharded resources for a client that wants isolation).
+  - Connect to it with a [SQL IDE](https://www.jetbrains.com/datagrip/features/postgresql/) for outage mitigation or inspection in a tabular (Excel-like) view.
+  - Connect it to a [graphical dashboard](https://grafana.com/docs/grafana/latest/datasources/postgres/) for live visualization of your infrastructure.
+  - Connect it to a [low code tool](https://docs.retool.com/docs/postgresql-integration) for rapid response dashboards for your infrastructure with buttons you can tackle known outage vectors near-instantly.
+
+## SQL is the best option for Infrastructure Management
```
depombo, line 77: I'm not sure if `Infra as data` makes more sense than `infra as SQL` in the title and this heading. I'm afraid of people thinking about trying to version control SQL - https://twitter.com/IMTantri/status/1632242097611849729.

It might be worth it to extend on this post that trended a while back https://www.reddit.com/r/kubernetes/comments/nz9fxd/infrastructure_as_software_vs_infrastructure_as/ where basically they make the a distinction between CDK/Pulumi and TF HCL. CDK/Pulumi being infra as software, TF being infra as code and IaSQL being infra as data.

site/blog/unlisted/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+- SQL is an old, irregular language to work with, but it is better known than HCL and we already have a Pulumi in the form of every ORM (like [TypeORM](https://typeorm.io/), [ActiveRecord](https://guides.rubyonrails.org/active_record_basics.html), etc) and QueryBuilder ([LINQ](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/linq/), [Knex](https://knexjs.org/), etc) in whatever programming language you prefer.
+- Peer review of changes is *different* versus IaC. [Migration tools](https://en.wikipedia.org/wiki/Schema_migration) *can* be used to approximate the IaC review process, with the CI system wrapping the migration SQL in an [IaSQL transaction](/docs/transaction) and previewing the operation, but since IaSQL transactions are intentionally *not* actual Postgres transactions, you can simply make the proposed infrastructure changes in the database within an uncommitted IaSQL transaction and allow your peers to inspect/review the database state and call the commit, with the audit log keeping track of the change history for you.
```
depombo, line 82: I would really like to remove this. We have learned the hard way that we don't want to recommend migration tools after attempting to use them ourselves and failing multiple times. I would call out and prioritize the idea of creating a GitHub module that makes read-only side-effect PRs that can be called alongside transactions that push preview output and sql from audit table to a markdown file that teammates can approve. We can link to the task in the post

site/blog/unlisted/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,94 @@
...
+- Database schemas can be overwhelming, and a schema accurately representing *all* of AWS would be dizzying. HCL and other IaC tools let you elide parts of AWS via a module system, so we have implemented our own [module system](/docs/module) on top of Postgres inspired equally by programming language module systems and Linux distro package managers.
+- IaC tools have higher-level modules for even simpler, cookie-cutter use-cases, so we [did the same](/docs/low-level-vs-high-level) in IaSQL.
+- IaSQL is declarative like IaC, so the leaky abstraction can still impact things, but [using transactions](/docs/transaction) you can enqueue changes and preview what actions the engine will take before commiting to it (or rolling it back). Going beyond that and acknowledging that sometimes you need to directly work in the lower abstraction layer, IaSQL exposes optional [raw SDK access](https://iasql.com/docs/modules/aws/aws_sdk/) within the database to allow for
```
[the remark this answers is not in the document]
depombo: IaSQL transactions are only graced on and I think they are one of the cornerstones of why infra as data is the best of the worse. There should be at least one or two sentences on the fact that every infrastructure change should be treated as a transaction that can be rolled back

site/blog/unlisted/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,94 @@
...
+  - Connect it to a [graphical dashboard](https://grafana.com/docs/grafana/latest/datasources/postgres/) for live visualization of your infrastructure.
+  - Connect it to a [low code tool](https://docs.retool.com/docs/postgresql-integration) for rapid response dashboards for your infrastructure with buttons you can tackle known outage vectors near-instantly.
+
+## SQL is the best option for Infrastructure Management
+
+There are negatives to representing your infrastructure as SQL, of course, as there is no perfect solution for all situations, but we believe IaSQL is a superset versus IaC in all but one category.
+
+- SQL is an old, irregular language to work with, but it is better known than HCL and we already have a Pulumi in the form of every ORM (like [TypeORM](https://typeorm.io/), [ActiveRecord](https://guides.rubyonrails.org/active_record_basics.html), etc) and QueryBuilder ([LINQ](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/linq/), [Knex](https://knexjs.org/), etc) in whatever programming language you prefer.
```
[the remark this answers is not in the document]
depombo: also maybe to our tutorials of doing introspection with Prisma and Django

site/blog/unlisted/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+Needing to learn Terraform's HCL language to make infrastructure changes has been cited as a [weakness](https://www.pulumi.com/why-pulumi/) that can be mitigated by wrapping it in the syntaxes more familiar to the user. This *is* a weakness, but we believe that the HCL-to-Statefile transformation being [surjective](https://en.wikipedia.org/wiki/Surjective_function) instead of [bijective](https://en.wikipedia.org/wiki/Bijection) is actually the bigger issue.
+
+<!-- TODO: This section could use actual HCL code, and ideally some way you can trigger a corruption of the statefile vs AWS's actual state -->
+
+## The Objective is Bi- hmmmm....
```
depombo, line 51: I think there is a whole thread to pull on that could probably be its own section, or its own post about how we got to where we are in thinking that declarative is best for infrastructure. TF is basically a cloud-native Puppet. Puppet works in a world where data center resources are homogenous and static meaning most servers are alike, are programmed similarly and the amount of them doesn't change as often. However, in the cloud era there are 250+ services (or developer abstractions) only in AWS built on top of bare VMs which can be used on demand which makes a declarative approach much harder to define.

Als interesting that even then, an infra as data approach was already around https://www.techtarget.com/searchdatacenter/definition/configuration-management-database

site/blog/unlisted/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+IaSQL is [currently in beta](/blog/beta) and can be used locally with [`docker`](https://hub.docker.com/r/iasql/iasql) or via our [SaaS](https://app.iasql.com). Even if you're currently using an IaC like Terraform, IaSQL can re-sync with your cloud on external changes to the infrastructure, so it won't step on the toes of any service you're managing with it, so feel free to try it out on a greenfield project, or just monitor what Terraform is doing to your infrastructure via the audit log. And if you're not using an IaC tool because trying to hand-write HCL representing your current cloud infrastructure is too daunting, definitely give IaSQL a try, when an AWS module is installed, it automatically syncs the state of your infrastructure into the database at that time, and you can begin management immediately.
+
+<!-- TODO: Better closing paragraph, I think. It feels a bit flat. Also, there should probably be images/diagrams in this blog post, related to the examples, screenshots of the various tools I suggested could be connected to IaSQL, etc, maybe even a joke reference to something like https://github.com/centerofci/mathesar -->
```
aguillenv, line 94: maybe summarizing why SQL or treating infra as data is better?

aguillenv:
btw, I really like where this blog post is going! Easy to follow

mtp1376 (approved):
+1 for pointing out the state file in Terraform which exists, but due to low quality it's not exposed to the user.

site/blog/unlisted/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,94 @@
...
+
+Why does Terraform need this JSON-based statefile when it has the much cleaner HCL format that you use as a developer? This is because HCL, as a language that provides affordances to the developer's own intuition on how to structure their representation of their infrastructure, does not have a [canonical form](https://en.wikipedia.org/wiki/Canonical_form) to make these comparisons with. Terraform must first pre-process the HCL into an internal data structure which it can then generate a [diff](https://en.wikipedia.org/wiki/Diff) that can then be used to determine the operations to perform. The JSON-based statefile is likely a very close representation of this internal data structure, and manual editing of it is highly discouraged because the details of it likely vary from release to release with an internal migration mechanism performed on upgrade of the Terraform CLI.
```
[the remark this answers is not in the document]
mtp1376: Yes. But for that purpose, TF supports using S3 as the state file backend and DynamoDB to hold the lock to prevent others from obtaining it while a user is modifying the infrastructure:
https://angelo-malatacca83.medium.com/aws-terraform-s3-and-dynamodb-backend-3b28431a76c1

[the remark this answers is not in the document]
dfellis: I'm not sure whether to be impressed or appalled with that solution. ;)

site/blog/unlisted/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,94 @@
...
+- IaC tools have higher-level modules for even simpler, cookie-cutter use-cases, so we [did the same](/docs/low-level-vs-high-level) in IaSQL.
+- IaSQL is declarative like IaC, so the leaky abstraction can still impact things, but [using transactions](/docs/transaction) you can enqueue changes and preview what actions the engine will take before commiting to it (or rolling it back). Going beyond that and acknowledging that sometimes you need to directly work in the lower abstraction layer, IaSQL exposes optional [raw SDK access](https://iasql.com/docs/modules/aws/aws_sdk/) within the database to allow for
+- As IaSQL is able to both push changes from the database to the cloud and pull changes from the cloud to the database versus the one-way push from HCL that Terraform provides, you can make changes directly in the AWS console, then inspect the audit log to see what SQL records are necessary and automatically generate all of the rest of the resources you need.
```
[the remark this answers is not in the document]
dfellis: Well, I feel like this is oriented more to the non-power users that don't quite have the hang of IaSQL tables and relations, yet. ;)

The point is that IaSQL can be used cooperatively with your existing tools, and has more flexibility in what flows are possible.

I think this has convinced me that these bullet points do need to be broken into paragraphs of their own, even if it makes this already long blog post longer.

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -168,6 +168,18 @@ resource "aws_instance" "login" {
 }
 ```
 
+This example still requires the following lines from the imperative example to get the public IP address of the new EC2 instance and log into it:
+
+```bash
+PUBLIC_IP=$(aws ec2 describe-instances \
```
mtp1376, line 174: Did you consider making this an [HCL output](https://developer.hashicorp.com/terraform/language/values/outputs) and [accessing it via bash](https://stackoverflow.com/a/69260021/2594912)?

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -102,8 +102,69 @@ Infrastructure management *generally* is not that straightforward. Changes to in
...
+        # we don't accidentally re-create the file between 'plan' and 'apply'
+        yes '' | ssh-keygen -t ed25519 -m PEM -f login.pem
+      fi
+      chmod 600 login.pem
+      # The output of this script needs to be JSON formatted. This is a bit wonky
+      # but building it over time via 'jq' operators is actually harder to follow
+      echo {
+      echo   '"access_key_id": "'$(aws configure get aws_access_key_id)'",'
+      echo   '"secret_access_key": "'$(aws configure get aws_secret_access_key)'",'
+      echo   '"public_key": "'$(cat login.pem.pub)'"'
```
mtp1376, line 131: I think Terraform will use the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables. Also, for the key we can use [`local_file`](https://registry.terraform.io/providers/hashicorp/local/latest/docs/data-sources/file) instead?

mtp1376:
Looks very comprehensive. It added to my own knowledge.

site/blog/updates/why-sql-for-infrastructure.mdx
dfellis: So, the terraform code was the longest of the three, so I thought adding that and then still having a bash script afterwards would unfairly paint Terraform even worse, but I suppose an argument could be made that I'm unfairly painting Terraform as being unable to get that public IP address to use in the bash script.

I *know* the Terraform code is going to draw rotten tomatoes at us either way, but let me think about which is "worse." (And go with that one? :zany_face:)

site/blog/updates/why-sql-for-infrastructure.mdx
dfellis: Yes, but I didn't like the not-self-contained nature of that approach versus this one for both situations. It requires us to invoke `terraform plan` like `AWS_ACCESS_KEY_ID=$(aws configure get aws_access_key_id) AWS_SECRET_ACCESS_KEY=$(aws configure get aws_secret_access_key) terraform plan` if we don't want to make the assumption that those environment variables were already set.

Also, I hate environment variables because they are common footguns when trying to reproduce what some piece of code does, and I prefer to minimize them as much as possible.

dfellis: I can add a comment into the Terraform code that the shell script isn't necessary if you set the appropriate environment variables outside and create the ssh key ahead of time and just reference it here, but those would be manual steps you have to ensure you do before running the rest of the Terraform code which makes actually executing the example more difficult.

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+
+"But," you say, "programming languages with solid type systems, like Rust, can also do that for you." That is true, and this was not an attempt to disparage them, merely to show that you aren't losing anything by using SQL as the representation. There are, however, ways SQL goes above and beyond. By defining [unique, not-null, and check constraints](https://www.postgresql.org/docs/current/ddl-constraints.html) on top of a [rich type system](https://www.postgresql.org/docs/current/datatype.html) the database can prevent you from inserting faulty data to a greater degree than even [Rust](https://www.rust-lang.org), as the constraints can be against the live state of your infrastructure itself. Instead of code needing to deal with unexpected values further down the line, the faulty data can be prevented from ever being inserted. Rust's compiler making sure that you handle every possible branch is an awesome thing, but SQL has the power to make more of these branches impossible in the first place.
```
aguillenv, line 334: nit:
```suggestion
"But," you say, "programming languages with solid type systems, like [Rust](https://www.rust-lang.org), can also do that for you." That is true, and this was not an attempt to disparage them, merely to show that you aren't losing anything by using SQL as the representation. There are, however, ways SQL goes above and beyond. By defining [unique, not-null, and check constraints](https://www.postgresql.org/docs/current/ddl-constraints.html) on top of a [rich type system](https://www.postgresql.org/docs/current/datatype.html) the database can prevent you from inserting faulty data to a greater degree than even Rust, as the constraints can be against the live state of your infrastructure itself. Instead of code needing to deal with unexpected values further down the line, the faulty data can be prevented from ever being inserted. Rust's compiler making sure that you handle every possible branch is an awesome thing, but SQL has the power to make more of these branches impossible in the first place.
```

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+VALUES ('resolve:ssm:/aws/service/canonical/ubuntu/server/20.04/stable/current/amd64/hvm/ebs-gp2/ami-id', 't3.small', 'login', 'running', '{"name":"login"}'::jsonb, 'f', (SELECT region FROM aws_regions WHERE region = 'us-west-2'), (SELECT id FROM subnet WHERE subnet_id = 'subnet-06140fd708a495450' AND region = (SELECT region FROM aws_regions WHERE region = 'us-west-2')));
+
+
+INSERT INTO instance_security_groups (instance_id, security_group_id)
+VALUES ((SELECT id FROM instance WHERE instance_id = NULL AND region = (SELECT region FROM aws_regions WHERE region = 'us-west-2')), (SELECT id FROM security_group WHERE group_id = NULL AND region = (SELECT region FROM aws_regions WHERE region = 'us-west-2')));
+```
+
+which in this case is very similar to (though a bit more verbose than) our original SQL statements, but if you were going back and forth on what exact changes you're going to make, it can be a good summary.
+
+If we're good with this, we can `SELECT iasql_apply()` to push this to our cloud account and then update our `ssh_credentials` record with the public IP address and check the connection:
```
aguillenv, line 455:
```suggestion
If we're good with this, we can `SELECT iasql_commit()` to push this to our cloud account and then update our `ssh_credentials` record with the public IP address and check the connection:
```

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+
+*By decree of the king, a restaurant shall be built!*
+
+:::
+
+In declarative programming you have some initial state and in comparatively dense and high-level declarative code define the desired state. In many use-cases (like web browsers and sometimes for restaurant-building contractors) the initial state is "blank" and the engine that transforms the declarative code into imperative operations can often be a relatively straightforward [interpreter](https://en.wikipedia.org/wiki/Interpreter_%28computing%29) walking the [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree) of the declarative code.
+
+Infrastructure management *generally* is not that straightforward. Changes to infrastructure require in-place mutations of the current state. A full [blue/green deployment](https://en.wikipedia.org/wiki/Blue-green_deployment) of the entirety of your production infrastructure is very expensive, requiring all resource costs to be doubled, and deployments to require some amount of downtime for users to swap over and state to be resynchronized. Particularly for databases this approach is essentially impossible as there is too much state to swap.
```
depombo, line 100:
```suggestion
Infrastructure management *generally* is not that straightforward. Changes to infrastructure require in-place mutations of the current state, or a full [blue/green deployment](https://en.wikipedia.org/wiki/Blue-green_deployment) of the entirety of your production infrastructure is very expensive, requiring all resource costs to be doubled, and deployments to require some amount of downtime for users to swap over and state to be resynchronized. Particularly for databases this approach is essentially impossible as there is too much state to swap which is why database migrations are a nightmare.
```

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+
+Needing to learn Terraform's HCL language to make infrastructure changes has been cited as a [weakness](https://www.pulumi.com/why-pulumi/) that can be mitigated by wrapping it in the syntaxes more familiar to the user like [Terraform CDK](https://developer.hashicorp.com/terraform/cdktf), [Pulumi](https://www.pulumi.com), or [AWS CDK](https://docs.aws.amazon.com/cdk/v2/guide/home.html). This *is* a weakness, but we believe that the HCL-to-Statefile transformation being [surjective](https://en.wikipedia.org/wiki/Surjective_function) instead of [bijective](https://en.wikipedia.org/wiki/Bijection) is actually the bigger issue.
+
```
depombo, line 307: let's include a surjective diagram analogous to the bijective one in the next section

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+The answer is ~~[water vapor](https://en.wikipedia.org/wiki/Cloud)~~ [Entities](https://en.wikipedia.org/wiki/Entity#In_computer_science). Each cloud resource has some unique identifier and a set of properties that configure it. Some of those properties can be relations or dependencies on other cloud resources. The ordering of the entities does not matter, so long as the dependencies are explicitly defined to make determination of the order of operations possible. You may have noticed that this sounds a lot like [Objects](https://en.wikipedia.org/wiki/Object_%28computer_science%29) in [Object-Oriented Programming](https://en.wikipedia.org/wiki/Object-oriented_programming). It is very much the same, with the only difference being that the unique identifier *must* be unique, so if we considered an array of pointers to these objects as the entity IDs, multiple pointers to the same object wouldn't be allowed.
```
depombo, line 322:
```suggestion
The best answer in most cases is ~~[water vapor](https://en.wikipedia.org/wiki/Cloud)~~ [Entities](https://en.wikipedia.org/wiki/Entity#In_computer_science). Each cloud resource has some unique identifier and a set of properties that configure it. Some of those properties can be relations or dependencies on other cloud resources. The ordering of the entities does not matter, so long as the dependencies are explicitly defined to make determination of the order of operations possible. You may have noticed that this sounds a lot like [Objects](https://en.wikipedia.org/wiki/Object_%28computer_science%29) in [Object-Oriented Programming](https://en.wikipedia.org/wiki/Object-oriented_programming). It is very much the same, with the only difference being that the unique identifier *must* be unique, so if we considered an array of pointers to these objects as the entity IDs, multiple pointers to the same object wouldn't be allowed.
```

make it less sales-y?

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+
+SQL Database [Snapshotting](https://en.wikipedia.org/wiki/Snapshot_%28computer_storage%29) could be used to quickly revert all infrastructure changes for infrastructure without hidden state (eg, the snapshot of your load balancers' prior state would successfully restore the load balancer configuration, but a dropped database restored via a snapshot would not restore the data within, only that you had such a database in the first place and you would still need to manually re-load a backup snapshot of those, as well).
+
+Database engines also represent a standardized protocol to interface with your data. This standard opens up a world of possibilities that can encompass the imperative and declarative infrastructure management styles and go beyond them. You can:
+  - Perform operations on your database via scripts and the [`psql` CLI](https://www.postgresql.org/docs/current/app-psql.html) tool, using whatever mix of declarative and imperative styles you prefer, mutating various tables declaratively and calling [`AWS SDK functions directly`](/docs/modules/aws/aws_sdk/?view=code-examples) where and when you see fit.
```
depombo, line 347: I would mention that declarative is possible, but more cumbersome in SQL

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+
+Database schemas can be overwhelming, and a schema accurately representing *all* of AWS would be dizzying. HCL and other IaC tools let you elide parts of AWS via a module system, so we have implemented our own [module system](/docs/module) on top of Postgres inspired equally by programming language module systems and Linux distro package managers. This module system takes the place the usual database [schema migration system](https://en.wikipedia.org/wiki/Schema_migration) for the management of the tables in the database, which may feel unusual at first, but we believe it is superior (for this use-case).
+
+Even with IaC tools letting you ignore details you don't specifically need, AWS and other clouds expose many details in multiple layers for you to control when often you just need the same subset of functionality over-and-over, so these IaC tools have higher-level modules for these cookie-cutter use-cases. For IaSQL we [did the same](/docs/low-level-vs-high-level), though in a way that coexists with the lower-level modules so you can still query those details later, if you like.
```
depombo, line 361:
```suggestion
Even with IaC tools letting you ignore details you don't specifically need, AWS and other clouds expose many details in multiple layers for you to control when often you just need the same subset of functionality over-and-over, so these IaC tools have higher-level modules for these cookie-cutter use-cases. For IaSQL we [did the same](/docs/low-level-vs-high-level), though in a way that coexists with the lower-level modules so you can still query those details later, if you like. [Here](/blog/ecs-simplified) is an example of a high-level module that greatly simplifies ECS fargate. 
```

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+
+SQL immediately stands out here because it was designed for making [relational algebra](https://en.wikipedia.org/wiki/Relational_algebra), the other side of the [Entity-Relationship model](https://en.wikipedia.org/wiki/Entity%E2%80%93relationship_model), accessible. There are likely more people who know SQL than any programming language (for IaC) or data format you could choose to represent your cloud infrastructure. Many non-programmers know it, as well, such as data scientists, business analysts, accountants, etc, and there is an entire ecosystem of useful functionality you gain "for free" by using a SQL representation than just about any format out there. Furthermore, a powerful SQL engine like the open source [PostgreSQL](https://postgresql.org) provides tools capable of elevating the infrastructure management *experience* beyond what IaC can do today.
+
+### The Advantages of SQL Databases for Infrastructure Management
```
depombo, line 330:
```suggestion
## The Advantages of SQL Databases for Infrastructure Management
```

depombo: ?

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+
+Data, on the other hand, is much easier to normalize in such a way, and relational data models, as found in SQL databases, can naturally model the relationships between cloud services, making it possible to produce a bijective representation. IaSQL takes advantage of this to make a declarative model based on tables and foreign keys that can automatically handle changes done outside of this declarative model, either by dropping down into its own imperative mode or by using any other tooling to manage your infrastructure, giving it a flexibility and resiliency that IaC tools don't have.
+
+On top of this foundation, tooling to replicate or substitute for existing IaC tooling has been added, along with all of the past ~50 years of database tooling and expertise that you can take advantage of, providing a superset of functionality that makes existing use-cases easier and new use-cases possible.
+
+IaSQL is [currently in beta](/blog/beta) and can be used locally with [`docker`](https://hub.docker.com/r/iasql/iasql) or via our [SaaS](https://app.iasql.com). We're proud of what we have built, and can't wait to hear from you on feature requests and bug reports.
```
depombo, line 482:
```suggestion
IaSQL is [currently in beta](/blog/beta) and can be used locally with [`docker`](https://hub.docker.com/r/iasql/iasql) or via our [SaaS](/hosted). We're proud of what we have built, and can't wait to hear from you on feature requests and bug reports.
```

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+  INNER JOIN instance i ON im.instance_id = i.instance_id
+  WHERE i.tags ->> 'name' = 'login'
+  LIMIT 1
+)
+UPDATE ssh_credentials SET hostname = h.hostname FROM ssh_credentials, h WHERE hostname = 'tbd';
+
+SELECT * FROM ssh_exec('login', 'uname -a');
+```
+
+## SQL is the best option for Infrastructure Management
```
depombo, line 470:
```suggestion
## SQL as the best of the worst option for Infrastructure Management
```

a bit nihilistic or cynical, but keeps the tone from the beginning?

depombo (approved):
this looks great!

site/blog/updates/why-sql-for-infrastructure.mdx
```diff
@@ -0,0 +1,482 @@
...
+
+1. Design restaurant.
+2. Flatten the ground.
+3. Prop up pieces of wood next to each other.
+4. Bang nails into wood.
+5. Is restaurant finished? No, go to 3.
+
+:::
+
+For your cloud infrastructure, this is similar to using the [AWS SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/) and directly calling the various methods, checking their results and eventually arriving at the desired infrastructure. It is *also* like following a step-by-step guide clicking through the various AWS console UI elements to set up your infrastructure, like in [this guide](https://towardsdatascience.com/deploy-your-python-app-with-aws-fargate-tutorial-7a48535da586). The latter may not be automated, to you, but to your company executives it is as they don't care how the infrastructure building was accomplished as long as it was done quickly, cheaply, and reliably, and you know they always want all three. 😉
```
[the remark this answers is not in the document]
dfellis: Do you think we can include it without distracting? I never used ansible, so I wouldn't want to undersell/underdescribe it.

site/blog/updates/why-sql-for-infrastructure.mdx
dfellis: I don't know if I want to say they're a nightmare? I didn't really have much issue with them at CarDash. It gets hard when there are concurrent migration efforts happening by different engineers, which is what I think caused issues back in the early days at Uber.

I'd probably tone it down a bit to:

```suggestion
Infrastructure management *generally* is not that straightforward. Changes to infrastructure require in-place mutations of the current state or a full [blue/green deployment](https://en.wikipedia.org/wiki/Blue-green_deployment) of the entirety of your production infrastructure is very expensive, requiring all resource costs to be doubled, and deployments to require some amount of downtime for users to swap over and state to be resynchronized. Particularly for databases this approach is essentially impossible as there is too much state to swap, which is what makes database migrations tricky when dropping data.
```

site/blog/updates/why-sql-for-infrastructure.mdx
dfellis: > using whatever mix of declarative and imperative styles you prefer

This doesn't cover that?

dfellis: It's also from the perspective of "the DB records themselves are the declarative 'code'" not "you can make idempotent SQL queries to make the query itself quasi-declarative" which I didn't touch on.

site/blog/updates/why-sql-for-infrastructure.mdx
dfellis: Why this suggestion? I considered it a sub-section of the section it's currently in.

site/blog/updates/why-sql-for-infrastructure.mdx
dfellis: "the best of the worst" reads a bit weird to me. I think this would work a bit better?

```suggestion
## SQL is the least bad option for Infrastructure Management
```

site/blog/updates/why-sql-for-infrastructure.mdx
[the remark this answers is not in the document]
depombo: I think it would be too much. That can come in another blog post about the `ssh-*` modules and we can point back to this one

site/blog/updates/why-sql-for-infrastructure.mdx
depombo: I see. I guess I don't think about it in the first way that you describe it 🤔

site/blog/updates/why-sql-for-infrastructure.mdx
depombo: that is better!

site/blog/updates/why-sql-for-infrastructure.mdx
dfellis: :::note Surjective Function *f* Diagram

Surjective Diagram

[*From Wikipedia*](https://en.wikipedia.org/wiki/Surjective_function#/media/File:Surjection.svg) A function *f* is surjective if all of its inputs *X* map to all outputs *Y*, but overlaps can occur.

This is like multiple Terraform HCL representations all mapping to the exact same cloud infrastructure.

depombo (approved):

PR #2126, opened 2023-03-07
fix: modify threshold datatype to be numeric

yrobla:
the threshold needs to be able to accept decimal numbers


src/modules/aws_cloudwatch/entity/metric_alarm.ts
```diff
@@ -316,8 +317,9 @@ export class MetricAlarm {
    * @public
    * The value against which the specified statistic is compared.
    */
-  @Column({
+  @Column('numeric', {
```
dfellis, line 320: So according to the documentation this field is [basically a `double](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-cloudwatch/interfaces/putmetricalarmcommandinput.html#threshold)

Instead of `numeric` and then converted it internally into a double, why not use the [`double precision`](https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-FLOAT) type in Postgres directly and avoid the conversion noise?

src/modules/aws_cloudwatch/entity/metric_alarm.ts
```diff
@@ -316,8 +317,9 @@ export class MetricAlarm {
    * @public
    * The value against which the specified statistic is compared.
    */
-  @Column({
+  @Column('double precision', {
     nullable: true,
+    transformer: new ColumnNumericTransformer(),
```
dfellis, line 322: You can drop this now

```suggestion
```

src/modules/aws_cloudwatch/entity/metric_alarm.ts
```diff
@@ -2,6 +2,7 @@ import { Column, Entity, Index, JoinColumn, ManyToOne, PrimaryGeneratedColumn }
 
 import { cloudId } from '../../../services/cloud-id';
 import { AwsRegions } from '../../aws_account/entity';
+import { ColumnNumericTransformer } from '../../transformers';
```
dfellis, line 5:
```suggestion
```

src/modules/transformers.ts
```diff
@@ -0,0 +1,10 @@
+// transformer to convert string to decimal
```
dfellis, line 1: This whole file can be deleted now

dfellis (approved):

test/modules/aws-cloudwatch-integration.ts
```diff
@@ -451,7 +451,7 @@ describe('AwsCloudwatch Integration Testing', () => {
     query(
       `
       INSERT INTO metric_alarm (alarm_name, alarm_description, actions_enabled, comparison_operator, datapoints_to_alarm, dimensions, metric_name, namespace, period, statistic, threshold, evaluation_periods)
-      VALUES ('${alarmName}', 'Metric alarm description', true, 'GreaterThanThreshold', 1, '[{"Name": "InstanceId", "Value": "test"}]', 'CPUUtilization', 'AWS/EC2', 60, 'Average', 10, 1);
+      VALUES ('${alarmName}', 'Metric alarm description', true, 'GreaterThanThreshold', 1, '[{"Name": "InstanceId", "Value": "test"}]', 'CPUUtilization', 'AWS/EC2', 60, 'Average', 10.5, 1);
```
aguillenv, line 454: would be good to add a test with the number that failed

depombo: +1

src/modules/transformers.ts
aguillenv: @yrobla can we drop it? or is still necessary?

dfellis (approved):

PR #2128, opened 2023-03-07
Force run on install and uninstall

### LINKED ISSUE #2119: Auto-run install queries

aguillenv:
Resolves #2119

### DIFF
--- a/dashboard/src/components/providers/AppProvider.tsx
+++ b/dashboard/src/components/providers/AppProvider.tsx
@@ -166,7 +166,13 @@ const reducer = (state: AppState, payload: Payload): AppState => {
       tabsCopy[tabIdx].queryRes = queryRes;
       if (runSqlUpdatedDbs !== null && runSqlUpdatedDbs !== undefined) {
         const current = runSqlUpdatedDbs.find((d: any) => d.alias === state.selectedDb.alias);
-        return { ...state, databases: runSqlUpdatedDbs, selectedDb: current, editorTabs: tabsCopy };
+        return {
+          ...state,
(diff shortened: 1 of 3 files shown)


dashboard/src/components/providers/AppProvider.tsx
```diff
@@ -166,7 +166,13 @@ const reducer = (state: AppState, payload: Payload): AppState => {
       tabsCopy[tabIdx].queryRes = queryRes;
       if (runSqlUpdatedDbs !== null && runSqlUpdatedDbs !== undefined) {
         const current = runSqlUpdatedDbs.find((d: any) => d.alias === state.selectedDb.alias);
-        return { ...state, databases: runSqlUpdatedDbs, selectedDb: current, editorTabs: tabsCopy };
+        return {
+          ...state,
+          databases: runSqlUpdatedDbs,
+          selectedDb: current,
+          editorTabs: tabsCopy,
+          forceRun: false,
```
dfellis, line 174: Looking at this fix, would it be possible to make the return type a dependent type of the `ActionType` value so we can force all returns to include the necessary values at compile time?

Not a blocker on this PR since it'll probably take a bit of thought, but a suggestion of some improvement to this centrally important reducer function.

dfellis (approved):

dashboard/src/components/providers/AppProvider.tsx
aguillenv: Indeed, we need to work on types in general for inputs and outputs in the dashboard. I think we do not have a task for it, I'll create it as tech debt

PR #2135, opened 2023-03-07
link to guide before /connect

depombo:


dfellis:
I think we should just disable this linter warning, if possible: https://github.com/iasql/iasql/actions/runs/4357829150/jobs/7617621416#step:5:17

depombo:
> I think we should just disable this linter warning, if possible: https://github.com/iasql/iasql/actions/runs/4357829150/jobs/7617621416#step:5:17

makes sense. I just added it for now

PR #2160, opened 2023-03-08
Add support for temporary session tokens

### LINKED ISSUE #2141: Allow supporting temporary STS credentials
Per the request of @md5, some organizations discourage the use of IAM users which are currently required by IaSQL to generate credentials and connect an AWS account. This task gives initial support for an STS session token that can be provided as an additional, optional field via the dashboard and `aws_account` table. Once the token expires, the credentials will become invalid and/or the record deleted from the table. There still needs to be some more investigation, but this is roughly how it would work supposedly https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html#using-temp-creds-sdk.

The start of the discord conversation: https://discord.com/channels/920684574231646229/920684574835609622/1082770524620460052

dfellis:
Resolves #2141

Tested with credentials generated with `aws sts get-session-token`. It also provides an expiration field but that is not necessary to use the temporary token.


dfellis:
## Manual testing

We decided to do manual testing for the first pass and worry about automating the test later.

### 1. Temporary credential generation

Here is the generation of the temporary credentials. I looked and there really is no way to do this via the AWS console. By default they only last 1 hour!

Screenshot from 2023-03-08 14-52-05

### 2. New Optional Session Token Field

I couldn't think of any visual way to indicate that the session token is optional, so I labeled it "(Optional)".

Screenshot from 2023-03-08 14-50-02

And here you can see that it doesn't block the Next button once the first two credential fields are provided.

Screenshot from 2023-03-08 14-50-23

Still a bit worried that this feature will confuse newbies, but not sure what else to do here.

### 3. Session Token *is* defined in the DB

Self explanatory

Screenshot from 2023-03-08 14-49-45

### 4. Can still install modules and get the expected results out of it

Everything is working just fine with these temporary credentials :)

Screenshot from 2023-03-08 14-49-32

dashboard/src/components/providers/AppProvider.tsx
```diff
@@ -402,18 +394,27 @@ const middlewareReducer = async (
...
+            VALUES ('${awsAccessKeyId}', '${awsSecretAccessKey}');
           `,
           );
         }
+        await DbActions.run(
+          token,
+          backendUrl,
+          alias,
+          `
+          SELECT * FROM iasql_begin();
```
depombo, line 407: shouldn't the begin be before?

depombo (approved):

depombo:
it might be good to mention it in the /docs/aws part of the documentation

dashboard/src/components/providers/AppProvider.tsx
dfellis: It actually doesn't matter on that front, just forcing `commit` to immediately trigger the loading of the AWS region data. The `aws_account` table's mapper isn't actually connected to anything on the cloud side. I think when I first wrote it I originally had the two select statements in the same request and someone must have separated them out. It was also put here because of the original conditional for before and after transactions (non-transactions was `iasql_apply`, with transactions was `iasql_begin .. iasql_commit`) and it reduced the number of conditionals needed from two to one this way.

depombo (approved):

aguillenv (approved):

mtp1376 (commented):
Looks good!

PR #2165, opened 2023-03-09
Disconnect on error creating new db

### LINKED ISSUE #2155: Immediately disconnect a database if connect fails from bad creds
This can lead cause confusion about the onboarding actually working

aguillenv:
Resolves #2155

### DIFF
--- a/dashboard/src/components/providers/AppProvider.tsx
+++ b/dashboard/src/components/providers/AppProvider.tsx
@@ -424,6 +424,13 @@ const middlewareReducer = async (
         `,
         );
       } catch (e: any) {
+        middlewareReducer(config, dispatch, {
+          action: ActionType.DisconnectDb,
+          token,
+          data: {
+            dbAlias: alias,
+          },
+        });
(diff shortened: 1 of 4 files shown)


dfellis (approved):

dashboard/src/components/providers/AppProvider.tsx
```diff
@@ -424,6 +424,13 @@ const middlewareReducer = async (
         `,
         );
       } catch (e: any) {
+        middlewareReducer(config, dispatch, {
```
depombo, line 427: why is this not a dispatch call?

depombo (approved):

dashboard/src/components/providers/AppProvider.tsx
aguillenv: In the `AppProvider` we have a `middleware` where we do the complex logic + API calls and then that's transferred to the `reducer` where we just make the transformations needed to the state and return. If we are inside the `middleware` a `dispatch` call will continue directly to the `reducer` but in this case we want to pass though the `middleware` again to make the API call to actually disconnect.

PR #2166, opened 2023-03-09
Handle connection interruptions

### LINKED ISSUE #2164: Disconnect error
My main suspicion is a problem with next.js `run` endpoint. While a query is executing behind scenes, then the disconnect occurs and causes the failure for the dashboard due to an unhandled exception. This is because `disconnect` kills all open connections.

aguillenv:
Resolves #2164

### DIFF
--- a/dashboard/src/pages/api/run.ts
+++ b/dashboard/src/pages/api/run.ts
@@ -6,6 +6,7 @@ import format from 'pg-format';
 import { parse, deparse } from 'pgsql-parser';
 import { DataSource } from 'typeorm';
 import { PostgresConnectionOptions } from 'typeorm/driver/postgres/PostgresConnectionOptions';
+import { v4 as uuidv4 } from 'uuid';
 
 import { throwError } from '@/config/config';
 import config from '@/server-config';
@@ -116,16 +117,21 @@ async function runSql(
     });
     // Based on https://node-postgres.com/apis/client#error
     connTemp.on('error', e => {
+      const errorId = uuidv4();
       logger.error?.('Connection error', {
         app: 'run',
         env: process.env.IASQL_ENV,
         meta: {
+          errorId,
           sql,
           error: e.message,
           stack: e.stack,
         },
       });
-      res.status(500).json({ error: `Connection interruption while executing query ${sql}` });
+      res.status(500).json({
+        error: `Connection interruption while executing query ${sql}
+Please provide this error ID when reporting this bug: ${errorId}`,
+      });
     });
     await connTemp.connect();
     const deparsedStmt = deparse(stmt);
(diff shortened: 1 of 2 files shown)


dashboard/src/pages/api/run.ts
```diff
@@ -108,8 +108,18 @@ async function runSql(sql: string, dbAlias: string, username: string, password:
       host: baseConnConfig.host,
       ssl: baseConnConfig.extra.ssl,
     });
+    // Based on https://node-postgres.com/apis/client#error
+    connTemp.on('error', e => {
```
dfellis, line 112: Should we also pass the `res` object through to here and if this case is hit return an error to the front-end?

dashboard/src/pages/api/run.ts
```diff
@@ -108,8 +108,18 @@ async function runSql(sql: string, dbAlias: string, username: string, password:
...
+        app: 'run',
+        env: process.env.IASQL_ENV,
+        meta: {
+          error: e.message,
+          stack: e.stack,
+        },
+      });
+    });
     await connTemp.connect();
-    // await connTemp.query(dbMan.setPostgresRoleQuery(database));
```
depombo, line 112: why is the commented out? Isn't just catching the error enough

aguillenv: this line was commented, I just removed assuming is not necessary

dashboard/src/pages/api/run.ts
aguillenv: I think we cannot since the listener is defined already in the `pg` lib. For the cases when this is hit due to `disconnect` killing open connections, it is ok not to do anything since that's exactly what disconnect is doing, cancel all ongoing paths.

dfellis: hmm, so the handler is being defined within the `runSql` function call so each call will get a new handler. This means that we could make `res` another argument to `runSql` and then use it here as part of the closure.

Whether that's actually necessary is something else; in this particular case, no, but if there's an error in the future due to some other issue the error could be useful.

aguillenv: ohh I see your point, yes that could be useful. I could give it a try in this PR

dashboard/src/pages/api/run.ts
depombo: oh my bad, got it

depombo (approved):

dashboard/src/pages/api/run.ts
```diff
@@ -114,10 +120,12 @@ async function runSql(sql: string, dbAlias: string, username: string, password:
         app: 'run',
         env: process.env.IASQL_ENV,
         meta: {
+          sql,
           error: e.message,
           stack: e.stack,
         },
       });
+      res.status(500).json({ error: `Connection interruption while executing query ${sql}` });
```
dfellis, line 128: If we're going to hide the actual message it may make sense to generate a UUID and include it in both the `logger.error` and this JSON blob so if we trigger the error while running locally we can check the mezmo logs for that exact UUID and get the details.

dfellis (approved):

PR #2169, opened 2023-03-09
feat: enable multi-az cluster support

### LINKED ISSUE #2118: Support clusters in `aws_rds`
This will unlock Aurora

yrobla:
Closes #2118


aguillenv:
Apart from the tags, looks good to me!

aguillenv (approved):

src/modules/aws_rds/entity/index.ts
```diff
@@ -1,2 +1,4 @@
+export * from './db_cluster';
+export * from './db_subnet_group';
```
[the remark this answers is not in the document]
aguillenv: talked offline, will be added as a follow-up!

src/modules/aws_rds/mappers/db_cluster.ts
```diff
@@ -0,0 +1,346 @@
...
+  entity = DBCluster;
+  equals = (a: DBCluster, b: DBCluster) =>
+    Object.is(a.allocatedStorage, b.allocatedStorage) &&
+    Object.is(a.iops, b.iops) &&
+    Object.is(a.backupRetentionPeriod, b.backupRetentionPeriod) &&
+    Object.is(a.dbClusterInstanceClass, b.dbClusterInstanceClass) &&
+    Object.is(a.deletionProtection, b.deletionProtection) &&
+    Object.is(a.engine, b.engine) &&
+    Object.is(a.engineVersion, b.engineVersion) &&
+    !a.masterUserPassword && // Special case, if master password defined, will update the password
```
dfellis, line 31: Note for the future: I know we do this in normal RDS instances, too, but with RPCs, we may want to pull this out and have an `update_cluster_password` RPC function here. Not blocking.

src/modules/aws_rds/entity/rds.ts
dfellis, line 1: Perhaps we should rename this entity to `DBInstance` to better match AWS terminology and make it clearer with respect to the new `DBCluster` entity?

dfellis (approved):
Two non-blocking comments.

src/modules/aws_rds/mappers/db_cluster.ts
yrobla: let me create an issue for it

src/modules/aws_rds/entity/rds.ts
yrobla: makes sense to me as well, and it will actually be clear because dbinstance is the name that the entity receives in aws

PR #2170, opened 2023-03-09
Remove duplicated modules from the list on install

### LINKED ISSUE #2105: Table already exists error fails query installing multiple modules and one is already exists
The query fails, but it should probably skip it and carry on with the rest of the modules to install

aguillenv:
Resolves #2105 

Before:
image

After:
image

### DIFF
--- a/test/basic_integration/install.ts
+++ b/test/basic_integration/install.ts
@@ -1,10 +1,19 @@
 import * as iasql from '../../src/services/iasql';
-import { runQuery, finish, execComposeUp, execComposeDown, runInstall, runUninstallAll } from '../helpers';
+import {
+  runQuery,
+  finish,
+  execComposeUp,
+  execComposeDown,
+  runInstall,
+  runUninstallAll,
+  runUninstall,
+} from '../helpers';
 
 const dbAlias = 'installtest';
 
 const query = runQuery.bind(null, dbAlias);
 const install = runInstall.bind(null, dbAlias);
+const uninstall = runUninstall.bind(null, dbAlias);
 const uninstallAll = runUninstallAll.bind(null, dbAlias);
 const uid = '12345';
 const email = 'test@example.com';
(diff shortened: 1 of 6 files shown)


dfellis (approved):

dfellis:
The new test is failing here: https://github.com/iasql/iasql/actions/runs/4377646129/jobs/7661411984#step:6:6743

test/basic_integration/install.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  it('reinstalls existing modules', install(['aws_ec2', 'aws_vpc']));
+
+  it('uninstalls all', uninstallAll());
+
+  it('installs complex and dependent sql module', install(['aws_ecs_fargate', 'aws_ecs_simplified']));
+
+  it('fails having an install with a typo', done =>
+    void query(`
+    SELECT * FROM iasql_install(['aws_ec3'])
```
dfellis, line 66:
```suggestion
    SELECT * FROM iasql_install('aws_ec3')
```

dfellis (approved):

test/basic_integration/install.ts
```diff
@@ -0,0 +1,79 @@
...
+    ),
+  );
+
+  it('installs a single module', install(['aws_lambda']));
+
+  it('installs another module that share dependencies with the previous one', install(['aws_ec2']));
+
+  it('uninstalls all', uninstallAll());
+
+  it('installs having a duplication in the list', install(['aws_ecr', 'aws_ecr']));
```
depombo, line 54: would uninstalling duplicates blow up?

mtp1376: Same question as @depombo

mtp1376 (commented):
It's :stamp: after the test with duplicate module name on `iasql_uninstall` is added.

test/basic_integration/install.ts
aguillenv: good catch!

dfellis (approved):

depombo (approved):

PR #2171, opened 2023-03-09
replicate changes across envs how-to guide + minor naming improvements

depombo:


site/docs/how-to/envs.md
```diff
@@ -0,0 +1,37 @@
...
+--- more infrastructure changes
+
+--- calls iasql_preview to see the result in the cloud
+select * from iasql_preview();
+
+--- calls iasql_commit should create
+select * from iasql_commit();
+```
+
+Once the transaction is committed successfully, IaSQL automatically rolls back failed transactions, and the desired changes are committed into the DB for one environment, we have a PG function that looks at the audit log and generates the SQL representing changes done from a given point in time.
```
mtp1376: Here the first sentence says "Once the transaction is committed successfully, IaSQL automatically rolls back failed **transactions**", which seems like the second transaction is something else than the first. I think this sentence can be written more clearly to show the unsuccessful transactions are automatically rolled back.

mtp1376: Do we have partial apply? Or a transaction is either 1 (applied) or 0 (rolled back)?

site/docs/how-to/envs.md
```diff
@@ -0,0 +1,37 @@
...
+
+```sql
+-- gets SQL from the audit log from a given point in time
+SELECT * FROM iasql_get_sql_since((now() - interval '2 hours')::text);
+```
+
+Copy the SQL queries generated from invoking the function in the first database as a starting point for the SQL you will run in the second DB / environment, make any changes to the SQL if needed and include it within a new transaction in the second database. We do not recommend using database migrations systems to try to version control changes as outlined by this [RFC](https://github.com/iasql/iasql/blob/main/rfcs/006%20-%20Replicate%20changes%20between%20staging%20and%20prod%20RFC.md). One of the value propositions of IaSQL is that you get type safety and validation via FKs. However, these are different in each DB / environment. If you wish to version control or peer review infrastructure transactions, there is a Github Module in our roadmap that automatically puts up the SQL generated from the ongoing transaction and the preview output in a markdown file within a PR that others can approve. Let us know on [Discord](https://discord.iasql.com) and we can prioritize it.
```
mtp1376, line 37: Idea: `iasql_begin(transaction_name varchar)` and then `iasql_get_sql(transaction_name varchar)` to be able to get SQL for a specific transaction.

mtp1376 (commented):
nit

site/docs/how-to/envs.md
depombo: yes, we mentioned that in the discord channel. we should definitely make it happen!

dfellis: This idea is a bit *better* because the user can name their transaction and then reference it with that name. That's good for providing a comment on what it's about similar to git commit names.

We would still have to do something to make sure the names are unique, either explicitly failing if not unique or implicitly including a timestamp and assuming the latest one?

site/docs/how-to/envs.md
depombo: good point, fixed

site/docs/how-to/envs.md
mtp1376: Maybe the `iasql_begin(transaction_name varchar)` overload just uses the user-provided transaction name and the default `iasql_begin` names the transaction with a UUID generation.

re: a non-unique name provided by the user, I'd go with raising an error. Or maybe we can provide a function to rename (or replace their name with a UUID) an older transaction.

site/docs/concepts/transaction.md
```diff
@@ -5,7 +5,7 @@ slug: '/transaction'
 
 # IaSQL transactions at a high level
 
-IaSQL does not any special syntax, akin to `apply` in IaC, to work normally. You can just `SELECT/INSERT/UPDATE/DELETE` records and eventually the changes are reflected in your cloud account. This also works with migration systems without issues, though in an eventually-consistent fashion. The most basic architectural diagram is simply your IaSQL database and your cloud account with IaSQL in between.
+IaSQL does not any special syntax, akin to `apply` in IaC, to work normally. You can just `SELECT/INSERT/UPDATE/DELETE` records and eventually the changes are reflected in your cloud account. The most basic architectural diagram is simply your IaSQL database and your cloud account with IaSQL in between.
```
mtp1376, line 8:
```suggestion
IaSQL does not have any special syntax, akin to `apply` in IaC, to work normally. You can just `SELECT/INSERT/UPDATE/DELETE` records and eventually the changes are reflected in your cloud account. The most basic architectural diagram is simply your IaSQL database and your cloud account with IaSQL in between.
```
or
```suggestion
IaSQL does not need any special syntax, akin to `apply` in IaC, to work normally. You can just `SELECT/INSERT/UPDATE/DELETE` records and eventually the changes are reflected in your cloud account. The most basic architectural diagram is simply your IaSQL database and your cloud account with IaSQL in between.
```

mtp1376 (approved):
Not sure if the sentence is gramatically incorrect, but feels that way to me.

site/docs/concepts/transaction.md
```diff
@@ -24,7 +24,7 @@ IaSQL, on the other hand, is completely stateless. It acquires the state of both
...
+It is possible to perform changes to your cloud account synchronously by temporarily turning off the normal two-way propagation between the cloud and database through the usage of an IaSQL transaction using the provided PostgreSQL function `iasql_begin`. This lets you batch, or stage, changes together and then calling `iasql_commit` to mark the end of the transaction and propagate the changes from the database to the cloud account akin to a traditional [database transaction](https://en.wikipedia.org/wiki/Database_transaction). And similarly to database transactions, IaSQL transactions automatically roll back if the `commit` fails. As you might have learned the hard way, you can never count on a given cloud provider to behave consistently or predictably.
```
dfellis, line 27: Don't know about the snark in the last sentence.

site/docs/how-to/envs.md
depombo: > I would also drop any mention of the "github module on our roadmap." If someone explicitly asks us for a solution like that, then fine, but I'd rather not box ourselves in on a solution before we try to actually consider all of the alternatives and implement a solution in case that one turns out to be sub-optimal.

I think this a tricky balance to strike. I somewhat agree and somewhat disagree. Users want to version control their changes, but it will be up to us how we help them do it. That said, I will keep the details out as we collect more info about the use case and if we think of something better.

site/docs/how-to/envs.md
```diff
@@ -0,0 +1,37 @@
...
+
+--- calls iasql_commit and make the resulting changes in the cloud account
+select * from iasql_commit();
+```
+
+Once the transaction is committed successfully and the desired changes are committed into the DB for that environment, you can use another IaSQL function that looks at the audit log and generates the SQL needed to represent changes done from a given point in time.
+
+```sql
+-- gets SQL from the audit log from a given point in time
+SELECT * FROM iasql_get_sql_since((now() - interval '2 hours')::text);
```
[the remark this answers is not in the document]
depombo: good suggestion

dfellis (approved):

aguillenv (approved):

PR #2175, opened 2023-03-09
`aws_opensearch` module

mtp1376:


aguillenv (commented):
newly added opensearch tests seem to be failing

src/modules/aws_opensearch/entity/domain.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * @public
+   * Name to identify the domain
+   */
+  @Column({ nullable: false })
+  @cloudId
+  domainName: string;
+
+  @Column({ nullable: true })
```
yrobla, line 43: you are missing the description here on several fields, it means it will not be rendered properly on the documentation

src/modules/aws_opensearch/entity/domain.ts
```diff
@@ -0,0 +1,134 @@
...
+  endpoint?: string; // comes from the cloud
+  /**
+   * @public
+   * Region for the domain
+   */
+  @Column({
+    type: 'character varying',
+    nullable: false,
+    default: () => 'default_aws_region()',
+  })
```
yrobla, line 129: does OpenSearch support tags? we are adding tags everywhere in all the models

src/modules/aws_opensearch/mappers/domain.ts
```diff
@@ -0,0 +1,398 @@
...
+  private async domainMapper(domain: AwsDomain, region: string, ctx: Context) {
+    const out: Domain = new Domain();
+    if (domain.Deleted) return undefined;
+
+    out.domainName = domain.DomainName!;
+    if (domain.DomainEndpointOptions?.CustomEndpoint) {
+      const certificateId = awsAcmModule.certificate.generateId({
+        arn: domain.DomainEndpointOptions?.CustomEndpointCertificateArn ?? '',
+      });
+      out.customEndpoint = domain.DomainEndpointOptions.CustomEndpoint;
```
yrobla, line 103: do custom endpoints only need to have a custom certificate?

test/modules/aws-opensearch-integration.ts
```diff
@@ -0,0 +1,208 @@
...
+                  '${region}');
+      `,
+      undefined,
+      true,
+      () => ({ username, password }),
+    ),
+  );
+  it('applies creation of the domain and waits for the creation', commit());
+
+  it(
```
yrobla, line 125: worth adding to itDocs here

test/modules/aws-opensearch-integration.ts
```diff
@@ -0,0 +1,208 @@
...
+          .then((r: any) => r.text())
+          .then((r: string) => expect(r).toBe('Unauthorized'));
+      },
+    ),
+  );
+
+  it('starts a transaction', begin());
+  it(
+    'changes the password of the domain',
+    query(
```
yrobla, line 158: add to itDocs

package.json
```diff
@@ -91,6 +93,7 @@
     "@aws-sdk/client-iam": "^3.297.0",
     "@aws-sdk/client-lambda": "^3.297.0",
     "@aws-sdk/client-memorydb": "^3.297.0",
+    "@aws-sdk/client-opensearch": "^3.282.0",
```
dfellis, line 96: This needs to be updated to `^3.299.0`

package.json
```diff
@@ -63,6 +64,7 @@
     "aws-sdk-client-mock": "^2.1.1",
     "copyfiles": "^2.4.1",
     "jest": "^27.5.1",
+    "node-fetch": "2",
```
dfellis, line 67: Let's cut this out.

package.json
```diff
@@ -54,6 +54,7 @@
     "@types/lodash.isequal": "^4.5.6",
     "@types/lodash.pick": "^4.4.7",
     "@types/node": "^18.15.5",
+    "@types/node-fetch": "2",
```
dfellis, line 57: Same here, should remove this.

test/modules/aws-opensearch-integration.ts
```diff
@@ -0,0 +1,208 @@
...
+  finish,
+  getPrefix,
+  itDocs,
+  runBegin,
+  runCommit,
+  runInstall,
+  runQuery,
+} from '../helpers';
+
+const fetch = require('node-fetch');
```
dfellis, line 17: This isn't necessary in Node 18. If Jest is giving you issues with it not being here, you can [`declare` the `fetch` function similarly to this](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/14796/files#diff-0876e65c08560a069314bd2c7285249db01a47cfb3e8d6ee8f0ce71b7d46549eR111) perhaps replacing the arguments with `any` for simplicity.

src/modules/aws_opensearch/mappers/domain.ts
mtp1376: Yes. Only custom endpoints use it.

src/modules/aws_opensearch/entity/domain.ts
mtp1376: We can do that as a follow-up. I think I tried to tackle it at first but there was some complexities - and they're immutable once created.

src/modules/aws_opensearch/mappers/domain.ts
```diff
@@ -0,0 +1,398 @@
+import _ from 'lodash';
```
aguillenv, line 1: If I'm not wrong we have been importing only the functions we need and not the whole lib.

src/modules/aws_opensearch/mappers/domain.ts
aguillenv: and is it a must to have a certificate associated?

src/modules/aws_opensearch/mappers/domain.ts
```diff
@@ -0,0 +1,398 @@
...
+            WarmEnabled: !!e.warmInstanceCount,
+            WarmType: e.warmInstanceType,
+            WarmCount: e.warmInstanceCount,
+            ColdStorageOptions: {
+              Enabled: e.coldStorage,
+            },
+          },
+          AccessPolicies: JSON.stringify(e.accessPolicy),
+          SnapshotOptions: {
+            AutomatedSnapshotStartHour: 0, // input?
```
aguillenv, line 171: maybe set this as a `todo` if we want to make it configurable in the future

src/modules/aws_opensearch/mappers/domain.ts
```diff
@@ -0,0 +1,398 @@
...
+        if (!rawDomain || !rawDomain.DomainStatus) continue;
+        // now let's replace values with cloud ones to avoid infinite loop
+        const domain = await this.domainMapper(rawDomain.DomainStatus, e.region, ctx);
+        if (!domain) continue;
+        e.endpoint = domain.endpoint;
+        e.fineGrainedAccessControlMasterUsername =
+          e.fineGrainedAccessControlMasterPassword =
+          e.fineGrainedAccessControlUserArn =
+            undefined;
+        await this.module.domain.db.update(e, ctx);
```
aguillenv, line 242: Maybe it is better to do something like the following to work with the latest object?
```suggestion
        const domain = await this.domainMapper(rawDomain.DomainStatus, e.region, ctx);
        if (!domain) continue;
        domain.id = e.id;
        domain.fineGrainedAccessControlMasterUsername =
          domain.fineGrainedAccessControlMasterPassword =
          domain.fineGrainedAccessControlUserArn =
            undefined;
        await this.module.domain.db.update(domain, ctx);
```

src/modules/aws_opensearch/entity/domain.ts
```diff
@@ -0,0 +1,223 @@
...
+   */
+  @Column({ type: 'varchar', enum: OpenSearchPartitionInstanceType })
+  instanceType: OpenSearchPartitionInstanceType;
+
+  /**
+   * @public
+   * Number of instances that'll run opensearch
+   */
+  @Min(1)
+  @Max(80)
```
aguillenv, line 97: nit: maybe add the constraint in the documentation comment?
```suggestion
  /**
   * @public
   * Number of instances that'll run opensearch. Min: 1 - Max: 80
   */
  @Min(1)
  @Max(80)
```

test/modules/aws-opensearch-integration.ts
```diff
@@ -0,0 +1,206 @@
...
+        fetch('https://' + res[0].endpoint, {
+          method: 'GET',
+          headers,
+        })
+          .then((r: any) => r.json())
+          .then((r: any) => expect(r.version.number).toBe('2.3.0'));
+      },
+    ),
+  );
+
```
aguillenv, line 190: can we add a test here uninstalling and then installing again the module and check if the `domain` syncs properly from cloud?

src/modules/aws_opensearch/mappers/domain.ts
mtp1376: Yes. 
image

yrobla (approved):

aguillenv (approved):

PR #2180, opened 2023-03-10
Add getting started hint

### LINKED ISSUE #2103: Hint module installation after connecting the database
A few users are not understanding what to do next after connecting. "modules not installed" playlist. There is several ways this could be addressed.

It probably makes sense to open two tabs. The first install aws_ec2 and the second calls iasql help.

aguillenv:
Resolves #2103 

Any suggestion in the content is very welcome!!

image


mtp1376 (commented):
The help string feels a bit dense to me. Maybe this can be replaced by something like a tour?
Like https://introjs.com/

dfellis:
> The help string feels a bit dense to me. Maybe this can be replaced by something like a tour? Like https://introjs.com/

@depombo vetoed those as annoying and something he just clicks through anyways (I agree, though not as vehemently). This approach has the advantage of letting someone click around if they want, and then if they get confused or are unsure how to accomplish something, it's sitting there for them to read.

This may be down to different personality types, though, and we'll have to see what works better for most users?

aguillenv:
> The help string feels a bit dense to me. Maybe this can be replaced by something like a tour?
> Like https://introjs.com/

I've suggested something like that but @depombo and @dfellis do not like that much that kind of product tour.

dashboard/src/components/providers/AppProvider.tsx
```diff
@@ -110,6 +110,31 @@ const initializingQueries = `
...
+
+-- Going to the sidebar's "Schema" tab you can see the tables and columns of the installed modules. There you
+-- can check the documentation of each module. You can also do a quick select query on any table with data.
+
+-- This editor can autocomplete your query's table and column names based on the installed modules.
+
+-- If you want to see all the available IaSQL functions you can execute the following query:
+-- SELECT * FROM iasql_help()
+
+-- To disconnect your database, click "Settings" and then "Disconnect" on the top right.
```
dfellis, line 130: Oh, I thought we were going to unwind the settings dropdown? Did we decide against that?

mtp1376: I don't think we've decided against it. I was doing the task of removing export buttons and maybe I should've done it in my PR. I can submit another PR removing it.

depombo (approved):

dashboard/src/components/IasqlEditor.tsx
```diff
@@ -45,7 +45,7 @@ export default function IasqlEditor() {
 
   // Handlers
   const getInitialQuery = useCallback((sql: string | null) => {
-    let initialQuery = 'SELECT * FROM iasql_help();';
+    let initialQuery = editorTabs?.[editorSelectedTab]?.content || 'SELECT * FROM iasql_help();';
```
dfellis, line 48: Super minor nit. You're technically supposed to use `??` here instead of `||` since `false || 'something'` will become `'something'` while `false ?? 'something'` will be `false`, and only if the value is `null` or `undefined` will it fall through to the backup value.

dfellis (approved):

dashboard/src/components/IasqlEditor.tsx
aguillenv: yes, that's better, lets's fix it!
```suggestion
    let initialQuery = editorTabs?.[editorSelectedTab]?.content ?? 'SELECT * FROM iasql_help();';
```

PR #2182, opened 2023-03-10
Add docs about restricting AWS creds to a region

### LINKED ISSUE #2056: Add documentation about restricting AWS creds to specific regions + how to set up that db

aguillenv:
Resolves #2056

### DIFF
--- a/site/docs/how-to/aws.mdx
+++ b/site/docs/how-to/aws.mdx
@@ -49,3 +49,36 @@ SELECT
 FROM
   iasql_install ('aws_iam', 'aws_cloudwatch', 'aws_ecr', 'aws_ecs_fargate', 'aws_elb', 'aws_security_group', 'aws_vpc');
 ```
+
+## Restricting AWS credentials to specific regions
+
+Sometimes, you may want or need to restrict your credentials to a specific region. To do this, you will need to create a custom policy that restricts access to AWS resources in that region. The following [guide](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_create.html) will walk you through creating a new policy. Your policy will need to look something like this:
+
+```json
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": "*",
+      "Resource": "*",
+      "Condition": {
+        "StringEquals": {
+          "aws:RequestedRegion": "us-west-2"
+        }
+      }
+    }
+  ]
+}
+```
+
+You can replace 'us-west-2' with your desired region.
+
+Once you have the policy in place, you need to update your IAM identity (user, user group or role) and [attach the new policy](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_manage-attach-detach.html). By default, IaSQL attempts to load resources from all enabled regions. Assuming that our credentials only have access to a single region, let's say `us-west-2`, we need to run a specific query after connecting to our database to inform IaSQL that we only want to manage resources in that region:
+
+```sql
+-- Make sure our default region is `us-west-2` if we did not set it up correctly while connecting
+SELECT * FROM default_aws_region('us-west-2');
+-- Disable in IaSQL all other regions
+UPDATE aws_regions SET is_enabled = FALSE WHERE is_default = FALSE;
+```


site/docs/how-to/aws.mdx
```diff
@@ -49,3 +49,36 @@ SELECT
 FROM
   iasql_install ('aws_iam', 'aws_cloudwatch', 'aws_ecr', 'aws_ecs_fargate', 'aws_elb', 'aws_security_group', 'aws_vpc');
 ```
+
+## Restricting AWS credentials to specific regions
+
+Sometimes, you may want or need to restrict your credentials to a specific region. To do this, you will need to create a custom policy that restricts access to AWS resources in that region. The following [guide](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_create.html) will walk you through creating a new policy. Your policy will need to look something like this:
```
dfellis, line 55: I'm assuming we don't have the ability to do this in IaSQL itself yet, right?

If we do, we *may* want to also include that, though noting that you probably don't want to edit your own IAM identity this way, because we may lose the ability to perform the operations in the middle of performing said operations, and that it should be done from a different account.

aguillenv: hmm, we do not have policies management yet, but we do have a field in `user` and `role` entities that let you update the attached policies.

mtp1376: We can create that policy as a public policy in one of our account and use the ARN to attach that through IaSQL. That's a bit hacky and if we want to use a non-hacky way we can add the support for IAM policies.

dfellis: I think it can make sense to have both views, one for "you already have a restricted IAM account and need to keep IaSQL from trying to access regions it's not allowed to" which really only needs the last part, and the other view of "I am trying to set up a restricted IAM user to then use with IaSQL" which could have both "do it in the console" and "use my powerful IAM user to create the less powerful IAM user" explanations.

I guess what *this* particular article is aiming at is the question, but I assumed it was for people who already had restricted IAM accounts.

depombo: The first section of this guide already covers how to create the IAM credentials or how to associate them and including this modification around that seems sensible. I don't think that we need to lay out again how to provision an IAM role again or whether they should create a new one. We can simply say the credentials provided need a policy that looks like this:

```suggestion
Sometimes, you may want or need to restrict access to a specific region. To do this, you will need to make sure that the credentials for the IAM role given to IaSQL have a policy that restricts access only to AWS resources in that region. This can be done by creating a custom policy and attaching it to the relevant IAM role. The following [guide](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_create.html) will walk you through creating a new policy. Your policy will need to look something like this:
```

Is it possible to attach the policy inline?

aguillenv: yes, you can

yrobla (approved):

mtp1376 (approved):

site/docs/how-to/aws.mdx
```diff
@@ -49,3 +49,36 @@ SELECT
 FROM
   iasql_install ('aws_iam', 'aws_cloudwatch', 'aws_ecr', 'aws_ecs_fargate', 'aws_elb', 'aws_security_group', 'aws_vpc');
 ```
+
+## Restricting AWS credentials to specific regions
+
+Sometimes, you may want or need to restrict access to a specific region. To do this, you will need to make sure that the credentials for the IAM role given to IaSQL have a policy that restricts access only to AWS resources in that region. This can be done by creating a custom policy and attaching it to the relevant IAM role. The following [guide](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_create.html) will walk you through creating a new policy. Your policy will need to look something like this:
+
+```json
```
dfellis, line 57: One comment on this policy: It will only work as explained here if the IAM user in question doesn't have any policy giving access to AWS regions, yet. If that IAM user already has access to all regions, this policy will do nothing.

Perhaps there should be a second statement with an `"Effect": "Deny"` and the condition the mutated to `StringNotEquals`? (Which I just confirmed [is a real conditional](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html) )

PR #2184, opened 2023-03-10
Upgrade the module migration postgres to Postgres 14

dfellis:
We're using Postgres 14 within our docker container but building migrations with Postgres 13. Normally shouldn't matter, but in case there's a Postgres 14-only feature we want to use, would be better to keep them in sync

### DIFF
--- a/.module_migration.sh
+++ b/.module_migration.sh
@@ -15,7 +15,7 @@ docker volume prune -f
 docker image prune -f
 
 # Start a postgres container
-docker container run -p 5432:5432 -e POSTGRES_PASSWORD=test --name migrate-postgres -d postgres:13.4
+docker container run -p 5432:5432 -e POSTGRES_PASSWORD=test --name migrate-postgres -d postgres:14.7
 
 sleep 5 # Just in case
 psql postgresql://postgres:test@localhost:5432/postgres -c "CREATE DATABASE __example__"


mtp1376 (approved):
Isn't there any other places in which Postgres version is declared?

depombo (approved):

dfellis:
> Isn't there any other places in which Postgres version is declared?

`git grep` returns:

```
damocles@zelbinion:~/alantech/iasql(main)$ git grep 13.4
.module_migration.sh:docker container run -p 5432:5432 -e POSTGRES_PASSWORD=test --name migrate-postgres -d postgres:13.4
dashboard/package.json:    "pgsql-parser": "^13.4.1",
Binary file dashboard/yarn.lock matches
Binary file site/yarn.lock matches
src/modules/aws_acm/migration/1666218426906-aws_acm.ts:      `CREATE TABLE "certificate" ("id" SERIAL NOT NULL, "arn" character varying, "certificate_id" character varying, "domain_name" character varying NOT NULL, "certificate_type" "public"."certificate_certificate_type_enum", "status" "public"."certificate_status_enum", "renewal_eligibility" "public"."certificate_renewal_eligibility_enum", "in_use" boolean NOT NULL DEFAULT false, "region" character varying NOT NULL DEFAULT default_aws_region(), CONSTRAINT "UQ_f863cd2ad563468a1aa9341384f" UNIQUE ("arn"), CONSTRAINT "PK_8daddfc65f59e341c2bbc9c9e43" PRIMARY KEY ("id"))`,
test/modules/aws-rds-integration.ts:        VALUES ('${prefix}test', 20, 'db.t3.micro', 'test', 'testpass', (SELECT name FROM availability_zone WHERE region = '${region}' LIMIT 1), 'postgres:13.4', 0);
test/modules/aws-rds-integration.ts:        VALUES ('${prefix}test', 20, 'db.t3.micro', 'test', 'testpass', (SELECT name FROM availability_zone WHERE region = '${region}' LIMIT 1), 'postgres:13.4', 0);
test/modules/aws-rds-multi-region.ts:        VALUES ('${prefix}test', 20, 'db.t3.micro', 'test', 'testpass', (SELECT name FROM availability_zone WHERE region = '${region}' LIMIT 1), 'postgres:13.4', 0);
Binary file yarn.lock matches
```

The references to `postgres:13.4` in the RDS tests are explicitly about upgrading a database from an older version to a newer version, so those need to stay.

yrobla (approved):

PR #2230, opened 2023-03-14
Modules? In My Database?

dfellis:
*It's more likely than you think*


site/blog/unlisted/modules-in-my-database.mdx
```diff
@@ -0,0 +1,45 @@
...
+slug: modules-in-my-database
+title: Modules? In My Database? It's more likely than you think.
+authors: [dfellis]
+---
+
+# Modules? In My Database?
+
+## *It's more likely than you think*
+
+For IaSQL, as the primary user interface is PostgreSQL databases, we have some unique pressures on our tech stack compared to many other projects. One of these that [we didn't realize at first](https://github.com/iasql/iasql/commit/8467bde2709d32cbe625c2fc2d1a312b5eefaf2d), was that we can't use a traditional linear migration system for database schema management. Different users will want to work with a different subset of AWS services, and requiring the users to deal with *all* of the AWS schema would be too overwhelming; we felt overwhelmed looking at it ourselves!
```
depombo, line 11: I feel like the primary motivation is allowing versioning of schema changes (from us and the cloud) over time, even more than being able to use AWS piecemeal with IaSQL, so it might be better to start with that?

PR #2246, opened 2023-03-15
Fix local mezmo logging

dfellis:
It currently emits an error locally because we didn't have a path for logging to mezmo *and* disabling debug logs.

### DIFF
--- a/src/services/logger.ts
+++ b/src/services/logger.ts
@@ -22,22 +22,42 @@ const logFactory: LogFunctionFactory<Logger> = (scope: any) => {
       console.log(`Fatal error in LogDNA ${util.inspect(event)}`);
     });
     if (config.logger.forceLocal) {
-      return async (level, message, meta) => {
-        const { dbId } = scope;
-        let userId: string | undefined;
-        if (dbId) {
-          const user = await MetadataRepo.getUserFromDbId(dbId);
-          userId = user?.id;
-        }
-        console.log(`${level}: ${message} ${util.inspect({ ...meta, userId, dbId })}`);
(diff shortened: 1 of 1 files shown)


yrobla (approved):

src/services/logger.ts
```diff
@@ -22,22 +22,42 @@ const logFactory: LogFunctionFactory<Logger> = (scope: any) => {
...
+            level: level === 'warning' ? 'warn' : level, // Graphile Logger vs LogDNA levels fix
+            meta: { ...meta, userId, dbId },
+            indexMeta: true,
+            app: 'iasql-engine',
+            env: process.env.IASQL_ENV,
+          });
+        };
+      } else {
+        return async (level, message, meta) => {
+          if (level === 'debug') return;
```
aguillenv, line 44: but how does this fix current behaviour if our current `local` has `debug=false`?

dfellis: The `logger.debug` function calls *still* happen, and are eventually passed to our log handling function. In this function where the config disables `debug` we check if the log is a `debug` type and drop it. Otherwise we let it through, assuming it is one of the three levels we have enabled (`info`, `warn`, `error`).

aguillenv (approved):

depombo (approved):
Looks good, can/did we test this locally?

dfellis:
> Looks good, can/did we test this locally?

Yes, I monitored the mezmo logs and the CLI output when that was happening. The dashboard logs were the only ones I saw locally (since I didn't trigger an error, only warnings) and I saw the warnings in mezmo.

PR #2249, opened 2023-03-15
Don't show the "x" button on DB creation step

mtp1376:
Before connection step:
Screen Shot 2023-03-15 at 15 44 20

While connecting:
Screen Shot 2023-03-15 at 15 44 29

On error:
Screen Shot 2023-03-15 at 15 46 48

Screen Shot 2023-03-15 at 15 47 06

### DIFF
--- a/dashboard/src/components/Connect.tsx
+++ b/dashboard/src/components/Connect.tsx
@@ -63,7 +63,6 @@ export default function Connect({ closable }: { closable: boolean }) {
         }
       }}
       backEnabled={backEnabled}
-      closeable={closable}
       onClose={() => {
         dispatch({ action: ActionType.ShowConnect, data: { showConnect: false } });
         if (newDb) dispatch({ action: ActionType.ResetNewDb });
--- a/dashboard/src/components/Connect.tsx
+++ b/dashboard/src/components/Connect.tsx
@@ -20,12 +20,14 @@ export default function Connect({ closable }: { closable: boolean }) {
 
   let nextEnabled = true;
   let backEnabled = false;
+  let closeButtonEnabled = true;
   const current = stack[stack.length - 1];
   // Check relevant state per step to determine automatic actions to perform, such as deciding if
   // the Next button should be enabled or not
   switch (current) {
     case 'createdb':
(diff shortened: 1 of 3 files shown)


dashboard/src/components/common/Wizard.tsx
```diff
@@ -49,7 +47,7 @@ export function Wizard({
     <Modal
       icon={icon}
       title={title}
-      closeable={closeable ?? false}
+      closeable={(currentStep as any)?.props?.id !== 'execdbcreate'}
```
aguillenv, line 50: I don't think this is the best solution. The `Wizard` component should not know about the specifics of the `Connect` component.

dashboard/src/components/Connect.tsx
```diff
@@ -63,7 +63,6 @@ export default function Connect({ closable }: { closable: boolean }) {
         }
       }}
       backEnabled={backEnabled}
-      closeable={closable}
```
aguillenv, line 66: if we just pass `false` here, then there no way back if you click on "connect", right?

mtp1376: Yes. If we do that then there's no way to close this dialog:
image

dashboard/src/components/common/Wizard.tsx
mtp1376: I get your point and I'll refactor the code to not depend on the internals of its caller component. But the `Wizard` component has not been used anywhere else.

dashboard/src/components/Connect.tsx
aguillenv: I'd suggest maybe taking a look at how the `nextEnabled` works and do something similar

dashboard/src/components/Connect.tsx
```diff
@@ -63,7 +67,7 @@ export default function Connect({ closable }: { closable: boolean }) {
...
+      closeable={closeButtonEnabled}
       onClose={() => {
```
aguillenv, line 71: if I'm not wrong this function needs to be updated too (like on `onNext`)! Since it is a modal I expect it to be closable also if you click outside the modal.

dfellis (approved):

aguillenv (approved):

Issue #2159, opened 2023-03-08
Add tag support for resources other than just ec2 instances


depombo:
- [x] aws_acm https://github.com/iasql/iasql/pull/2234
- [x] aws_apigateway https://github.com/iasql/iasql/pull/2252
- [x] aws_appsync https://github.com/iasql/iasql/pull/2254
- [ ] aws_cloudfront 
- [ ] aws_cloudwatch
- [ ] aws_codedeploy
- [ ] aws_codepipeline
- [ ] aws_dynamo
- [ ] aws_ec2 (apart from instance and gpv)
- [ ] aws_ecr
- [ ] aws_ecs
- [ ] aws_elasticache
- [ ] aws_elb
- [ ] aws_iam
- [x] aws_rds (@yrobla https://github.com/iasql/iasql/pull/2304)
- [ ] aws_s3 (@mtp1376 )
- [ ] aws_secrets_manager
- [ ] aws_security_group
- [ ] aws_sns
- [ ] aws_sqs
- [ ] aws_vpc (most of them have it except subnet at least)
- [ ] aws_ecs_simplified?

depombo:
This code will be very similar, but ideally we can have a macro or something to DRY this implementation out or at least for majority of the modules.

PR #2252, opened 2023-03-15
Add `aws_apigateway` tags

aguillenv:
Part of #2159


src/modules/aws_api_gateway/mappers/api.ts
```diff
@@ -0,0 +1,199 @@
...
+          if (!eqTags(r.tags, cloudRecord.tags)) {
+            const apiArn = this.buildApiArn(r.apiId, r.region);
+            await this.untagRes(client.apiGatewayClient, {
+              ResourceArn: apiArn,
+              TagKeys: Object.keys(cloudRecord.tags ?? {}),
+            });
+            // sleep for a few seconds to allow the tags to be applied
+            await new Promise(resolve => setTimeout(resolve, 5000));
+            await this.tagRes(client.apiGatewayClient, { ResourceArn: apiArn, Tags: r.tags });
+            // sleep for a few seconds to allow the tags to be applied
```
mtp1376, line 151: Is there a way with API calls to do this? I know tags should not take more than 5 seconds to apply, but if there's a way to wait for the result in a more programmatic way I think it'd be better.

aguillenv: yeah, agree it is definitely not ideal but already spent too much time on this change. There were a lot of inconsistencies with this API. I can try to reduce the time to a couple of seconds and see if tests still pass and the programmatic version can come as a follow-up.

mtp1376: The AWS API caches based on the exact input for 40 seconds. It might be the case since I had that issue with the inconsistencies: https://gist.github.com/mtp1376/478860e7acbe55fab3727db968375eaf

aguillenv: Addressed using waiters!!

yrobla (approved):

depombo (approved):

PR #2256, opened 2023-03-15
Change input type of `iasql_get_sql_since` to be a Postgres date

mtp1376:


src/modules/iasql_functions/rpcs/iasql_get_sql_since.ts
```diff
@@ -45,7 +45,7 @@ export class IasqlGetSqlSince extends RpcBase {
 
   documentation = {
     description: 'Generate SQL from the audit log from a given point in time',
-    sampleUsage: "SELECT * FROM iasql_get_sql_since((now() - interval '30 minutes')::text)",
```
depombo, line 48: /docs/envs also needs to change

PR #2260, opened 2023-03-16
Assign `transactionId` to `iasql_audit_log` records

### LINKED ISSUE #2258: Associate audit log entries with transactions

aguillenv:
Resolves #2258 

Audit logs:
image

iasql_begin gives you the new transaction id:
image

A new UPDATE executed
image

Run `iasql_preview`, `iasql_commit` or `iasql_rollback` to associate the change to the current transaction:
image

Get sql for the transaction:
image


src/modules/iasql_functions/rpcs/iasql_begin.ts
```diff
@@ -33,9 +34,10 @@ export class IasqlBegin extends RpcBase {
   call = async (
     _dbId: string,
     _dbUser: string,
-    _ctx: Context,
+    ctx: Context,
   ): Promise<RpcResponseObject<typeof this.outputTable>[]> => {
-    const message = 'Transaction started';
+    const currentTransactionId = await getCurrentTransactionId(ctx.orm);
+    const message = `Transaction ${currentTransactionId} started`;
     return [{ message }];
```
dfellis, line 41: Nit: We should also return the `transaction_id` as a second column so it is more easily joinable/usable within their own logic, like placing it into a pl/pgsql variable to reference for other calls.

dfellis (approved):

PR #2264, opened 2023-03-16
Add proxy support to IaSQL

dfellis:
This may resolve the issue being seen


src/services/aws_macros.ts
```diff
@@ -26,9 +26,13 @@ import { SQS } from '@aws-sdk/client-sqs';
 import { SSM } from '@aws-sdk/client-ssm';
 import { STS } from '@aws-sdk/client-sts';
 import { defaultRetryDecider, StandardRetryStrategy } from '@aws-sdk/middleware-retry';
+import { NodeHttpHandler } from '@aws-sdk/node-http-handler';
+import { HttpProxyAgent, HttpsProxyAgent } from 'hpagent';
 
 import config from '../config';
 
+const useProxy = !!process.env.HTTPS_PROXY && !!process.env.HTTP_PROXY;
```
aguillenv, line 34: Is not possible to be defined just one of them?

aguillenv (commented):
These new env variables will need to be added to the docker run command, right?

src/services/aws_macros.ts
dfellis: It is, but I would consider that a misconfiguration (for now, at least). The [AWS SDK Docs](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-proxies.html) imply in the call-out at the end that both are necessary to be sure the SDK works. It's possible that this isn't really true and we can get away with just the `HTTPS_PROXY` instead of both, but I am erring on the safe side for now -- especially since the user who needs this defines both.

depombo (approved):

yrobla (commented):
is there any way to add some tests for it?

aguillenv (approved):

mtp1376 (approved):

mtp1376:
If the proxy is remote, that's easy to use by simply setting those values. But if it's running on the Docker host machine, then it should be something like `host.docker.internal:8888` since the container is trying to reach that address, not the host.

dfellis:
> If the proxy is remote, that's easy to use by simply setting those values. But if it's running on the Docker host machine, then it should be something like `host.docker.internal:8888` since the container is trying to reach that address, not the host.

So it's a proxy provided by the user. The user who needs this shared their configuration and it's a bare IP address their IT team gave them, so there's no issue in accessing the "outer" `localhost` here. :)

dfellis:
> is there any way to add some tests for it?

That is my plan after this works; however sharing this particular branch with the user did not resolve their issues, so I'm trying to reproduce things on my end in a VM and will come back to this later, as I may also need to make more changes to this PR.

dfellis:
This did not produce a difference in behavior.

PR #2265, opened 2023-03-16
Add Domain, SSL, ELB part to IaSQL on EC2 blog post

mtp1376:


site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -203,3 +203,138 @@ WHERE id = (SELECT id
...
+
+The second query confirms that we have a new certificate for our domain, and it's ready to be used (status = `ISSUED`). Now we're ready to launch our load balancer, listen for HTTPS traffic with this new certificate, and forward the incoming traffic to our EC2 instance's 9876 port.
+
+#### Spawn An Elastic Load Balancer
+To set up an elastic load balancer, we need some kinds of resources:
+- `load_balancer` itself
+- A `target_group` pointing to our EC2 instance on port 9876
+- A `listener` on port 443 of the load balancer
+
+<img src={'/img/iasql-on-ec2/load-balancer-parts.png'} style={{maxWidth: 300}} />
```
depombo, line 282: I don't this zoomed-in diagram with so few moving pieces is adding any value clarity. same for the one that is already in the blog post further up. we should remove these and just keep the big, high-level one

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -203,3 +203,138 @@ WHERE id = (SELECT id
...
+FROM resource_record_set
+WHERE parent_hosted_zone_id = (SELECT id FROM hosted_zone WHERE domain_name = '<your-domain>')
+  AND record_type = 'NS';
+```
+After updating the name servers in the domain settings, we need to wait for those values to be propagated. This can happen in minutes, or even up to 48 hours depending on the registrar.
+
+#### Issue An SSL Certificate For Our Domain
+Having the domain setup in Route53, it's easy to use IaSQL's `certificate_request` RPC to issue a new certificate for it. If you try to go with this procedure through AWS console, you'll need to manually add the CNAME records from the AWS ACM to your hosted zone. But this RPC is implemented in a way that automagically [sets the DNS records](https://github.com/iasql/iasql/blob/ed1b1ef92ceaa92e7345449992c35a56b62fe571/src/modules/aws_acm/rpcs/request.ts#L177-L188), [waits for the certificate to be issued](https://github.com/iasql/iasql/blob/ed1b1ef92ceaa92e7345449992c35a56b62fe571/src/modules/aws_acm/rpcs/request.ts#L197), and then [removes those validation DNS records](https://github.com/iasql/iasql/blob/ed1b1ef92ceaa92e7345449992c35a56b62fe571/src/modules/aws_acm/rpcs/request.ts#L200).
```
yrobla, line 262: why are you pointing to that specific commit there, instead of pointing to main?

site/blog/tutorials/iasql-on-ec2.mdx
```diff
@@ -203,3 +203,138 @@ WHERE id = (SELECT id
...
+       (true, '-1', -1, -1, '0.0.0.0/32', 'allow outgoing traffic from load balancer',
+        (SELECT id FROM lb_security_group), default_aws_region());
+```
+
+Then perform creation of the actual load balancer, the target group, and the listener:
+```sql
+INSERT INTO load_balancer (load_balancer_name, scheme, load_balancer_type, ip_address_type, region)
+VALUES ('iasql-lb', 'internet-facing', 'application', 'ipv4', default_aws_region());
+
+INSERT INTO target_group (target_group_name, target_type, protocol, port)
```
yrobla, line 305: should not be HTTP? listener is HTTPS but traffic inside VM is just HTTP right?

site/blog/tutorials/iasql-on-ec2.mdx
mtp1376: Because the `main` branch is subject to change but these links are permalinks that are valid even when the code on `main` changes.

site/blog/tutorials/iasql-on-ec2.mdx
mtp1376: Yes you're right. I'll fix it.

site/blog/tutorials/iasql-on-ec2.mdx
dfellis: Could we change it to one of the versioned tags, though? Linking to `v0.1.6` is a less wonky URL and also includes some information about how "old" these links are relative to whatever the latest release is at the time the user reads this.

site/blog/unlisted/iasql-on-ec2.mdx
```diff
@@ -208,3 +200,130 @@ WHERE id = (SELECT id
...
+SELECT iasql_commit();
+```
+
+:::note
+Simple `INSERT`, right? But wait! What are those `iasql_begin` and `iasql_commit` functions called before and after our insert command?
+
+Quick answer is that if you need your changes to be applied now instead of eventually in the future, you'll wrap the changes between an `iasql_begin` and an `iasql_commit`. For more info on what they do, visit [this page](https://iasql.com/docs/transaction/).
+:::
+
+Now that we have created our hosted zone, we need to set our domain's nameservers to the ones that Route53 provides us. For each public hosted zone that you create, Amazon Route53 [automatically creates](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/SOA-NSrecords.html) a name server (NS) record and a start of authority (SOA) record. So we should log into the domain registrar, and set the name server values to the four name server values that Route53 has set. To get those values, we can simply run a `SELECT` query:
```
dfellis, line 243: Nit

```suggestion
Now that we have created our hosted zone, we need to set our domain's nameservers to the ones that Route53 provides us. For each public hosted zone that you create, Amazon Route53 [automatically creates](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/SOA-NSrecords.html) a name server (NS) record and a start of authority (SOA) record. So we need to log into our domain registrar and set the name server values to the four name server values that Route53 has set. How this is done exactly varies from registrar to registrar, but we can get those NS values with a `SELECT` query:
```

site/blog/unlisted/iasql-on-ec2.mdx
```diff
@@ -208,3 +200,130 @@ WHERE id = (SELECT id
...
+```sql
+SELECT record as name_server
+FROM resource_record_set
+WHERE parent_hosted_zone_id = (SELECT id FROM hosted_zone WHERE domain_name = '<your-domain>')
+  AND record_type = 'NS';
+```
+After updating the name servers in the domain settings, we need to wait for those values to be propagated. This can happen in minutes, or even up to 48 hours depending on the registrar.
```
dfellis, line 250: Providing some more actionable advice here to know when you can move on in the tutorial.

````suggestion
After updating the name servers in the domain settings, we need to wait for those values to be propagated. This can happen in minutes, or even up to 48 hours depending on the registrar.

You can find out which NS records are accessible from your DNS with the [`dig` command](https://linux.die.net/man/1/dig):

```sh
dig <your_domain_name> ns
```

It should look something like this when ready:

```sh
$ dig skybase.dev ns

; <<>> DiG 9.18.1-1ubuntu1.3-Ubuntu <<>> skybase.dev ns
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 27469
;; flags: qr rd ra; QUERY: 1, ANSWER: 4, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 65494
;; QUESTION SECTION:
;skybase.dev.		                IN	NS

;; ANSWER SECTION:
skybase.dev.	        21600	IN	NS	ns-1243.awsdns-27.org.
skybase.dev.	        21600	IN	NS	ns-1944.awsdns-51.co.uk.
skybase.dev.	        21600	IN	NS	ns-49.awsdns-06.com.
skybase.dev.	        21600	IN	NS	ns-697.awsdns-23.net.

;; Query time: 60 msec
;; SERVER: 127.0.0.53#53(127.0.0.53) (UDP)
;; WHEN: Wed Mar 29 20:48:04 CDT 2023
;; MSG SIZE  rcvd: 179
```
````

File: site/blog/unlisted/iasql-on-ec2.mdx [hunk 6]
```diff
@@ -208,3 +200,130 @@ WHERE id = (SELECT id
...
+
+#### Issue An SSL Certificate For Our Domain
+Having the domain setup in Route53, it's easy to use IaSQL's `certificate_request` RPC to issue a new certificate for it. If you try to go with this procedure through AWS console, you'll need to manually add the CNAME records from the AWS ACM to your hosted zone. But this RPC is implemented in a way that automagically [sets the DNS records](https://github.com/iasql/iasql/blob/v0.1.7/src/modules/aws_acm/rpcs/request.ts#L177-L188), [waits for the certificate to be issued](https://github.com/iasql/iasql/blob/v0.1.7/src/modules/aws_acm/rpcs/request.ts#L197), and then [removes those validation DNS records](https://github.com/iasql/iasql/blob/v0.1.7/src/modules/aws_acm/rpcs/request.ts#L200).
+
+```sql
+-- SELECT iasql_install('aws_acm'); -- no need for this because aws_route53 depends on it and it's automatically installed
+SELECT certificate_request('<your-domain>', 'DNS', default_aws_region(), '{}');
```
dfellis, line 257: Note for a future feature addition: there should be a simplification of this function to not need the last 2 (or even 3?) arguments. These seem like reasonable defaults to me.

site/blog/unlisted/iasql-on-ec2.mdx
```diff
@@ -208,3 +200,130 @@ WHERE id = (SELECT id
...
+```
+
+The only step remaining is to allow incoming traffic to our instance on port 9876 only from our load balancer:
+```sql
+UPDATE security_group_rule
+SET source_security_group = (SELECT id FROM security_group WHERE group_name = 'iasql-lb'),
+    cidr_ipv4             = NULL
+WHERE cidr_ipv4 = '<my-ip>';
+```
+(we're basically updating the previous rule allowing direct access from our IP to allow access from load balancer security group)
```
dfellis, line 316: I think this makes it a bit clearer *exactly* what is going on.

```suggestion
(we're basically updating the previous rule that allowed direct access from our IP to no longer allow direct access and instead only allow access from load balancer security group)
```

site/blog/unlisted/iasql-on-ec2.mdx
```diff
@@ -208,3 +200,130 @@ WHERE id = (SELECT id
...
+
+Waiting for enough time for the above changes to be propagated to the cloud, the load balancer should be up and listening on a DNS name:
+```sql
+SELECT dns_name FROM load_balancer WHERE load_balancer_name = 'iasql-lb'
+```
+
+Making a CNAME record pointing to that load balancer will make IaSQL working under our domain name:
```
dfellis, line 323:
```suggestion
Finally, we will create a CNAME record pointing to that load balancer that will make IaSQL working under our domain name:
```

dfellis (approved):

site/blog/unlisted/iasql-on-ec2.mdx
depombo: agreed

site/blog/unlisted/iasql-on-ec2.mdx
```diff
@@ -208,3 +201,167 @@ WHERE id = (SELECT id
...
+skybase.dev.	        21600	IN	NS	ns-697.awsdns-23.net.
+
+;; Query time: 60 msec
+;; SERVER: 127.0.0.53#53(127.0.0.53) (UDP)
+;; WHEN: Wed Mar 29 20:48:04 CDT 2023
+;; MSG SIZE  rcvd: 179
+```
+:::
+
+#### Issue An SSL Certificate For Our Domain
```
depombo, line 290: what does this look like? the fact that this grayed out makes me think if might not properly parsing the sections

depombo (approved):

site/blog/unlisted/iasql-on-ec2.mdx
dfellis: Docusaurus' MDX markdown is *not* Github-style markdown, so I do highly recommend actually pulling the branch, and running:

```
cd site
yarn
yarn start:local
```

and navigating to the blogpost to see what it'll actually look like. That's what I've been doing on this PRs.

mtp1376: image

PR #2272, opened 2023-03-17
Fix query recreation

### LINKED ISSUE #2263: Queries are not being recreated properly in `iasql_get_sql_since` and future `iasql_get_sql_for_transaction`
This needs to be fixed for RFC007

aguillenv:
Resolves #2263


.github/workflows/tests.yml
```diff
@@ -19,7 +19,7 @@ on:
         'Dockerfile',
       ]
   pull_request:
-    branches: [main]
+    branches: [main, transaction-and-log]
```
depombo, line 22: forgot to rm this?

depombo (approved):
minor comment, but looks good otherwise

.github/workflows/tests.yml
aguillenv: Added it to force the test suite to run since it is not based on `main` but `transaction-and-log` branch

dfellis (approved):

depombo (approved):

PR #2275, opened 2023-03-18
ssh_accounts module

dfellis:


src/modules/ssh_accounts/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      ctx.orm.find(
+        SshCredentials,
+        name
+          ? {
+              where: {
+                name,
+              },
+            }
+          : undefined,
+      ),
```
aguillenv, line 25: aren't we abusing the `cloud.read` method using a db call?

test/modules/ssh-accounts-integration.ts
```diff
@@ -193,6 +193,8 @@ describe('EC2 Integration Testing', () => {
 
   it('installs the ssh_accounts and aws_ec2_metadata modules', install(['ssh_accounts', 'aws_ec2_metadata']));
 
+  it('starts a transaction', begin());
```
dfellis, line 196: This shouldn't matter because the `ssh_credentials` table, like the `aws_credentials` table, doesn't actually sync with anything in any cloud.

dfellis:
> Cool feature. I think it's good to have a docs page or blog post on this.

Yes to both, I will try to figure out the docs integration in this PR and a blog post will come after we have at least two other modules (the three I'm thinking off are: a module to manage systemd services, a module to manage apt packages, and a module to manage docker containers. Docker is definitely priority number one in that list, but not sure which of the other two to tackle next. Also feel free to take any of these or suggest your own. I figured something that replicates `top` could be useful, too?)

yrobla (commented):
Just a general question... why do specific commands such as rm, mkdir, etc... and not just simply allow to execute any generic command?

test/modules/ssh-accounts-integration.ts
```diff
@@ -0,0 +1,397 @@
...
+  // generate keypairs
+  it(
+    'generates a new keypair',
+    query(
+      `
+    SELECT * FROM key_pair_request ('${prefix}-key-request', '${region}');
+  `,
+      (res: any[]) => {
+        expect(res.length).toBe(1);
+        console.log(res);
```
aguillenv, line 140:
```suggestion
```

src/modules/ssh_accounts/rpcs/exec.ts
aguillenv, line 1: should we add comments for documentation in all these RPCs as we have in other RPCs?

dfellis:
> Just a general question... why do specific commands such as rm, mkdir, etc... and not just simply allow to execute any generic command?

So, two reasons:

1. Those are actually following the `sftp` protocol, which should make it very slightly more efficient than passing it through a shell to re-parse the commands (though users can certainly do so by literally just running `ssh_exec('<servername>', 'ls')` for example)
2. The more important reason is that anything coming out of `ssh_exec` is only an unparsed string that you'd have to format yourself, while `ssh_ls` immediately returns a table with useful columns you can query on, like this:

Screenshot from 2023-03-18 16-14-22

yrobla (approved):

mtp1376 (approved):

aguillenv (approved):

PR #2283, opened 2023-03-20
add RDS tag support

yrobla:


aguillenv:
Not sure what was the scope you had in mind, but could you please add the tags support for all entities in this `aws_rds` module? That way could be part of #2159

PR #2299, opened 2023-03-20
Add logdna to the dashboard

dfellis:


dashboard/src/pages/index.tsx
```diff
@@ -30,6 +31,13 @@ export default function App() {
     if (telemetry !== undefined && telemetry === 'on') {
       Sentry.init(config);
       Posthog.init(config);
+      if (config?.logdna?.key) {
+        logdna.init(config.logdna.key);
```
depombo, line 35: do we need a different one for the server per https://github.com/logdna/logdna-browser#nextjs-and-server-side-rendering or we are good ?

aguillenv (commented):
do we need to add explicit log calls?

dashboard/src/pages/index.tsx
dfellis: Technically we do, but we have only two server endpoints with the `dashboard/src/pages/api/config.ts` and `run.ts` files. In both cases they immediately punt data to the front end so I figured the logging there would handle anything relevant?

dfellis:
> do we need to add explicit log calls?

By [default](https://github.com/logdna/logdna-browser#configuration-options) it intercepts *all* `console.log/warn/error` calls, so we should be fine.

dashboard/src/pages/index.tsx
depombo: well, `run.ts` could have something off in the Postgres querying? not blocking though, we can add it later if we want

aguillenv (approved):

depombo (approved):

PR #2307, opened 2023-03-21
fix: specify engine version on test

yrobla:


test/modules/aws-rds-cluster-integration.ts
```diff
@@ -40,6 +40,7 @@ const region = defaultRegion([
 ]);
 
 const modules = ['aws_security_group', 'aws_rds', 'aws_vpc'];
+const engineVersion = '8.0.28';
```
aguillenv, line 43: are we sure this one exists across all regions?

PR #2311, opened 2023-03-21
set expectation on how long install and commit will take

depombo:
Untitled 2


dashboard/src/components/QueryResult.tsx
```diff
@@ -1,14 +1,27 @@
...
+  const query = editorTabs[editorSelectedTab]?.content;
+  const longRunningRpc = query.includes('iasql_commit()') || query.includes('iasql_install(');
 
   // TODO: show all statements with its respective query and not just the last one
   return editorTabs?.[editorSelectedTab]?.isRunning ? (
-    <Spinner />
+    longRunningRpc ? (
+      <VBox>
+        <p className='pt-20 px-20 font-bold'>
+          We are importing the entire state of the infrastructure from the cloud account and due to default
```
aguillenv, line 16: maybe the `We are importing the entire state...` does not align 100% with `iasql_commit`?

depombo: technically we do a sync on commit, but yes. perhaps `reading` or `parsing`?

aguillenv: What about: `Cloud operations can take several minutes due to default cloud SDK rate limiters depending on how many resources are in the account.  Please be patient.`

aguillenv (approved):
looks good to me! Maybe update the text but not blocking

dfellis (approved):

PR #2315, opened 2023-03-21
Shorter get started comment

depombo:
Users don't seem to be reading through the entire text so making it a bit shorter to make sure the most important points come across


dfellis (approved):
I guess that's one way to fix it. :zany_face:

PR #2324, opened 2023-03-22
fix: add better support and example for cluster deletion

yrobla:


src/modules/aws_rds/mappers/rds.ts
```diff
@@ -396,6 +396,17 @@ export class RdsMapper extends MapperBase<RDS> {
       for (const e of es) {
         // if instance has an associated cluster we do not delete it
         if (e.dbCluster) {
+          // check if it exists in the db, or has been deleted due to cascade
+          const cluster = await ctx.orm.findOne(DBCluster, {
+            where: { dbClusterIdentifier: e.dbCluster.dbClusterIdentifier },
+          });
```
aguillenv, line 402: why don't we call the mapper? `this.module.dbCluster.db.read(id, ctx)`

src/modules/aws_rds/mappers/db_cluster.ts
```diff
@@ -330,6 +330,12 @@ export class DBClusterMapper extends MapperBase<DBCluster> {
     delete: async (es: DBCluster[], ctx: Context) => {
       for (const e of es) {
         const client = (await ctx.getAwsClient(e.region)) as AWS;
+
+        // check if deletion protection is enabled and throw error
+        if (e.deletionProtection)
+          throw new Error(
+            "Cannot delete a cluster with deletion protection. Please set the cluster's deletion protection to false and try again.",
+          );
```
aguillenv, line 338:
```suggestion
        if (e.deletionProtection) {
          throw new Error(
            "Cannot delete a cluster with deletion protection. Please set the cluster's deletion protection to false and try again.",
          );
        }
```

aguillenv: also, should we `throw` immediately or should we `continue` and `throw` later?

src/modules/aws_rds/mappers/rds.ts
yrobla: i wanted to avoid any possible caching. So mapper could be affected by caching, and entry could be there, even if the entry has been removed from database right? I mostly want to avoid recreation in case that it has been removed due to cascade

src/modules/aws_rds/mappers/db_cluster.ts
yrobla: you mean, throwing after all other entries have been processed?

src/modules/aws_rds/mappers/rds.ts
aguillenv: I see, you could avoid it by deleting the cache first before calling the mapper. This is definitely a change we need to add to the mappers call, the possibility to pass a `force` or `no-cache` flag.

src/modules/aws_rds/mappers/db_cluster.ts
aguillenv: yes, I think we do not do it in other mappers, but thinking about it, it might be the right way to do it. That way we process what we can always. What do you think?

src/modules/aws_rds/mappers/rds.ts
```diff
@@ -430,6 +430,17 @@ export class RdsMapper extends MapperBase<RDS> {
       for (const e of es) {
         // if instance has an associated cluster we do not delete it
         if (e.dbCluster) {
+          // check if it exists in the db, or has been deleted due to cascade
+          const dbId = this.module.dbCluster.entityId(e.dbCluster);
+          delete ctx.memo.db.dbCluster[dbId];
```
aguillenv, line 435:
```suggestion
          delete ctx?.memo?.db?.dbCluster?.[dbId];
```

aguillenv (approved):

test/modules/aws-rds-cluster-integration.ts
```diff
@@ -225,10 +225,10 @@ describe('DB Cluster Integration Testing', () => {
   it('starts a transaction', begin());
 
   itDocs(
```
dfellis, line 227: Can we put a query just before this where it tries to delete the delete-protected cluster, fails, and then runs a second query verifying that the record has been restored in the database?

dfellis (approved):

PR #2351, opened 2023-03-23
Remove retry from tutorials

aguillenv:
It does not make sense to retry in the tutorials since they use migrations

### DIFF
--- a/.github/workflows/tests-django.yml
+++ b/.github/workflows/tests-django.yml
@@ -126,15 +126,11 @@ jobs:
           PSQL_CONN: postgres://postgres:test@localhost:5432/iasql
 
       - name: Run deploy script
-        uses: nick-fields/retry@v2
-        with:
-          max_attempts: 2
-          retry_on: error
-          timeout_minutes: 40
-          command: |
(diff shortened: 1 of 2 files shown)


mtp1376 (approved):

dfellis:
Approved, but since we also don't recommend migration tools for this, anymore, I am wondering if we should rework this "end-to-end" test, anyway?

depombo:
I'm not sure if we should land this for prisma

> > > Approved, but since we also don't recommend migration tools for this, anymore, I am wondering if we should rework this "end-to-end" test, anyway?
> > 
> > 
> > great point. prisma doesn't use migrations, but Django does. created this issue to fix that #2354
> 
> That wasn't what I had in mind, exactly. I think both should be reworked to follow the guide you wrote for change management. The guide is pure SQL, and we should try to replicate these examples that way, and then we should see if/how to do change management that way via ORM, but that would require changes to the Prisma tutorial, too.

I think we should do that too, but that seems like another task / somewhat orthogonal to removing migrations from Django.  A tutorial is a quickstart on how to get started with something quickly. I think it might make sense to modify those existing blogs/tutorials, but it could also end up being another example altogether where the markdown file itself is committed and so on.

PR #2352, opened 2023-03-23
Only revert on apply failure and then try to sync

aguillenv:
Debugging another error I realized about this and I think this has more sense. On `commit` we should:

- `apply` only relevant changes
- If that fails, `apply` everything
- If that fails, `revert` changes
- `sync` no matter what happened
- `throw` error if any of the previous steps failed
- If nothing fails, return

### DIFF
--- a/src/modules/iasql_functions/iasql.ts
+++ b/src/modules/iasql_functions/iasql.ts
@@ -687,6 +687,14 @@ export async function commit(
         applyErr = e;
       }
     }
+    let revertErr;
+    if (applyErr) {
+      try {
+        await revert(dbId, context, installedModulesSorted, crupdes, currentTransactionId);
+      } catch (e) {
+        revertErr = e;
+      }
+    }
     let syncRes, syncErr;
     try {
       logger.scope({ dbId }).debug('Starting sync phase for all modules');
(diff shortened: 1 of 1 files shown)


aguillenv:
> > Debugging another error I realized about this and I think this has more sense. On `commit` we should:
> > ```
> > * `apply` only relevant changes
> > 
> > * If that fails, `apply` everything
> > 
> > * If that fails, `revert` changes
> > 
> > * `sync` no matter what happened
> > 
> > * `throw` error if any of the previous steps failed
> > 
> > * If nothing fails, return `sync` response
> > ```
> 
> That last point doesn't make as much sense to me. Wouldn't the user expect it to report on the changes that they created being successfully created (or not)? If nothing fails, I'd expect the `apply` output to be returned.

I did not express myself correctly there, it should return everything, `apply` and `sync`, since as we can see [here](https://github.com/iasql/iasql/blob/main/src/modules/iasql_functions/iasql.ts#L711) it `return`s the `syncRes` but the `sync` call [here](https://github.com/iasql/iasql/blob/main/src/modules/iasql_functions/iasql.ts#L693) is mutating the same `crudpes` object, so the final output will have all changes.

dfellis (approved):

depombo (approved):

yrobla (approved):

Issue #1788, opened 2023-01-13
Implement DHCPOptions, InternetGateway, NetworkAcl for EC2


depombo:
- [x] Internet Gateway: https://github.com/iasql/iasql-engine/pull/1795 @mtp1376 
- [x] DHCPOptions
- [ ] NetworkACL
- [ ] Elastic Network Interface

mtp1376:
@depombo should we also add ENI?

depombo:
> @depombo should we also add ENI?

https://discord.com/channels/920684574231646229/949065192275316736/1065663741011693568

depombo:
@mtp1376 / @aguillenv which are missing for this task ?

mtp1376:
> @mtp1376 / @aguillenv which are missing for this task ?

All except the InternetGateway.

yrobla:
will start working on DHCP options

yrobla:
will continue with NetworkACL

PR #2353, opened 2023-03-23
Add network ACL suppport

yrobla:
Partially closes: #1788


dfellis (approved):

src/modules/iasql_functions/iasql.ts
```diff
@@ -674,7 +674,7 @@ export async function commit(
         if (dryRun) return applyRes;
       }
     } catch (e) {
-      logger.scope({ dbId }).warn(`Something failed applying for modules with changes.\n${e}`);
+      logger.scope({ dbId, e }).warn(`Something failed applying for modules with changes.`);
```
aguillenv, line 677: why this change?

src/modules/aws_vpc/mappers/network_acl.ts
```diff
@@ -0,0 +1,205 @@
...
+        return await this.networkAclMapper(rawAcl, region, ctx);
+      } else {
+        const out: NetworkAcl[] = [];
+        const enabledRegions = (await ctx.getEnabledAwsRegions()) as string[];
+        await Promise.all(
+          enabledRegions.map(async region => {
+            const client = (await ctx.getAwsClient(region)) as AWS;
+            for (const eg of await this.getNetworkAcls(client.ec2client)) {
+              if (eg && !eg.IsDefault) {
+                // skip default acls
```
aguillenv, line 128: hmm, interesting. Why do we do this? no one can mess with them or it is to avoid issues like accidentally deleting them?

src/modules/aws_vpc/mappers/network_acl.ts
```diff
@@ -0,0 +1,205 @@
...
+      for (const e of es) {
+        const client = (await ctx.getAwsClient(e.region)) as AWS;
+        const cloudRecord = ctx?.memo?.cloud?.NetworkAcl?.[this.entityId(e)];
+        const isUpdate = this.module.networkAcl.cloud.updateOrReplace(cloudRecord, e) === 'update';
+        if (isUpdate) {
+          let update = false;
+          if (!isEqual(cloudRecord.entries, e.entries)) {
+            // remove all non-default rules and recreate them
+            for (const cloudRule of cloudRecord.entries ?? []) {
+              if (!cloudRule.RuleNumber || cloudRule.RuleNumber > 32766) continue;
```
aguillenv, line 154: nit: create a const for this magic number.

aguillenv: also, does it make sense to add this as a constraint?

test/modules/aws-vpc-network-acl-integration.ts
```diff
@@ -0,0 +1,411 @@
...
+    query(
+      `
+    SELECT * FROM network_acl WHERE tags ->> 'name' = '${prefix}-3';
+  `,
+      (res: any[]) => {
+        expect(res.length).toBe(1);
+      },
+    ),
+  );
+  it('deletes network ACLs', begin());
```
aguillenv, line 338: before deleting, can we uninstall and install the modules and add a test to make sure the sync works as expected?

src/modules/iasql_functions/iasql.ts
dfellis: You can't actually see the error, otherwise, and you can't inspect the error object in mezmo if this isn't done.

src/modules/aws_vpc/mappers/network_acl.ts
yrobla: will do

yrobla: the entries is set as a JSON type, mapped to AWS type directly of NetworkAclEntry. Cannot add the constraint in typeorm

aguillenv (approved):

PR #2364, opened 2023-03-26
remove characters that need encoding from client_reference_id

depombo:
This was not an issue locally because uuids generated locally conformed to the stripe requirements, but the ones generated by auth0 in staging did not

### DIFF
--- a/dashboard/src/pages/api/run.ts
+++ b/dashboard/src/pages/api/run.ts
@@ -300,11 +300,16 @@ async function run(req: NextApiRequest, res: NextApiResponse) {
         const { dbAlias, sql } = req.body;
         const { username, password } = await getUserAndPassword(tokenInfo, dbAlias);
         const isCust = await isCustomer(username);
-        if (!isCust)
+        if (!isCust) {
+          // https://stripe.com/docs/payment-links/url-parameters#streamline-reconciliation-with-a-url-parameter
+          // client_reference_id can be composed of alphanumeric characters, dashes, or underscores, and be any value up to 200 characters.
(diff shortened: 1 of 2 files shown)


dashboard/src/pages/api/run.ts
```diff
@@ -300,11 +311,12 @@ async function run(req: NextApiRequest, res: NextApiResponse) {
         const { dbAlias, sql } = req.body;
         const { username, password } = await getUserAndPassword(tokenInfo, dbAlias);
         const isCust = await isCustomer(username);
-        if (!isCust)
+        if (!isCust) {
           res.status(403).json({
             error: 'User is not a stripe customer',
-            paymentLink: `${config.stripe?.paymentLink}?client_reference_id=${encodeURI(username)}`,
```
dfellis, line 306: You may still want to `encodeURI` here because spaces are supposed to be `%20` in a URL.

depombo: That gets dropped by the regex because it would be invalid and silently dropped by stripe

dfellis: Ahh, got it. I misread the regex and didn't notice the `^` at the beginning of the character class.

dfellis (approved):

dashboard/src/pages/api/run.ts
dfellis: `uid.replace(/[^a-zA-Z 0-9 _ -]+/g, '');`

This character class includes a space (twice) so we still need to `encodeURI` the output here.

dfellis (approved):

PR #2374, opened 2023-03-27
Add `ssh_docker` module

mtp1376:


src/modules/ssh_docker/mappers/docker_container.ts
```diff
@@ -0,0 +1,134 @@
...
+        const docker = await this.getDockerForServer(ctx, e.serverName);
+        const container = await docker!.getContainer(e.containerId);
+        try {
+          await container.stop();
+        } catch (e: any) {
+          // statusCode 304 means the container is already stopped
+          if (e.statusCode !== 304) throw e;
+        }
+        // TODO: maybe we should not remove the container - just stopping would be enough?
+        await container.remove();
```
dfellis, line 124: No, I think this belongs here. There is the `status` column that would show the current status of the container. But that would require a change to the `update` method above.

src/modules/ssh_docker/mappers/docker_container.ts
```diff
@@ -0,0 +1,134 @@
...
+            out.push(container);
+          }
+        }
+        return out;
+      }
+    },
+    update: async (es: DockerContainer[], ctx: Context) => {
+      const out = [];
+      for (const e of es) {
+        // restore because you can't change a container
```
dfellis, line 105: I think the `status` column should be mutable so you can change the state of a container to `stopped` or `started`.

Also, it would be *better* if instead of undoing the changes for the other columns, it flags this update as a `replace` and does a delete/create flow.

src/modules/ssh_docker/mappers/docker_container.ts
```diff
@@ -0,0 +1,134 @@
...
+      await docker.ping();
+    } catch (e) {
+      return undefined;
+    }
+    return docker;
+  }
+
+  cloud = new Crud({
+    create: async (es: DockerContainer[], ctx: Context) => {
+      // create docker containers through RPC
```
dfellis, line 73: You should be able to create docker containers through this flow, I think. An RPC shouldn't be required.

dfellis: Also, I don't see the RPC you're talking about in this PR?

src/modules/ssh_docker/mappers/docker_container.ts
mtp1376: Makes sense. Implemented as you suggested. But I had to select fields such that they are in both `createContainer` and `inspectContainer` - or interchangably usable.

dfellis:
Looking really good. :) I think you need to rebase and/or merge to get the `package.json` (and presumably `yarn.lock`) synchronized, and add a test suite and this is good to go. :)

package.json
dfellis, line 1: Github won't let me leave a comment (and therefore a suggestion) on the correct line, but `@types/jest` should be updated, too.

src/modules/ssh_docker/entity/docker_container.ts
```diff
@@ -0,0 +1,103 @@
...
+  @Column({
+    type: 'varchar',
+    nullable: true,
+  })
+  name?: string;
+
+  @Column()
+  image: string;
+
+  @Column({ nullable: true, type: 'jsonb' })
```
dfellis, line 53: What about (based on [this](https://stackoverflow.com/a/57611955) )

```suggestion
  @Column({ nullable: true, type: 'varchar', array: true })
```

I also wonder if we can have a constraint requiring the `varname=value` syntax, but definitely not a blocker.

src/modules/ssh_docker/entity/docker_container.ts
```diff
@@ -0,0 +1,103 @@
...
+  env?: string[]; // ['A=B', 'C=D']
+
+  @Column({ nullable: true, type: 'jsonb' })
```
dfellis, line 56: Same here, but without the constraint part.

```suggestion
  @Column({ nullable: true, type: 'varchar', array: true })
```

src/modules/ssh_docker/entity/docker_container.ts
```diff
@@ -0,0 +1,103 @@
...
+  command?: string[];
+
+  @Column({ nullable: true, type: 'jsonb' })
```
dfellis, line 59:
```suggestion
  @Column({ nullable: true, type: 'varchar', array: true })
```

src/modules/ssh_docker/entity/docker_container.ts
```diff
@@ -0,0 +1,103 @@
...
+  labels?: { [label: string]: string };
+
+  @Column({ nullable: true, default: dockerContainerStates.running, enum: dockerContainerStates })
+  state?: string;
+
+  @Column({
+    type: 'jsonb',
+    nullable: true,
+  })
+  volumes?: { [volume: string]: {} };
```
dfellis, line 89: Are the values in this key-val pair really objects?

test/modules/ssh-docker-integration.ts
```diff
@@ -0,0 +1,380 @@
...
+  availabilityZone1 = availabilityZones.pop() ?? '';
+  const instanceTypesByAz1 = await getInstanceTypeOffering([availabilityZone1]);
+  instanceType1 = instanceTypesByAz1.InstanceTypeOfferings?.pop()?.InstanceType ?? '';
+  await execComposeUp();
+});
+afterAll(async () => await execComposeDown());
+
+let username: string, password: string, privateKey: string;
+
+describe('SSH Accounts Integration Testing', () => {
```
dfellis, line 85: Nit

```suggestion
describe('SSH Docker Integration Testing', () => {
```

src/modules/ssh_docker/entity/docker_container.ts
mtp1376: I think this one is doable, but I'm sure it was throwing errors in some cases (when the field is an `array` of `json objects`).

src/modules/ssh_docker/entity/docker_container.ts
mtp1376: Yes. That's the docker convention :(. They also use this for exposed ports (values that are empty objects).

dfellis: So that seems to me like it's just using Javascript objects for sets and the values don't actually matter. You can probably simplify this to an array of strings (like the other suggestions above) and work with `Object.keys(objFromDockerLib).sort()` as the `cloud.read` implementation for this column.

src/modules/ssh_docker/entity/docker_container.ts
mtp1376: I changed it to array, but deferred adding the `Check` as an improvement follow-up.

src/modules/ssh_docker/entity/docker_container.ts
```diff
@@ -0,0 +1,168 @@
...
+  /**
+   * @public
+   * State of the docker container - it can be used to start/stop/pause/unpause the container
+   */
+  @Column({ nullable: true, default: dockerContainerStates.running, enum: dockerContainerStates })
+  state?: string;
+
+  /**
+   * @public
+   * Volume definition of the docker container - for example {'vol-name': {}}
```
dfellis, line 139: Nit: This example needs updating

```suggestion
   * Volume definition of the docker container - for example {'vol-name1', 'vol-name2'}
```

dfellis (approved):

test/modules/ssh-docker-integration.ts
```diff
@@ -0,0 +1,489 @@
...
+  it('installs the ssh_docker modules', install(['ssh_docker']));
+
+  it('starts a transaction', begin());
+  itDocs(
+    'creates a new httpd container',
+    query(
+      `
+          INSERT INTO docker_container (server_name, name, image, ports, state)
+          VALUES ('${prefix}', 'httpd-container', 'httpd',
+                  '{"80/tcp": [{"HostIp": "", "HostPort": "8080"}]}', 'created');
```
aguillenv, line 276: what happens if we create the container without defining the `state`?

aguillenv:
this is really nice

src/modules/ssh_docker/mappers/docker_container.ts
```diff
@@ -0,0 +1,231 @@
...
+    out.command = container.Config.Cmd;
+    if (container.Config.Entrypoint) {
+      if (!Array.isArray(container.Config.Entrypoint)) out.entrypoint = [container.Config.Entrypoint];
+      else out.entrypoint = container.Config.Entrypoint;
+    }
+    out.created = new Date(container.Created);
+    out.ports = container.HostConfig.PortBindings ?? undefined;
+    out.labels = container.Config.Labels;
+    out.state = container.State.Status;
+    out.volumes = Object.keys(container.Config.Volumes ?? {}).sort() ?? undefined;
```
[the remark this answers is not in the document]
dfellis: I suggested this, it's so the array can be compared like a set immediately and a difference is just the first time the two arrays differ in an `O(n)` check instead of an `O(n^2)` check

test/modules/ssh-docker-integration.ts
mtp1376: Yes. A `cloud.create` will create it in the `CREATED` state, then a subsequent `cloud.update` will start it.

dfellis (approved):

src/modules/ssh_docker/entity/docker_container.ts
```diff
@@ -8,7 +8,7 @@ import { SshCredentials } from '../../ssh_accounts/entity';
  * @enum
  * Different states a docker container can be - it's used to start/pause/unpause/stop the container
  */
-export enum dockerContainerStates {
+export enum DockerContainerStates {
```
aguillenv, line 11: I think you need to re-generate the migrations too 😬

test/modules/ssh-docker-integration.ts
```diff
@@ -372,6 +372,21 @@ newgrp docker'
     ),
   );
 
+  it('uninstalls the module', uninstall(['ssh_docker']));
+  it('installs the module again', uninstall(['ssh_docker']));
```
aguillenv, line 376: install*

mtp1376: Oops 🤦🏻

test/modules/ssh-docker-integration.ts
dfellis: @mtp1376 he was asking what happens if you *don't* set the `state` parameter, so I told him that it [defaults to `running`](https://github.com/iasql/iasql/pull/2374/files#diff-5ecdbef8c988b5c5dda58c5acec1505812ddd51348015ca5b234ffba947fd7edR137) and so it should immediately start the docker image on the other side.

mtp1376: Yes your explanation is right, but the cloud.create will just create the container and then a cloud.update will start it (because the cloud state is not exactly like local: created vs running).

dfellis: Oh? So it actually always goes through that two-step process?

mtp1376: Yes. It'll go through a two step process currently.

aguillenv (approved):

PR #2386, opened 2023-04-02
ssh_apt

dfellis:
Still WIP, but pushing this up early to point out we have inefficient scaling somewhere in the codebase that needs investigation.

This module populates all of the packages available to install for the user, and install was meant to be just marking `installed` `true`. Reading all of that data from the server (a little over 80k packages on our staging server, for instance) takes ~1 second, construction of the `Package` objects is sub-1-sec (haven't timed it precisely yet, just going by logs) and writing them to the database takes ~5 seconds.

*Then* for whatever reason our engine is pegged at 100% CPU for about 15-30 minutes (I didn't time this properly, yet) before finally releasing back. We must have some super inefficient `O(n^x)` (for some value of `x > 2`) to cause this, and we'll have to investigate and fix that before this PR can be merged (and/or do that work inside of this PR and merge them together).


src/modules/ssh_apt/entity/index.ts
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * @public
+   * Whether or not the package is installed
+   */
+  @Column({
+    type: 'boolean',
+    nullable: false,
+    default: false,
+  })
+  installed: boolean;
```
mtp1376, line 80: So you're keeping all the "possible" packages here and setting `installed: true` for those who are going to be/are installed?

mtp1376: Why not just keeping the installed packages? And `INSERT` when we're going to install a new package?

src/modules/ssh_apt/mappers/index.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export class PackageMapper extends MapperBase<Package> {
+  module: SshAptModule;
+  entity = Package;
+
+  equals = (a: Package, b: Package) =>
+    Object.is(a.server, b.server) &&
+    Object.is(a.package, b.package) &&
+    Object.is(a.version, b.version) &&
+    Object.is(a.architecture, b.architecture) &&
```
mtp1376, line 13: `server`, `package`, `version`, and `architecture` are annotated with `@cloudId` decorator, therefore when two objects are being checked for equality, they're already equal to each other.

src/modules/ssh_apt/mappers/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    Object.is(a.description, b.description) &&
+    Object.is(a.installed, b.installed) &&
+    Object.is(a.upgradable, b.upgradable);
```
mtp1376, line 16:
```suggestion
    Object.is(a.description, b.description) &&
    Object.is(a.installed, b.installed) &&
    Object.is(a.upgradable, b.upgradable);
```

src/modules/ssh_apt/mappers/index.ts
```diff
@@ -0,0 +1,130 @@
...
+            .filter((line: string) => !!line)
+            .map((line: string) => line.trim());
+          const installedPackages = (await client.exec('sudo apt list --installed 2>/dev/null | cat'))
+            .split('\n')
+            .slice(1)
+            .filter((line: string) => !!line);
+          const upgradablePackages = (await client.exec('sudo apt list --upgradable 2>/dev/null | cat'))
+            .split('\n')
+            .slice(1)
+            .filter((line: string) => !!line);
```
mtp1376, line 63: Wondering if there's an npm package that can do these tasks over SSH?

src/modules/ssh_apt/entity/index.ts
dfellis: So there are several reasons why I went with this approach:

1. You can still see just the set that are installed with a `SELECT * FROM package WHERE installed = TRUE`
2. If you remember some phrase that describes the package you want to install but can't remember the exact package name, you can `SELECT * FROM package WHERE description LIKE '%sound editor%' AND installed = FALSE` and then look for the actual package you want to install.
3. If you can't remember the exact spelling of the package, you can use one of the [fuzzy string matching functions](https://www.postgresql.org/docs/current/fuzzystrmatch.html) to help you out, like `SELECT *, levenshtein_less_equal(package, "emacks", 5) AS score FROM package WHERE score < 5 ORDER BY score`
4. We could potentially put a [full-text search index](https://www.postgresql.org/docs/current/textsearch-intro.html#TEXTSEARCH-MATCHING) on the package description to allow easier Google-style searching with minimal symbols in the search text (though this feature is aimed at developers exposing that to end users and still requires a lot of ceremony in the query itself; I also didn't add it to the PR (yet?) because I didn't want to conflate potential sources for the current performance issues by using a new index type)
5. By listing all packages, you can also see which versions are available for installation with `SELECT version FROM package WHERE package = 'vim'`

Basically, there's a lot of classic database value :tm: that you get when the table is fully populated in this way that you would lose if it only listed the currently installed packages. What packages can I install? I dunno! What versions are available? I dunno! Does my CPU architecture have a special version of it? I dunno!

On top of that, Ubuntu/Debian VMs on AWS are pretty "fat" by default. With almost nothing extra installed (just Docker), there's over 8k packages installed on this machine, so the performance issues would probably still rear their head here and we should tackle that, anyways.

src/modules/ssh_apt/mappers/index.ts
dfellis: I wasn't sure exactly which columns would be part of the ID at first so I added everything, then never went back to clean this up.

src/modules/ssh_apt/mappers/index.ts
dfellis: I looked. I didn't see one, but I don't think this code is that bad?

depombo:
> 📏 Image size: 240.12MB (compressed) - 778.59MB (uncompressed)
> ⏳ All modules installed in: 47.969944531 seconds ⌛️
> ⏳ All modules uninstalled in: 1.174685437 seconds ⌛️

shouldn't these numbers be lower now?

dfellis:
> > straight_ruler Image size: 240.12MB (compressed) - 778.59MB (uncompressed)
> > hourglass_flowing_sand All modules installed in: 47.969944531 seconds hourglass
> > hourglass_flowing_sand All modules uninstalled in: 1.174685437 seconds hourglass
> 
> shouldn't these numbers be lower now?

They've hovered a little over 800MB uncompressed for a while now. What do you think it should be at?

src/services/diff.ts
```diff
@@ -1,31 +1,43 @@
...
+  // I feel like this can be done with a single pass, but I'm not sure exactly how, yet. At least it's
+  // not three passes
+  for (const aVal of a) {
+    if (b.has(aVal[0])) intersect.set(aVal[0], [aVal[1], b.get(aVal[0])!]);
+    else left.set(...aVal);
+  }
+  for (const bVal of b) {
+    if (!a.has(bVal[0])) right.set(...bVal);
+  }
+  return [left, intersect, right];
```
depombo, line 14: is there a way enforce or passively keep track of the timings for this module to make sure there are not regressions on this logic?

yrobla (approved):

src/modules/ssh_apt/mappers/index.ts
```diff
@@ -0,0 +1,219 @@
...
+      // only those in the database
+      const newPackages = [];
+      for (const e of es) {
+        const id = this.generateId({
+          server: e.server,
+          package: e.package,
+          version: e.version,
+          architecture: e.architecture,
+        });
+        const newPackage = (await this.cloud.read(ctx, id)) as Package;
```
mtp1376, line 198: nit: can use `cloud.read(ctx)` without the `id` field and do the filtering here instead?

mtp1376 (approved):

src/modules/ssh_apt/mappers/index.ts
dfellis: This keeps it DRY instead? (And the `memo` cache should be hit so it doesn't actually make another call to the server.)

PR #2401, opened 2023-04-05
Add icons to query table results

mtp1376:


dashboard/package.json
```diff
@@ -21,6 +21,7 @@
     "@logdna/logger": "^2.6.6",
     "@next/font": "13.1.6",
     "@sentry/react": "^7.37.1",
+    "@types/react-icons": "^3.0.0",
```
aguillenv, line 24: hmm, aren't the hero icons enough for this?

dashboard/src/components/common/Table.tsx
```diff
@@ -1,48 +1,86 @@
 import { Label } from '.';
+import { IconContext } from 'react-icons';
+import { RxCode } from 'react-icons/rx';
+import { VscSymbolBoolean, VscSymbolKey } from 'react-icons/vsc';
+import { MdAccessTime, MdDateRange, MdNumbers, MdQuestionMark } from 'react-icons/md';
 
-export default function Table({ data }: { data: { [key: string]: any }[] }) {
+function getIconForDataType(dataType: number | undefined) {
+  switch (dataType) {
+    case 16: // boolean
```
aguillenv, line 9: interesting, where are these numbers coming from?

dashboard/src/components/common/Table.tsx
```diff
@@ -1,48 +1,86 @@
...
+        {columns.map(column => (
+          <th
+            className='group px-6 py-3 text-left text-xs font-medium text-gray-500 uppercase tracking-wider'
+            scope='col'
+            key={column}
+          >
+            <IconContext.Provider value={{ style: { display: 'inline-block' } }}>
+              {getIconForDataType(dataTypes?.[column])}
+            </IconContext.Provider>
+            <div style={{display: 'inline-block', paddingLeft: '5px'}}>{column}</div>
```
aguillenv, line 56: your problem moving lines maybe will get fixed using a HBox as parent of these 2 elements.

Also, why do we use the `IconContext`?

Lastly, let's avoid using inline styles, you could use classes for those like `inline-block` and/or `pl-x`

dashboard/src/components/common/Table.tsx
mtp1376: https://github.com/brianc/node-pg-types/blob/master/lib/builtins.js#L12-L73

dashboard/src/components/common/Table.tsx
mtp1376: `IconContext` is the proposed way to set styles in the `react-icons` library.

dashboard/package.json
mtp1376: I checked hero icons and it does not contain the one we need here.

dashboard/src/components/common/Table.tsx
```diff
@@ -1,6 +1,51 @@
...
+      return <MdDataObject />;
+    case 1082: // date
+    case 1083: // time
+      return <MdDateRange />;
+    case 1114:
+    case 1184: // timestamp
+      return <MdAccessTime />;
+    case 2950: // uuid
+    default:
+      return <MdQuestionMark />;
```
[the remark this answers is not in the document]
mtp1376: Those numbers are coming from here:
 https://github.com/brianc/node-pg-types/blob/8d2200fd9cecf41f40acc1432c01a316178e711f/lib/builtins.js#L2

I think each user defined enum has a specific id and I'm treating everything "else" as unknown variables. I'll check if there's a specific range of values for user defined types.

For UUID case, I didn't find a good icon. So I made it fall back to the default icon.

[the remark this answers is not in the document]
dfellis: So we can get the user-created type ids with a minor update on [this query](https://dba.stackexchange.com/a/239727)

```sql
SELECT
  pg_type.oid,
  pg_type.typname, 
  pg_enum.enumlabel
FROM
  pg_type 
JOIN
  pg_enum ON pg_enum.enumtypid = pg_type.oid;
```

Probably only need the `pg_type.oid` here and *might* be able to get away with no join like this:

```sql
SELECT pg_enum.enumtypid from pg_enum
```

This could be put into the initialization queries and referenced here with something like:

```suggestion
      return dbEnumIds.includes(dataType) ? <MdSomething /> : <MdQuestionMark />;
```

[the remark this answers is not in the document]
dfellis: The simpler solution for now may just be to use an appropriate icon for enums and not worry about a mismatch. Do we have a non-enum/non-uuid type in our schemas that would get caught by this `?` path? If not, default to "assume it's an enum" seems reasonable to me?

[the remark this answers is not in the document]
aguillenv: > For UUID case, I didn't find a good icon. So I made it fall back to the default icon.

I think UUID could fall into the `text` category

dashboard/src/components/common/Table.tsx
```diff
@@ -15,7 +60,12 @@ export default function Table({ data }: { data: { [key: string]: any }[] }) {
               scope='col'
               key={column}
             >
-              {column}
+              <HBox alignment={align.start}>
+                <IconContext.Provider value={{ className: 'inline-block', size: '1.5em' }}>
+                  {getIconForDataType(dataTypes?.[column])}
+                </IconContext.Provider>
+                <div className={'inline-block pl-3'}>{column}</div>
```
aguillenv, line 67: nit: try it out but I think I'd reduce the left padding a bit, I feel too much distance between the icon and the text
```suggestion
                <div className={'inline-block pl-1'}>{column}</div>
```

aguillenv (approved):

dfellis (approved):

PR #2407, opened 2023-04-06
Fix deletion of routes in the default route table

mtp1376:


dfellis (approved):

depombo (approved):

test/modules/aws-vpc-routetable-integration.ts
```diff
@@ -290,16 +290,19 @@ describe('RouteTable Integration Testing', () => {
...
-      `,
-      (res: any[]) => expect(res.length).toBeGreaterThan(0),
+          JOIN route_table rt on rt.id = route.route_table_id
+          JOIN vpc v on rt.vpc_id = v.id
+          JOIN internet_gateway ig on v.id = ig.vpc_id
+        WHERE v.is_default AND route.region = '${region}'
+        GROUP BY route.route_table_id, route.region, ig.internet_gateway_id, rt.route_table_id
+        HAVING ('0.0.0.0/0' <> ALL (array_agg(destination)));
+               `,
+      (res: any[]) => expect(res.length).toBe(0),
```
dfellis, line 305: This is confirming zero returned records. Is that correct? I would have assumed it should be greater than zero like before?

yrobla (approved):

PR #2409, opened 2023-04-06
Don't use migrations with Django

mtp1376:


test/tutorials/django-ci.sh
```diff
@@ -12,9 +12,28 @@ cd examples/ecs-fargate/django/app
...
 
-echo "\nRun Django migrations"
+# install ecs_simplified and codebuild modules
+psql "postgres://$IASQL_USERNAME:$IASQL_PASSWORD@localhost:5432/iasql" -c "
+  SELECT * FROM iasql_install(
+      'aws_ecs_simplified', 'aws_codebuild'
+  );
+"
+
+echo "\nRun Django migrations - no real action on the database"
```
dfellis, line 22: Is this a requirement of Django itself? (Eg, if we comment this out follow-up stuff doesn't work?)

test/tutorials/django-ci.sh
```diff
@@ -12,9 +12,28 @@ cd examples/ecs-fargate/django/app
...
 python manage.py migrate --database infra infra
 
+# create the ecs_simplified app
+psql "postgres://$IASQL_USERNAME:$IASQL_PASSWORD@localhost:5432/iasql" -c "
+  SELECT iasql_begin();
+"
+psql "postgres://$IASQL_USERNAME:$IASQL_PASSWORD@localhost:5432/iasql" -c "
+  INSERT INTO ecs_simplified (app_name, public_ip, app_port, image_tag, cpu_mem, desired_count)
+  VALUES ('quickstart', true, 8088, 'latest', 'vCPU2-8GB', 1);
+"
```
dfellis, line 32: With this being here, what's actually done in Python?

test/tutorials/django-ci.sh
mtp1376: No. We can omit it and nothing will fall apart.

test/tutorials/django-ci.sh
mtp1376: The webserver itself which will be deployed. And the infra project has the tables imported as models.

dfellis: But these tables aren't used to actually manage the deployment of the infrastructure, which was kinda the point of these examples? A Pulumi-like layer on top of IaSQL in whatever language you want, but you don't need a Pulumi-like company to get there, as we show examples how to do it with a couple of popular languages using a couple of popular ORMs and the ORMs are the Pulumi-like layer.

dfellis (approved):
Very nice :)

aguillenv (approved):

PR #2412, opened 2023-04-07
Add tag support for S3 buckets

mtp1376:


dfellis (approved):

test/modules/aws-s3-integration.ts
```diff
@@ -463,6 +463,36 @@ describe('S3 Integration Testing', () => {
     );
   });
 
+  describe('S3 bucket tags testing', () => {
```
aguillenv, line 466: can we test tags on the create path as well?

dfellis (approved):

depombo (approved):

PR #2414, opened 2023-04-07
Reworking and unifying query recreation

aguillenv:
...and a couple of load balancer fixes.


src/modules/iasql_functions/iasql.ts
```diff
@@ -783,91 +798,262 @@ async function getChangeLogsSinceLastBegin(orm: TypeormWrapper): Promise<IasqlAu
...
+/**
+ * @internal
+ * Recreate entities from change logs indexed by entity name
+ */
+async function recreateEntitiesFromChangelogs(
+  changeLogs: IasqlAuditLog[],
+  modsIndexedByTable: ModsIndexedByTable,
+  orm: TypeormWrapper,
+): Promise<EntitiesIndexesByEntityName> {
+  const entityMapper: { [key: string]: MapperBase<any> } = {};
```
dfellis, line 922: Nit: This logically should be within the `for` loop, as far as I can tell. It's never used outside of the for loop

src/modules/iasql_functions/iasql.ts
```diff
@@ -923,46 +1109,207 @@ async function getTableMetadata(
...
+      );
+      break;
     }
+    case AuditLogChangeType.DELETE: {
+      query = format(
+        `
+          DELETE FROM %I
+          WHERE ${firstValues
+            // We need to add an special case for AMIs since we know the revolve string can be used and it will not match with the actual AMI assigned
+            .filter(e => e.key !== 'ami')
```
dfellis, line 1150: # :cry:

dfellis: Is there *any* way we could generalize this? There are some other places in the AWS SDK where `resolve:` is allowed that we'll eventually support, and *no one* is going to remember this trickery here, but if we create our own `@awsResolveSupport` decorator or something that can be used here to filter them out we stand a chance of not constantly stubbing our toe on this.

src/modules/iasql_functions/iasql.ts
```diff
@@ -923,46 +1109,207 @@ async function getTableMetadata(
...
+            .map(e => `${formatValue(e) === 'NULL' ? '%I IS %s' : e.isJson ? '%I::jsonb = %s' : '%I = %s'}`)
+            .join(' AND ')};
+        `,
+        tableName,
+        ...firstValues
+          // We need to add an special case for AMIs since we know the revolve string can be used and it will not match with the actual AMI assigned
+          .filter(e => e.key !== 'ami')
```
dfellis, line 1157: Same as above, this shouldn't be hardwired.

src/modules/iasql_functions/iasql.ts
```diff
@@ -923,46 +1109,207 @@ async function getTableMetadata(
...
+      // while updatedValues are the ones that are going to be used in the SET clause
+      const originalValues = inverse ? secondValues : firstValues;
+      const updatedValues = inverse ? firstValues : secondValues;
+      query = format(
+        `
+        UPDATE %I
+        SET ${updatedValues?.map(_ => `%I = %s`).join(', ')}
+        WHERE ${originalValues
+          // We need to add an special case for AMIs since we know the revolve string can be used and it will not match with the actual AMI assigned
+          ?.filter(e => e.key !== 'ami')
```
dfellis, line 1183: another one

src/modules/iasql_functions/iasql.ts
```diff
@@ -923,46 +1109,207 @@ async function getTableMetadata(
...
+          ?.map(e => `${formatValue(e) === 'NULL' ? '%I IS %s' : e.isJson ? '%I::jsonb = %s' : '%I = %s'}`)
+          .join(' AND ')};
+      `,
+        tableName,
+        ...(updatedValues?.flatMap(e =>
+          e.isJson && Array.isArray(e.value) ? [e.key, `${formatValue(e)}::jsonb`] : [e.key, formatValue(e)],
+        ) ?? []),
+        ...(originalValues
+          // We need to add an special case for AMIs since we know the revolve string can be used and it will not match with the actual AMI assigned
+          ?.filter(e => e.key !== 'ami')
```
dfellis, line 1193: Here, too

src/modules/iasql_functions/iasql.ts
```diff
@@ -923,46 +1109,207 @@ async function getTableMetadata(
...
+              orm,
+              shouldRecreateSubQueries,
+              recreatedEntitiesFromChangelogs,
+            ),
+        ),
+      );
+    }
+    const subQuery = format(
+      `SELECT %I FROM %I WHERE ${values
+        ?.filter(v => v.key !== 'ami')
```
dfellis, line 1303: Don't forget this one

test/helpers.ts
```diff
@@ -107,9 +107,14 @@ export function runQuery(
   assertFn?: (res: any[]) => void,
   log = true,
   withUserAndPassword = () => ({ username: config.db.user, password: config.db.password }),
+  queryReplacements: () => { [key: string]: string } = () => ({}),
```
dfellis, line 110: What is this, and why can't it be done with normal `` ` `` strings?

aguillenv: no, jest does like a first pass grabbing all statements and the current variable values. I guess it is because you can run tests in parallel and tests should be independent? Since we are assigning these values in another test assert step, we have to do this trickery to have the latest value for the variable. It replicates the same behaviour as `username` and `password`.

dfellis: oh, that can be resolved by wrapping in an extra `async (done) => ...` closure scope. We do that in a few places already.

aguillenv: I tried something like that too, but it was not working when calling the `query` method inside since it has its own logic returning a (done) => void function itself

dfellis: It's ugly, but [this is the pattern we use](https://github.com/iasql/iasql/blob/main/test/modules/ssh-apt-integration.ts#L204-L223)

dfellis (approved):

depombo (approved):

PR #2427, opened 2023-04-10
Maybe fix aurora intermittent error

### LINKED ISSUE #2406: RDS test flakiness
This error has appeared at least a couple of times now. Apparently is picking the wrong engine version while creating the aurora cluster.

```
RDS Integration Testing › Aurora instances not supported or messed with › creates an aurora cluster
[117471](https://github.com/iasql/iasql/actions/runs/4621801892/jobs/8173785312#step:9:117472)

[117472](https://github.com/iasql/iasql/actions/runs/4621801892/jobs/8173785312#step:9:117473)
    QueryFailedError: Engine version not found: oscar/null
```

https://github.com/iasql/iasql/actions/runs/4621801892/jobs/8173785312

aguillenv:
Maybe resolves #2406

### DIFF
--- a/test/modules/aws-rds-integration.ts
+++ b/test/modules/aws-rds-integration.ts
@@ -1,3 +1,5 @@
+import { EC2 } from '@aws-sdk/client-ec2';
+
 import * as iasql from '../../src/services/iasql';
 import {
   defaultRegion,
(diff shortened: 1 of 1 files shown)


test/modules/aws-rds-integration.ts
```diff
@@ -411,7 +440,7 @@ describe('RDS Integration Testing', () => {
       query(`
       SELECT invoke_rds(
         'createDBCluster',
-        '{"Engine": "aurora", "AvailabilityZone": "us-west-2a", "DBClusterIdentifier": "${prefix}hidden", "MasterUserPassword": "dontcare", "MasterUsername": "dontcare"}',
+        '{"Engine": "aurora", "AvailabilityZone": "${availabilityZone}", "DBClusterIdentifier": "${prefix}hidden", "MasterUserPassword": "dontcare", "MasterUsername": "dontcare"}',
```
mtp1376, line 443: I don't get how this can fix things

aguillenv: I'm not 100% sure, to be honest, but it was the main weird thing I saw... 

The issue is that in order to create the aurora cluster, we are calling the `aws_sdk` directly and it is responding with a wrong engine version error. My suspicion is that the fixed `availability_zone` might be interfering with the command to pick the right engine version for some regions (since the region is changing dynamically).

dfellis (approved):
We're doing this availability zone check in enough test suites that it's probably time to add an `AvailabilityZone` table to `aws_account`. Not a blocker on this PR, of course.

test/modules/aws-rds-integration.ts
aguillenv: yes

